PR #128, opened 2021-05-31
Use dummy names rather than actual project names in dummy data

aragilar:
Lake Mungo -> AstroSky (silly name to make it easy to refer to)
CSIRO Geochemistry -> NotParkes (because it's not the Parkes dish ;)


riniangreani:
Sorry I didn't see this PR before. Now it has conflicts with my latest. I'll resolve it first.

riniangreani:
I can no longer see "Lake Mungo" project on projects page now.

PR #132, opened 2021-06-02
Re-Label the 'Save/Add' button in the observation form

aidanfar0:
to resolve confusion about the alpha level features

https://faimsproject.atlassian.net/browse/FAIMS3-25


aidanfar0:
Yes, it's validate and prep for sync

PR #133, opened 2021-06-02
(For beta) Sync refactor

aidanfar0:
Breaks up sync/index.ts into multiple files,
and will add in the missing link between react having to listen directly to events to then get state.


aragilar:
(marking as draft as github doesn't recognise the WIP prefix)

PR #140, opened 2021-06-04
Fix location not staged

aidanfar0:
Addresses https://faimsproject.atlassian.net/browse/FAIMS3-216
This is quite a bit hackey. Listening on onChange/onFocus already felt a bit jankey, and this adds to it. Fixing these bits will come in staging-refactor branch

### DIFF
--- a/src/gui/components/observation/form.tsx
+++ b/src/gui/components/observation/form.tsx
@@ -463,15 +463,16 @@ class ObservationForm extends React.Component<
     this.stageInterval = window.setInterval(main_save_func, STAGING_SAVE_CYCLE);
   }
 
-  interceptChange<E>(
-    handleChange: (evt: E) => unknown,
+  interceptChange<E, R>(
+    handleChange: (evt: E) => R,
     formProps: FormikProps<any>,
     fieldName: string,
-    evt: E & {currentTarget: {name: string}}
-  ): void {
-    handleChange(evt);
+    evt: E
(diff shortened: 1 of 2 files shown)


Denubis:
I'm worried about this PR. Would it be more prudent to write long-lat to a toast instead? Remember, we're in alpha -- we can tell rather than show bits that we know are problematic. Take point was intended to show we can communicate with sensors (and has a whole heap of NMEA 0183 implications for late beta). 

Are you confident in this change and its reversability?

aidanfar0:
Location saving is not broken. It's working fine as it is as a button that you click, then lat-long saved to the DB.
The issue is that when you close & re-open the browser tab or something to that effect, while the rest of the values in the form come back (they were "staged") the lat & long doesn't.

If, as you say, we only need take point button for showing device communication, this PR can be closed, replaced with _just_ the refactor that i mentioned, since the functionality is there.

After discussion with James, the different approach I am more inclined to take is similar to this hack. (but not hackey). Each custom (non-HTML) component is passed into its props a "valueChanged" or "setValue" (name tbd) function.

- No need to diff the Formik form's values every render (which may get messy). 
- Custom components (e.g. maps) can save extra/different date to the staging area than they do to the actual form (i.e. if you're drawing something on a map, the draw points may be needed while you're drawing but not in the final data)
- However this needs separate logic for native HTML components that still need the onChange= & onBlur= handlers, and custom components that we pass in our "setValue" / "valueChanged" functions

PR #143, opened 2021-06-07
Set created/created_by (no more dummy values)

aragilar:
Fixes https://faimsproject.atlassian.net/browse/FAIMS3-233, https://faimsproject.atlassian.net/browse/FAIMS3-237.


aragilar:
Replaced by #144.

PR #144, opened 2021-06-08
Hide creation information for now

aragilar:
There is a planned change in how we're going to store the data to handle
the ACLs and work with revisions in a more user-friendly way, but we're
holding that off to beta. That will make managing the creation
information easier (so we don't need to try to keep track of it via the
form), so we're going to hide the dummy data for now.


aragilar:
Closes https://faimsproject.atlassian.net/browse/FAIMS3-233, https://faimsproject.atlassian.net/browse/FAIMS3-237

PR #164, opened 2021-06-10
Fix any remaining useEffect without dependency [] array

aidanfar0:


lmannering:
Looks good. Passing an empty array as the second argument to useEffect keeps it running only on mount and unmount, which should stop any infinite loops, but we might want to go back and add dependencies if a component needs to re-render e.g., if props are changed.

PR #172, opened 2021-08-05
Add new split data model

aragilar:
This adds the new data model as we've been discussing.

This doesn't include the automatic merging or similar thing, that will come after this.


Denubis:
Can we regex observation and the like to bring it closer in line with the glossary? I'm worried that if we leave it past here we will... end up with a major disconnect between code and UI. (And we're not exposing the term observation to humans).

aragilar:
@Denubis Yep, can do. I was going to refactor datamodel.ts into the different subsystems and then do the rename as two follow-up PRs (to keep the changes relatively small), but if that's going to cause more confusion, I'll add those to this PR.

Denubis:
So long as it's in the pipeline -- happy to comparmentalise PRs. Just so we're on the same page,
`project_id -> notebook_id` 
`observation_id -> atttribute_value_pair_id?` (What a mouthful but I think you use it more in the attribute-value-pair sense than the record sense, but I'm not confident) The problem is that we have things like ObservationList. But I'm easily confused...

PR #174, opened 2021-08-11
Staging refactor

aidanfar0:
Moves logic for staging area into a new object and file, out of the ObservationForm

### DIFF
--- a/LICENSE-TEMPLATE-FOR-SRC
+++ b/LICENSE-TEMPLATE-FOR-SRC
@@ -14,6 +14,6 @@
  * limitations under the License.
  *
  * Filename: actions.ts
- * Description:
+ * Description: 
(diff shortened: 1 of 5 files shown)


aragilar:
@aidanfar0 Can you rebase to resolve the conflicts? Once that's done then I'll look at the changes again.

aidanfar0:
Whilst I have rebased this, the staging area now doesn't work for draft documents.
This is because it previously used 'null' as the record ID for drafts, but now that is not possible (due to better types). 
Each time you make a new record, you create a new ID, instead of looking for an existing draft,
So don't merge this yet!

aidanfar0:
its fixed now

aidanfar0:
its fixed now

PR #175, opened 2021-08-11
Rename datum to aent_value (as per glossary)

aragilar:
@Denubis This is the first of the renaming PRs. I think AentValue is the closest in meaning to the use of datum, but it would be good to confirm.


aragilar:
Sure, Attribute Value Pair is cool, I've shortened it to avp in places.

Denubis:
Looks good to me. Thanks for putting up with me.

PR #177, opened 2021-08-16
Basic automerging

aragilar:
This currently adds the logic for basic automerging.

I've set this as draft as I need to add some tests. I'm creating the PR now as, I'd like people to look over this, and see if they can spot any issues.

Also, `mergeHeads` should be called by the sync code when there are incoming records, @aidanfar0 where's the best place to hook that in?

### DIFF
--- a/src/data_storage/merging.test.ts
+++ b/src/data_storage/merging.test.ts
@@ -18,24 +18,17 @@
  *   TODO
  */
 
-import {testProp, fc} from 'jest-fast-check';
 import PouchDB from 'pouchdb';
 
 import {ProjectID} from '../datamodel/core';
 import {Record} from '../datamodel/ui';
 import {
-  deleteFAIMSDataForID,
   generateFAIMSDataID,
-  getFirstRecordHead,
-  getFullRecordData,
-  listFAIMSProjectRevisions,
-  undeleteFAIMSDataForID,
   upsertFAIMSData,
+  setRecordAsDeleted,
 } from './index';
-import {getRecord} from './internals';
+import {getRecord, getRevision} from './internals';
 import {mergeHeads} from './merging';
-
-import {equals} from '../utils/eqTestSupport';
 
 import {getDataDB} from '../sync/index';
 
@@ -325,7 +318,7 @@ describe('test basic automerge', () => {
       updated: time,
     };
 
-    const revision_id2 = await upsertFAIMSData(project_id, doc2);
+    await upsertFAIMSData(project_id, doc2);
 
     const doc3: Record = {
       project_id: project_id,
@@ -407,7 +400,7 @@ describe('test basic automerge', () => {
       updated: time,
     };
 
-    const revision_id2 = await upsertFAIMSData(project_id, doc2);
+    await upsertFAIMSData(project_id, doc2);
 
     const doc3: Record = {
       project_id: project_id,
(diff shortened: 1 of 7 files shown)


aidanfar0:
I'm not sure what file it might go into, maybe sync/state.ts given how it's implemented, but

I think it'd look like (Like the other files in sync/state.ts) the following:

```
When a data DB updates, make sure to listen for changes from it:

function register_merge_resolver(initializeEvents: DirectoryEmitter) {
    let already_listening = new Set<string>();
    initializeEvents.on('project_data_paused', (listing, active, project, data) => {
        if (!(project._id in already_listening)) {
            already_listening.add(project._id);
            
            start_listening_for_changes(data);
        }
    });
}

// Listen When PouchDB Emits changes, 

function start_listening_for_changes(proj_id: NonUniqueProjectID, data_db: LocalDB<EncodedRecord>) {
    data_db.changes({since: /* NOTE1 */, include_docs: true}).on('change', doc => {
        mergeHeads(..., doc._id i think)
    });
}
```

NOTE1: This is something we'll have to add to local storage: Per project, we need to know what the last thing that was in the DB was. Alternatively, I could add an optional arg to 'project_data_paused' event that has the changes in it.

aragilar:
We'd need to check that the id is actually a record id, but otherwise that looks fine (though maybe the automerging should be called as part of watching for new/changed observations as part of the UI?).

aragilar:
I've added a list of tests for merging that I'm going to start implementing, if people can ideas for cases I've missed, post them in this PR.

src/data_storage/merging.test.ts
```diff
@@ -0,0 +1,109 @@
...
+describe('test basic automerge', () => {
+  test('single revision', async () => {
+    // This tests the case where there is a single revision (i.e. new record)
+  });
+  test('no merged needed', async () => {
+    // This tests the case where there is a linear history, so there's no need
+    // for merging
+  });
+  test('extra head', async () => {
+    // This tests the case where there is a linear history, but where an old
```
Denubis, line 81: I think this simulates a sync power/network failure? Making sure that we've got ACID commits here is probably useful.

src/data_storage/merging.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    // head was not removed (this shouldn't happen, but maybe there's some bad
+    // integration code that wrote to couchdb
+  });
+  test('same change', async () => {
+    // This tests the case where there has been a split, but the same change has
```
Denubis, line 86: Just to make sure I'm getting the context, this is operating at the *record* not *pair* level. And so this test is testing that the same change of the same record's pair, each client starting from the same head?

src/data_storage/merging.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    // been made
+  });
+  test('different change', async () => {
+    // This tests the case where there has been a split, and different changes
+    // have been made. This should cause the basic automerge to fail.
+  });
+  test('changes to different avps', async () => {
+    // This tests the case where there has been a split, but the changes have
```
Denubis, line 94: Does change imply insert as well as update? And are metadata updates handled as changes in this regard? Multivalued checkbox-selects?

src/data_storage/merging.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    // been to different avps
+  });
+  test('changes to different avps AND different change', async () => {
+    // This tests the case where there are three heads, of which two can be
+    // merged
+  });
+  test('changes to different avps AND different change 4 HEADS', async () => {
+    // This tests the case where there are 4 heads, of which three can be merged
```
Denubis, line 102: Just thinking about classical normalisation anomalies. If someone unchecks all items of a checkbox, that's still update, not delete in this context, right? (We should never be issuing a delete anyways). And I think derived values (I haven't heard any design discussion around this, so feel free to put this side of things off) will be derived at derivation-target-update, rather than at view, so to this system they just look like normal writes.

Ah, what about timetravel? Bad clocks are evil -- are we relying on timestamps for *literally anything?* as part of this process? I hope not. What if we have temporally distant updates?

Ex:

Device 1: 1. Make record -> 2. Sync -> 3. Go offline -> 18 or so. Edit record -> 19. Go Online 20. Sync
Device 2: 2. Sync, 4. Edit, 5. Sync, 6. Edit, 7. Sync. 8 Edit ... The fact that device 1's edit is so far back in the "history" of the thing... is that part of this testing regimen?

src/data_storage/merging.test.ts
aragilar: This won't happen due to a sync power/network failure, only if someone doesn't update the heads correctly in a record (so possible third parties). This isn't a dangerous case (in theory), it's more to test robustness of the process where there could be bad values in the heads.

src/data_storage/merging.test.ts
aragilar: Two people make the same change on two difference devices, basic automerging does not handle that (we need to be able to check the values and see that there the same, which basic automerge cannot do). An advanced version of automerge (which would be used to assist project leads/someone else who had full access to the data) would have doing this case in scope.

src/data_storage/merging.test.ts
aragilar: Metadata is per AVP, so it's all the same to the basic automerge.

src/data_storage/merging.test.ts
aragilar: Yep, changes to a form field would result in a new avp, so no deletion (though that brings up a case we need to check, what if you try to merge a revision which is marked as deleted, with a non-deleted one)?

Automerging doesn't care about timestamps, the only time timestamps (or time) comes up in the merging code is what time to set for the new revision (so nothing stops there being a merge revision where its parents have creation times after it because someone's local device clock is massively off, that's the users fault if they get themselves confused).

Note that in the case where there is a conflict the automerger *cannot* fix, and the *user* has to do a manual merge, if the user thinks that because one version has a newer timestamp (due to bad clocks) and goes with that one, that's not something I think we can handle (other than doing something really horrible like trying to correct timestamps in revisions given an offset computed for that bad device, which seems to me to be a path of madness).

Denubis: No, we are not expected to handle horrible timestamps. 2.6 had... a bad method for that, but it was needed. And yes, the case you raise is one we need to check for.

src/data_storage/merging.test.ts
Denubis: Great.

Denubis (approved):
Ok, I am happy with the current suite. I think we'll need to see stuff in action for further comments. This seems fine for now.

aragilar:
Other than the sync integration, this is ready.

aidanfar0 (approved):

PR #178, opened 2021-08-19
Dummy data setup was broken, fix it

aragilar:
Looks like something broke dummyData, this fixes it so we can continue testing new fields.


Denubis:
bors try

Denubis:
bors ping

Denubis:
bors try

Denubis:
@aragilar I think this is the point where I hand it to you. It runs, and spawns the trying things -- but since only one of the three builds failed, it might be a code thing?

PR #194, opened 2021-09-03
Variants

aidanfar0:
This adds a layer on top of record views: variants
Each record can be only 1 of the variants, and records can't change what variant they are (yet)
A record being a different variant means it has a whole different set of fields and different views

Views can be shared between variants (like how components can be shared between views, i think)

In terms the database datamodel, the variant type of a record/revision is stored on the Revision,
This is because it's not an attribute-value-pair, but I still would like to allow changes to the variant type (so it can't go on the Record)

TODO: Better UI

### DIFF
--- a/src/data_storage/merging.ts
+++ b/src/data_storage/merging.ts
@@ -201,6 +201,11 @@ export async function do3WayMerge(
     return await doFastForward(project_id, merge_result, base, them, us);
   }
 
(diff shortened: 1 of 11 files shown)


aidanfar0:
I haven't thorougly checked how the staging area interacts with this, but I think on a cursory analysis it's not loosing too much data

PR #199, opened 2021-09-14
Fix variants new record

aidanfar0:
This fixes a lot of new 404s that were occuring when trying to create new records.
The 'Add' button on the homepage now separates into different buttons for each viewset,

Technical note on the issue I discussed with James: To fix it,
/new-record/type/:type_id has changed to /new-record/type/:type_id/record/:record_id,
Where there is a redirect from the former to the latter that is where the UUID is generated now (It's a prop in RecordCreate element)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,7 +19,12 @@
  */
 
 import React from 'react';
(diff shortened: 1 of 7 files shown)


aragilar:
bors ping

PR #203, opened 2021-09-16
Add helper for creating a new local project

aragilar:
@KateSHENG This should allow you to create local projects for use in the project builder.
@aidanfar0 Can you sanity check this and then merge? There's also a hook for writing up to the conductor, once that's read.


aidanfar0:
!bors +r

PR #205, opened 2021-09-16
Make the mutli step form behave more nicely

aragilar:
This splits apart the logic in the form class into self-contained bits, leaving it handling the top level form state. Views and fields are now more self-contained, making it easier to reason about the state.

I'm still tossing up pulling out the initial values logic from the form.

We probably also want to add buttons at the bottom for prev/next, and have errors displayed somehow on the view bar at the top (I'll create an issue for this).


aidanfar0 (approved):
I've made changes to staging to get rid of the dependency on view name. Clearing the form now clears every view

aidanfar0:
bors +r

PR #209, opened 2021-09-30
Documentation demo

Denubis:
Added command `npm run docs` to build docs. Once accepted, I'll get github pages to point to `main` instead of the branch.


aragilar:
@Denubis Looks like this produces lots of files, which confuses the build, can we put the docs on a different branch?

Denubis:
Trivially. Right now it's rendering from `documentation-demo` without drama.

Denubis:
I'd say reject this PR, rename s/-demo// and call it good? In the fullness of time, we'll figure out how to get bors to commit into the thing, but... not right now?

PR #214, opened 2021-10-06
Ensure that local projects get connected to events

aragilar:
@aidanfar0 I'm getting an odd issue when I create a local project and refresh the page, the initialisation doesn't resolve, can you take a look?


aragilar:
Aidan's changes fix the issue I had, so merging.

bors r+

PR #216, opened 2021-10-07
Fix multiple drafts lists constantly loading

aidanfar0:


aragilar:
The drafts lack ids/links, so they can't be edited: [
Screenshot 2021-10-07 at 17-48-04 FAIMS3
](url)

Otherwise it seems to stop the continuous loading.

PR #227, opened 2021-10-08
Add editable_by for storing ACL information

aragilar:
I'm not sure if we want read-only access as well. We'd need to pass that information down to the field, which may or may not make sense to the field.


Denubis:
@aragilar  this seems *super* stale, is this relevant?

PR #232, opened 2021-10-13
Add dependency on Steve's map package

aragilar:
This should avoid the issues with the locks in #208. I'm going to add the rest of that PR on top of this branch, in a different branch (to keep things clean).


aragilar:
@Denubis @stevecassidy It looks like even though https is specified in `package.json`, npm wants to use ssh in the `package-lock.json`. I suspect the best way forward would be to upload the package to npm.

Denubis:
Want me to set up auto-CI to do this?

PR #269, opened 2021-11-10
Update take point

aragilar:
@Denubis @stevecassidy Does this match what you'd expect to get?

### DIFF
--- a/src/gui/fields/TakePoint.tsx
+++ b/src/gui/fields/TakePoint.tsx
@@ -21,27 +21,69 @@
 import React from 'react';
 import {FieldProps} from 'formik';
 import Button, {ButtonProps} from '@material-ui/core/Button';
-import {Plugins} from '@capacitor/core';
+import {Plugins, GeolocationPosition} from '@capacitor/core';
+
+import {FAIMSPosition} from '../../datamodel/geo';
 
 const {Geolocation} = Plugins;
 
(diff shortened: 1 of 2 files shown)


stevecassidy:
Looks good to me. You have just used Feature which seems to be entirely correct. The openlayers GeoJSON exporter seems to insist on using a FeatureCollection - I should work out why and get rid of it if I can.

src/gui/fields/TakePoint.tsx
```diff
@@ -21,27 +21,69 @@
...
+      timestamp: timestamp,
+      altitude: position.altitude ?? null,
+      speed: position.speed ?? null,
+      heading: position.heading ?? null,
+      accuracy: position.accuracy,
+      altitude_accuracy: position.altitudeAccuracy ?? null,
+    },
+    geometry: {
+      type: 'Point',
+      coordinates: [position.latitude, position.longitude],
```
Denubis, line 47: https://datatracker.ietf.org/doc/html/rfc7946#appendix-A.1 It's a trap (that I keep falling in. Order is long/lat (x,y)

Denubis:
Other than comment above, looks good to me. We're not using a holding collection, which makes shapefile export easier. Well done!

src/gui/fields/TakePoint.tsx
aragilar: Fixed

PR #365, opened 2022-01-06
Adding , 2022 to all licenses

Denubis:
GNU suggests that https://www.gnu.org/licenses/gpl-howto.html a comma-delimited set of dates is the way to go here. (Given that copyright law means endless copyright, there are arguments out there to have no date, but... eh)


aragilar:
This'll need a rebase to due to the new sync changes touching everything

Denubis:
Ok. Will do this ASAP.

PR #367, opened 2022-01-07
Upgrade faims3-map-input to 1.0.5 and use imported settings

stevecassidy:
Settings code has been moved to the plugin. Update to plugin also fixes other issues - adds a value display, uses previous form state, more useful label.


aragilar:
Rebased to fix package-lock.json conflict.

bors r+

PR #384, opened 2022-01-17
Fix syncing

aragilar:
First we point the datadb listener to datadb, rather than metadatadb. Next we make sure to call the project listener callback at least once, so changes are watched.


aragilar:
Don't rebase this.

PR #408, opened 2022-01-25
Fully connect token to pouch

aragilar:
This adds some logging to confirm that pouch is indeed sending the token, and refreshes the connection when the user receives/updates the token.


aragilar:
I realise this shouldn't be merged until we have the key ids sorted on couchdb/conductor.

PR #409, opened 2022-01-25
a fix for photo scaling in take-photo, and the editing toggle

Denubis:
This should scale the photo taking thing correctly, as well as reduce a click from the android version of "editor"

There are bugs still about overwriting the field value instead of appending to it for multiple photos, as well as work needed to lightbox and be able to delete (and annotate) photos -- but that's for later.


KateSHENG (approved):
.

PR #476, opened 2022-04-01
Fixing image aspect-ratio scaled 'thumbnail' for TakePhoto

Denubis:
@aragilar ok, I think I fixed the aspect ratio regression reported by client in FAIMS3-390. 

However, I'm going to keep tranche1-fixes separate from mainline, since we don't have time to test all of the merges you did yesterday. (So... don't delete this branch when you merge?)

Signed-off-by: Brian Ballsun-Stanton <Denubis@users.noreply.github.com>


aragilar:
@Denubis: @KateSHENG should be the one to review this, changing reviewers.

PR #489, opened 2022-04-14
Add filtering based on user

aragilar:
This will control the visibility of projects and records, there will need to be some (temporary) adjustments to the conductor to set up the couchdb roles, otherwise you won't see anything but your local projects (this will be draft until those are in the conductor).


aragilar:
Undrafted as https://github.com/FAIMS/FAIMS3-conductor/pull/103 has been merged.

PR #509, opened 2022-04-28
Add routing needed for merging

aragilar:
This adds a dummy page which will have the merging interface, and
modifies the routing and links so that users get directed to it when
needed.


Denubis:
@aragilar is this still needed?

PR #526, opened 2022-05-12
Android manual testing deploy to dev, testing; integrate fastlane

Denubis:
* Building dev and testing targets for manually built APKs
* Integrating prod `android-fastlane` changes into main so that testing uses prod's configuration.


Denubis:
Ok, I need to override the failing test because it's using the one from main, not the one I'm committing. @aragilar is this ok?

PR #529, opened 2022-05-12
Reflect sync changes in active_db

aragilar:
This should allow preservation of the syncing state across reloads (but
not wipes). This may cause some issues with tracking how many notebooks
are active for a listing, but that will only be an issue
post-autoactivate.


KateSHENG:
bors+

KateSHENG:
bors+

KateSHENG:
bors+

PR #587, opened 2022-06-01
Avoid using bind with evented hook

aragilar:
It looks like the hook with bind may be leaking memory, use a new
wrapper to avoid needing bind. Hopefully this change appears in fuite.

Don't merge until this has been tested in fuite.


aragilar:
I couldn't see much difference in fuite, this will need more testing.

PR #606, opened 2022-06-09
Add @capacitor/app to handle android back button

aragilar:
I want to do some quick prodding on apple to check iOS information related to the plugin, but @Denubis did you want to take a quick look and confirm that this adds the correct behaviour (it's implied by the documentation that this should fix things, but this plugin also gives us additional control in case it not sufficient/we want to change things slightly).

This should fix FAIMS3-472.


aragilar:
Ok, no changes to iOS from xcode it seems.

PR #607, opened 2022-06-09
Bump map plugin version to 1.0.6

stevecassidy:
New version uses MUI and Capacitor 3.


Denubis:
@aragilar are we safe to merge this in? @stevecassidy any ideas on a testing protocol for this that @riniangreani can execute?

stevecassidy:
This change doesn't modify the plugin at all*, just upgrades the libraries. It should operate as before.  Testing protocol should be to have one field for each type of input (point, line, polygon) and check that the required data is captured. 

*except that it fixes a but where the field label was ignored, now it becomes the label on the button.

PR #622, opened 2022-06-20
Fix grid spacing (FAIMS3-328)

Denubis:
@KateSHENG This is what I was talking about in terms of grid layouts of FAIMS3-328


Denubis:
Merging, since I don't think @KateSHENG monitors commits

PR #701, opened 2022-08-02
Fix 530

KateSHENG:
FAIMS3-530: fix padding and rebvert tab to steppers and fix padding for steppers


Denubis:
@KateSHENG I'm happy with these changes (now that I've caught the last nopadding stragglers). Safe to commit to main?

PR #710, opened 2022-08-22
Add user switcher to login page

aragilar:
This is somewhat simple for now, we likely want to allow users to
"logout" without switching, but we don't expect users to be logging out
of the device that much.


aragilar:
Merging because license error is unrelated...

PR #716, opened 2022-08-25
Add a button to dump the contents of all pouchdbs

aragilar:
This does not try to do things with visibilities, it simply dumps all
the databases we know about using allDocs.

### DIFF
--- a/src/gui/pages/about-build.tsx
+++ b/src/gui/pages/about-build.tsx
@@ -23,6 +23,8 @@ import {Box, Container} from '@mui/material';
 import Button from '@mui/material/Button';
 import * as ROUTES from '../../constants/routes';
 import {unregister as unregisterServiceWorker} from '../../serviceWorkerRegistration';
+import {downloadBlob} from '../../utils';
+import {getFullDBSystemDump} from '../../sync/data-dump';
 import {
   DIRECTORY_PROTOCOL,
   DIRECTORY_HOST,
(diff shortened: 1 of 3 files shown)


stevecassidy:
This in itself will be useful to help with recovery if the app is responding.   Perhaps we could add a configuration option to dump to a known location on a regular basis as well - overwriting previous dumps or rotating.   That would then be useful for cases like the Farmer Incident where the app is not responsive.

We could also do with a restore option to be able to load this into a running app to reproduce problematic instances.  While we can trawl through this dump to locate things it would be most useful to restore and view records that way.

aragilar:
Restoring the current dump to a phone is much harder than dumping, as we'd need to work out the conflicts for all the databases. I would do a backup/restore at the project level, and only do the data, as then conflicts can be managed by the user using the merging tools.

stevecassidy:
I was more thinking of a 'debugging restore' where we just overwrite whatever is there to restore the dumped state in the hope of reproducing whatever issue is being debugged.  Definitely not a user level feature.

Denubis:
For the backup, so long as it outputs some sort of file (bonus points with a username and datestamp in it) that they can pass to me for restores, that will allow us to solve Adela's current issues.

stevecassidy:
Well this PR would give us what you need @Denubis and is a step forward.  We should merge this in and then work towards some restore functionality in future after working out what it should look like.

Denubis:
Ok, merging in. I'll also backport this to the tag.

bors r+

PR #718, opened 2022-08-29
This switches minifauxton to use a local file

aragilar:
We'll need to come up with a sensible way to keep the file up-to-date.


Denubis:
bors r+

(I think we might have to redo this commit, as it's picking up the old license files and causing a timeout)

PR #733, opened 2022-09-08
Possible minifauxton fix

aragilar:
I think this was likely working around a issue on a specific OS/browser, not sure which one though.


Denubis:
This workaround is because chrome, when offline, will not load minifauxton.

PR #770, opened 2022-10-17
Merge app UI child and implement link relation

KateSHENG:
Merge app UI child and implement link relation: link relation still working in progress, App-ui is merged


lmannering (approved):
Moving work on the UI onto main ready for testing.

PR #773, opened 2022-10-19
Wire up sync status lights to sync code

aragilar:
This should mean the sync lights blink.


aragilar:
It looks like pouchdb isn't working on the latest bugfix LTS version (earlier LTS version and newer node versions work). Merging and leaving fixing that to later...

PR #791, opened 2022-10-25
User switcher fix

lmannering:


lmannering:
@aragilar do you need to push any further commits to this? otherwise I'll add the refresh and merge in

PR #831, opened 2022-11-14
Revert "Add prefer label"

Denubis:
Reverts FAIMS/FAIMS3#830. This patch and the previous have caused WSOD. See FAIMS3-650


Denubis:
I was incorrect. It was last week's activation that caused universal WSOD. No one tested over friday or today.

PR #848, opened 2022-11-17
fix: using the relation_preferred_label to control visbility of prefe…

Denubis:
…rred checkbox in child relationsships


Denubis:
@lmannering can you manage the codereview for this one, please? I'm not sure if @KateSHENG sees these requests.

PR #853, opened 2022-11-21
GeoTiff Basemap Attachments

stevecassidy:
Adds support for using geoTiff file attachments as basemaps in the map form field.  

Signed-off-by: Steve Cassidy <steve.cassidy@mq.edu.au>


stevecassidy:
One remaining bug to fix....selecting default basemap causes a crash. Probably fix on the plugin side to handle the 'default' option for the geoTiff property but will need to update the version imported here.

PR #870, opened 2022-11-25
Fix status typo

aragilar:
It appears that the CSS for the toast crashed due to the use of 'warnings' rather than 'warning'. This does not fix the issue causing the data to be unable to be loaded.


aragilar:
Fixes FAIMS3-691.

PR #887, opened 2022-11-30
Initial attempt for streaming dump

aragilar:
This tries to stream JSON into a blob on browsers, and to the file system on apps. On browsers this may use large amounts of memory, but absent an easy way to stream to the file system, we don't have many options.


aragilar:
See FAIMS3-547, FAIMS3-682, FAIMS3-695.

PR #989, opened 2023-09-07
Fix bug that prevents the record edit view showing

stevecassidy:
Problem with view initialisation that prevented it finding out what the viewname and viewset are for the record.


stevecassidy:
cap doctor is failing due to a known issue: https://github.com/ionic-team/capacitor/issues/6713

PR #1014, opened 2024-02-06
Fix child record navigation

stevecassidy:
When we finish editing a child record we should pop back to the parent rather than the notebook.


stevecassidy:
bors+

PR #1049, opened 2024-05-08
Upgrade Capacitor to 6

stevecassidy:


stevecassidy:
Also upgrades barcode scanning to use MLKit since the old version isn't compatible with Cap 6.

PR #1081, opened 2024-07-22
Monorepo initial setup

luke-mcfarlane-rocketlab:
Still ironing out a few issues with running some scripts and was not able to get the live reload working


stevecassidy (approved):
All looks good, tests running and passing in all sub projects. A few things to sort out in future PRs but very workable.

PR #1088, opened 2024-07-26
Ensure different known ports for designer and app

stevecassidy:
Both `app` and `designer` use vite and start up on random ports. Fix the ports at 5000 for the app and 5010 for the designer so we don't get confused.


luke-mcfarlane-rocketlab (approved):
Tested & LGTM

PR #1093, opened 2024-07-29
feat: BSS-158 add AWS deployment using CDK

PeterBaker0:
# feat: BSS-158 add AWS deployment using CDK

## JIRA 

[BSS-158](https://jira.csiro.au/browse/BSS-158)

## Description 

Adds a new project to the mono repo under `infrastructure/aws-cdk`. This is a CDK typescript project. CDK is an AWS specific infrastructure as code tool which allows you to use fully type checked code to deploy infrastructure programmatically. 

The following infrastructure is deployed for FAIMS3 

- CloudFront distributions and S3 buckets for front-end components (designer and faims3 app)
- ECS Fargate cluster with currently single node for Conductor API fronted by Application Load Balancer
- EC2 auto scaling group for CouchDB
- AWS Secret manager for Cookie secret, DB user/pass and public/private keypair for JWTs in Couch/Conductor
- AWS Cloud map with automatic registration and deregistration to future proof clustering needs for CouchDB

## CouchDB details

Deployed on EC2 in Auto Scaling Group. Fronted by Application Load Balancer (ALB). The default configuration and admin user/pass is included by pulling the value from AWS Secrets Manager at instance start up. This is visible in the user data script which is deployed to the instance. 

The DB is then initialised as per the conductor initialise endpoint.

## Conductor details 

Deployed as ECS cluster on ECS Fargate. Could easily be setup to scale given stateless operation. Fronted by Application Load Balancer.

## Front end details

S3 deployment used into bucket which auto deletes on destroy. CloudFront distributions have automatic invalidation when source asset hash changes. HTTPS setup.

## Lambda packages

There will likely be miscellaneous lambda packages to deploy. To be consistent, going to keep using typescript for these. I prefer building these outside of CDK so that we can apply normal build optimisations and caching. This required some hacking of tsconfig to get these packages to not consider the parent workspace context. You can build and prep AWS for deployment by running `npm run build-aws` from top level, or `cd infrastructure/aws-cdk` and `npm run build`. Similarly to package the production `dist` run `package` instead of build in the above commands. 

## Configuration setup 

See README in aws-cdk project for details on the configuration setup. TLDR config is managed separately from main repo for security purposes (BSS team see faims-config repo in private org). This is validated at CDK run time and can be dispatched safely without committing any files. 

### Key management for AWS

Previously conductor relied on mounted local files to get keys. This is hard to manage in an automated cloud deployment or a deployment where the keys could be dynamic, because it would require a) persistent mounted storage for a stateless API and b) the ability to redeploy all file mounts (meaning complete redeployment) when keys change. 

Changed approach by writing updated key generation script which generates the pub/priv key pair and uploading to AWS Secret Manager. The bash script outputs the ARN of the secrets. 

There are now two new env variables in conductor

- KEY_SOURCE: the protocol to use to retrieve keys, FILE is default and retains current behaviour, AWS_SM is the preferred for AWS 
- AWS_SECRET_KEY_ARN: (required if using AWS_SM source) the ARN to the private key secret in AWS SM 

I have followed a service dependency injection pattern to handle keys now, including

- an abstract base class for the key service
- an implementation for reading from local files as per previous implementation
- an implementation for AWS SM including caching with a configurable timeout
- a factory function which takes key source and builds the appropriate service

This means we can easily extend this in the future for other key sources, and safely maintain both implementations to support existing deployments, or for local usage. We could also support other cloud providers in the future. 

### Conductor initialise endpoint extended

The existing `/api/initialise` endpoint performs DB initialises in an idempotent fashion. Followed this pattern to add the `jwt_keys` configuration. This uses the above configured key source to tell the DB which public key to trust. This means the couchDB can be entirely configured from the API other than the admin username and password which is injected at instance startup from AWS SM.

### Dockerfile changes for API

I changed the Dockerfile for API so that it would work for me. This mainly involved building and packaging the data models library and changing the entrypoint.

## Potentially breaking changes made

- changed build output folder for FAIMS3/designer apps to `build` to be consistent
- changed Dockerfile for API - if other build tools using this it might cause problems 

## TODO remaining (moving to JIRA to be tackled later)

- [x]  refactor user data handling of local.ini file in the docker setup - maybe store?
- [x]  optimise docker build for conductor API
- [x]  deploy CDK AWS from dockerhub rather than CDK asset
- [x]  consider EBS recovery method given ASG for EC2
- [x]  add .env file for conductor, FAIMS3 etc to private config repo
- [x]  Make the below nicer i.e. key source enum - now created enum but source properly?

```
      // The AWS Secret Manager secret ARN for the public/private RSA key
      // pair (not secure value - protected by AWS retrieved at runtime)
      AWS_KEY_SECRET_ARN: props.privateKeySecretArn,
      // TODO
      KEY_SOURCE : "AWS_SM"

```

- [x]  remove compile step for running - should just run straight with node? This could slow down scaling launch - if compilation is necessary then put into docker step
- [x]  consolidate to single ELB for conductor and ASG to save $$$
- [x]  NAT gateway and private subnets when prod mode for AWS deployment
- [ ]  tighten permissions for the lambda deregister function
- [ ]  setup google auth for AWS deployment
- [ ]  setup email for AWS deployment
- [ ]  configure auto scaling for Conductor service
- [ ]  add domain names to configuration process
- [ ]  handle android and ios app URLs in configuration
- [ ]  one click deploy including key setup and API init
    - [ ]  could use custom resource
- [ ]  add instance size to configuration
- [ ]  implement clustering for couch DB - static cluster


luke-mcfarlane-rocketlab (approved):
LGTM - have tested the run/build scripts for api/app/designer and all appear to still work fine.

Have not tested the AWS deployment logic but have tested the deployment.

PR #1103, opened 2024-08-02
feat: Github Actions workflow for CDK operations

PeterBaker0:
# feat: Github Actions workflow for CDK operations

## Description

Implements a template github actions workflow for deploying the CDK application using the specialised organisation config.

The intention is for this not to be recognised as a legitimate workflow, and instead used as a basis for organisation + environment specific deployments e.g. for the BSS.


PeterBaker0 (commented):
Moving through as just actions stuff

PR #1107, opened 2024-08-06
BSS-225 Survey List Page - Update tabs to long list

luke-mcfarlane-rocketlab:
### JIRA
[BSS-225](https://jira.csiro.au/browse/BSS-225)

### Description
Added the ability to switch between tabs and a long list for the survey list page using a config file in the constants folder. This means future deployments can decide which layout to use rather than being forced into the new layout.

### How to test

- run app
- navigate to a notebook list
- update `NOTEBOOK_LIST_TYPE` in `app/src/constants/config.tsx` to "headings"
- verify the layout has changed


stevecassidy (approved):
Happy with these changes. Only comment is that your PR description mentions the wrong config file now.

PR #1113, opened 2024-08-14
Dev docker compose build fixes to allow local development

PeterBaker0:
Updating dev docker compose workflow to include full necessary build context and improving cd commands to not mask missing directories. Including env files in the docker compose for each service. Updating README accordingly.

**Note** I have not used the /src mount thing before, so unsure how that works? @luke-mcfarlane-rocketlab if you use that you might want to test.


ranisa-gupta16 (commented):
Looks good to me!

luke-mcfarlane-rocketlab:
LGTM - tested and resolves the issue with docker build

luke-mcfarlane-rocketlab (approved):

PR #1116, opened 2024-08-15
build: turbo cache, generic build dockerfile, dependency caching in github actions

PeterBaker0:
Makes various build improvements

- utilises turbo repo remote cache in build and lint pipeline
- produces an automated Dockerfile.build which is capable of building any node based target in the monorepo without specifying manually the dependencies to bundle in the output
- updates README to explain Dockerfile.build
- utilises `npm i` cache in the Github Actions


luke-mcfarlane-rocketlab:
LGTM - tested locally, and reviewed actions output

luke-mcfarlane-rocketlab (approved):

stevecassidy (approved):
This all looks good. My only comment is tangential - I updated the node version in a few places to 20.9.0 for an earlier PR but note that you're just specifying node 20 in a few places here.  That sounds like a better idea so should we use a less specific version in the engines clause in package.lock?

PR #1117, opened 2024-08-19
Fix/directory api server details

stevecassidy:
# fix/directory API server details

## JIRA Ticket

[BSS-269](https://jira.csiro.com/browse/BSS-269)

## Description

Completes some work that was part of the startup refactor by adding an API endpoint to give descriptive details of the server for display in the login panel. 

## Proposed Changes

- Add an API endpoint `/api/info` to return the name and description for the server from the config
- Update the app to request this info for the server on startup before displaying the login panels
- Display the retrieved name and description for a server in the login panel
- Enable multiple servers by allowing VITE_CONDUCTOR_URL to have more than one URL separated by spaces
- Add the name of the server to the notebook listing to disambiguate
- Update initJWTKeys to not try accessing couchdb if running under test
- Provide different keys for the notebook datagrids to turn off a React warning
- Remove some old config variables and make sure they aren't referenced

## How to Test

Configure `api/.env` to add:

```bash
# Public name for this instance
CONDUCTOR_INSTANCE_NAME="Development FAIMS Server"
CONDUCTOR_DESCRIPTION="Development server on localhost"
```
Rebuild and deploy the API via docker compose. 

Configure `app/.env` to add another server, eg:

```bash
VITE_CONDUCTOR_URL=http://localhost:8154 https://conductor.testing.fieldmark.app
```
Run the app.  You will need to reset local storage (in Chrome dev tools go to Application and click "Clear Site Data") this will force the app to recreate the database entries for server listings.   You should then see two servers on the login page.  You should also see your configured server name and description for your local server. 

The older server will get a default name of the hostname `conductor.testing.fieldmark.app` in this case. 

## Additional Information

Noting that we don't have a description of the configuration 

## Checklist


api/src/api/routes.ts
```diff
@@ -75,6 +78,18 @@ api.post('/initialise/', async (req, res) => {
...
 
+/**
+ * Handle info requests, basic identifying information for this server
+ */
+api.get('/info', async (req, res) => {
+  const info = {
+    name: CONDUCTOR_INSTANCE_NAME,
+    url: CONDUCTOR_PUBLIC_URL,
+    description: CONDUCTOR_DESCRIPTION,
+  };
```
PeterBaker0, line 89: Can you type this interface in the data models package so both app and API can reference it?

api/src/authkeys/initJWTKeys.ts
```diff
@@ -3,9 +3,12 @@ import {
   COUCHDB_INTERNAL_URL,
   KEY_SERVICE,
   LOCAL_COUCHDB_AUTH,
+  RUNNING_UNDER_TEST,
 } from '../buildconfig';
 
 export async function initialiseJWTKey(): Promise<void> {
+  // don't try to do this if we're testing
+  if (RUNNING_UNDER_TEST) return;
```
PeterBaker0, line 11: Might be good to log this skip in case someone misconfigures the API and is confused as to why keys are not being setup appropriately. Do we have logging with levels configured? This could be a debug log.

api/test/api.test.ts
```diff
@@ -98,6 +101,17 @@ describe('API tests', () => {
...
+    return request(app)
+      .get('/api/info')
+      .expect(200)
+      .expect(response => {
+        expect(response.body.name).to.equal(CONDUCTOR_INSTANCE_NAME);
+        expect(response.body.description).to.equal(CONDUCTOR_DESCRIPTION);
+        expect(response.body.url).to.equal(CONDUCTOR_PUBLIC_URL);
+      });
+  });
+
```
PeterBaker0, line 114: cool

app/README.md
```diff
@@ -1,89 +1,5 @@
...
-
-1. Build the source code
-   - `npm run webapp-build`
-1. Synchronise Gradle files
-   - `npm run webapp-sync` OR `cap sync`
-1. [Optional] Allow to copy to /Library/Ruby/Gems/2.3.0:
-   - `export GEM_HOME="$HOME/.gem"`
-1. [Optional] Resolve `xcode-select` error
-   - `sudo xcode-select -s /Applications/Xcode.app/Contents/Developer`
-1. Build the apk from the build/build bundle/build apk section
```
PeterBaker0, line 89: Was there valuable stuff here?

app/src/buildconfig.ts
```diff
@@ -288,12 +246,12 @@ function get_bugsnag_key(): string | false {
...
 }
 
-function get_conductor_url(): string {
-  const url = import.meta.env.VITE_CONDUCTOR_URL;
-  if (url) {
-    return url;
+function get_conductor_urls(): string[] {
+  const urls = import.meta.env.VITE_CONDUCTOR_URL;
+  if (urls) {
+    return urls.split(' ');
```
PeterBaker0, line 252: space is risky separator due to trailing whitespace - would propose comma as a more standard separator, and trim/strip each split value for whitespace, filtering out empty values.

app/src/buildconfig.ts
```diff
@@ -288,12 +246,12 @@ function get_bugsnag_key(): string | false {
...
   } else {
-    return 'http://localhost:8154';
+    return ['http://localhost:8154'];
```
PeterBaker0, line 254: Would be good to log that this fall through has occurred and put this as a constant somewhere in the build config rather than hardcoding here.

app/src/gui/components/ui/heading-grid.tsx
```diff
@@ -32,7 +32,7 @@ export default function HeadingGrid({
         Active
       </div>
       <DataGrid
-        key={'notebook_list_datagrid'}
+        key={'active_notebook_list_datagrid'}
```
PeterBaker0, line 35: nice

app/src/sync/process-initialization.ts
```diff
@@ -44,23 +44,13 @@ import {ensure_project_databases} from './projects';
 
 // called on startup to get the initial set of projects
 export async function update_directory() {
-  let listings = await getAllListings();
+  const listings = await getAllListings();
```
PeterBaker0, line 47: This code is non-obvious what's going on. From memory you are building out a local version of the directories database by fetching the details from the API? Would be good to have function doc string and inline comments here.

app/src/sync/process-initialization.ts
```diff
@@ -69,6 +59,35 @@ export async function update_directory() {
   ensureActiveProjects();
 }
 
+/**
+ * Create and store a listings object for this conductor instance
+ *
+ * @param url - URL of the Conductor instance
+ * @returns A Listings object
```
PeterBaker0, line 66: This is where that shared interface would come in handy?

app/src/sync/process-initialization.ts
```diff
@@ -69,6 +59,35 @@ export async function update_directory() {
...
+    .then(response => response.json())
+    .then(data => {
+      listing.name = data.name;
+      listing.description = data.description;
+    })
+    .catch(() => {
+      listing.description = 'No Description';
+    });
+
+  directory_db.local.put(listing);
```
PeterBaker0, line 87: What happens if this function hangs or errors? noting that we've lost the benefits of async in this code.

PeterBaker0 (changes requested):
Looks good but suggest some improvements to improve documentation, type safety - and just raising some questions around the handling of the async logic when loading notebook details. I'm also curious about when this listing is refreshed.

app/README.md
stevecassidy: That's all out of date. We do need to replace it but having it there is just confusing.

app/src/buildconfig.ts
stevecassidy: Hmm, I had a comma but I thought they are valid in a URL I chose space instead, seems like I was wrong, comma is not a valid char in a URL so we can use that instead.

api/src/api/routes.ts
stevecassidy: Modified slightly to fit the existing ListingInformation type.

app/src/buildconfig.ts
stevecassidy: Logging added.

app/src/sync/process-initialization.ts
stevecassidy: done

api/src/authkeys/initJWTKeys.ts
stevecassidy: This isn't a config setting, should only be true when tests are running. I think it's safe.

api/src/api/routes.ts
PeterBaker0: See my commit to address this - my comment was to suggest that the API should explicitly annotate the model that is being returned which is shared between the App and API. Currently, someone could come along and change 'name' to 'label' and there would be no warning anywhere.

api/src/authkeys/initJWTKeys.ts
PeterBaker0: Okay

api/src/api/routes.ts
PeterBaker0: Sorry realise you hadn't committed your updates yet. I can revert if needed.

PR #1126, opened 2024-08-21
fix: default route redirect

luke-mcfarlane-rocketlab:
## JIRA Ticket

[BSS-301](https://jira.csiro.au/browse/BSS-301)

## Description

App shows blank 404 page when there is an issue.

## Proposed Changes

- removed 404 component route (this component is redundant as it has no path)
- added default route that redirects home

## How to Test

- log into app (if necessary) and navigate to a notebook (survey)
- refresh the page and verify that it navigates home
- (optional) type garbage into the url (e.g. http://localhost:5000/abc123) and verify it navigates home

## Additional Information

This does not address the underlying issue but improves the user experience if something were to go wrong in production.

## Checklist

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -48,6 +48,7 @@ import {getTokenContentsForCurrentUser} from './users';
 import {useEffect, useState} from 'react';
 
 import {TokenContents} from '@faims3/data-model';
+import NotFound404 from './gui/pages/404';
 
 // type AppProps = {};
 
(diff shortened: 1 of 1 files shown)


PeterBaker0 (commented):
Functionally looks fine - is this standard practise for 404 behaviour to redirect to '/'? Maybe we have a 404 page which has a button "return home" or similar so at least they can take a screenshot to show the issue before it disappears?

luke-mcfarlane-rocketlab:
> Functionally looks fine - is this standard practise for 404 behaviour to redirect to '/'? Maybe we have a 404 page which has a button "return home" or similar so at least they can take a screenshot to show the issue before it disappears?

Agree with this apporach - has been resolved

stevecassidy (approved):
Looks fine.

PR #1129, opened 2024-08-25
Feat: Add address field

stevecassidy:
# feat: Add Address Field

## JIRA Ticket

[148](https://jira.csiro.au/browse/BSS-148)

## Description

Adds a new field type for addresses. 

## Proposed Changes

- [x] Add new field type `Address`.  This first version is a simple entry of the address fields with no auto completion etc.  
- [x] Add documentation on how to add new fields to FAIMS

## How to Test

[Sampler.json](https://github.com/user-attachments/files/16744238/Sampler.json)
 is a sample notebook that contains an address field.  Upload this to your development server to test the new field in your app. 

## Additional Information

## Checklist


docs/developer/Fields.md
luke-mcfarlane-rocketlab, line 1: Nice! This is super helpful

app/src/gui/components/footer/index.tsx
```diff
@@ -70,7 +70,7 @@ export default function Footer(props: FooterProps) {
 function DevelopTool() {
   const globalState = useContext(store);
   return (
-    <Grid container spacing={2}>
+    <Grid container spacing={2} sx={{width: '100vh', overflow: 'scroll'}}>
```
luke-mcfarlane-rocketlab, line 73: good find!

luke-mcfarlane-rocketlab (approved):
All good from my end - a couple of minor comments.
- mostly a matter of taste but I generally try to avoid having anything the user can interact with (such as buttons) move after the initial load. I would put the pencil icon to the left of the concatenation or to the right of the field name.
- the two commas appear in the concatenated address regardless of which fields are empty which for example results in ", , NSW" when only entering state. If this is intended, can ignore.

PR #1130, opened 2024-08-28
fix: missing conductor_url in aws code and improving handling of error/warn conditions

PeterBaker0:
# fix: missing conductor_url in aws code and improving handling of error/warn conditions

## JIRA Ticket

[BSS-224](https://jira.csiro.au/browse/BSS-224)

## Description

Fixes missing environment variable in conductor API configuration for AWS deployment. 

Refactors default conductor URLs to constants in the app and api code. 

Adds doc strings for conductor URL split method and handles error conditions more durably including filtering out empty values which could result from trailing commas, and warning under either empty string or empty list conditions. Falls through to the default value if the length of the list after splitting and filtering out empty values is zero, such as in the case where the environment variable is an empty string. 

## How to Test

1. Testing changes to conductor URL splitting logic: I added unit tests for this function. See `cd app && npm i && npx vitest buildconfig`.

2. Testing AWS deployment - not on critical path right now but once merged I'll update the downstream fork and redeploy and ensure it works. 

## Checklist

### DIFF
--- a/api/src/buildconfig.ts
+++ b/api/src/buildconfig.ts
@@ -30,6 +30,11 @@ const FALSEY_STRINGS = ['false', '0', 'off', 'no'];
 export const CLUSTER_ADMIN_GROUP_NAME = 'cluster-admin';
 export const NOTEBOOK_CREATOR_GROUP_NAME = 'notebook-creator';
 
+// Constants
+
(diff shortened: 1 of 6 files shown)


stevecassidy (approved):
Looks ok.

PR #1140, opened 2024-09-05
Feat/enhance-registration-login-flow

stevecassidy:
# Feat: Enhance Registration and Login Flows

## JIRA Ticket

[BSS-313](https://jira.csiro.com/browse/BSS-313)

## Description

Overall goal is to provide a more seamless registration workflow for new users invited to a notebook.  Will also try to improve the overall login experience.

## Proposed Changes

- [x] Modify the login flow to use redirects back to the app rather than pop-up windows (on web)
- [x] Use redirect on mobile back to a custom URL scheme to make login more seamless
- [x] Make registration use the redirect scheme as well so that it can be used from in app
- [x] Change registration invite ids to short codes - six character codes with a fixed prefix (BSS-123-ABC) so that they can be entered by hand
- [ ] Invite codes should be fixed for each role in the notebook rather than being generated on demand
- [x] Add QR code scan option from in the app to register for a notebook
- [x] Add option to enter short-code in app to register for a notebook
- [x] Tidy up login for Conductor web interface so that it works with new redirect scheme

## How to Test

Add configuration to the api:

```javascript
CONDUCTOR_SHORT_CODE_PREFIX="DEV"
```

You'll need to be running both the api and the app to test. On the app do a 'Wipe and Reset Everything' from About Build to clear local databases.  

### Conductor

- Login to conductor should be as before but instead of being taken to the 'send-token' page you should go to the home page after login.    The return to app button is removed.

### Login/logout
 
- Go to the conductor (API) URL and make sure you are not logged in
- In the app, click 'Sign in" on the login page, you will be redirected to Conductor to login, once login is successful you will be redirected back and should be properly logged in
- Check that the username is shown in the top right. Note that the menu is now gone and the button just links back to the sign-in page
- Logout from the app, you may notice a quick redirect but it should log out as before, however now you are also logged out from conductor (visit the conductor URL to verify)

### Short code registration

- In Conductor, select a notebook and generate an invite, select a user role (eg. admin)
- Copy the short code that is generated, note the prefix (as configured above)
- Logout of conductor
- In the app, enter the six character short code (without prefix), click Register
- You should see the registration screen, enter account details to make a new account
- Once you submit, you should be redirected back to the app and (after a delay) the notebook should be available to you
- Enter a random string of characters and you should get an error page from conductor with a button to return to the app

### QR Code registration

This is only available on Android or IOS

- As before, generate an invite for a notebook, this time display the QR Code
- In the app, select 'Scan QR Code' from the sign in page
- Scan the QR code, it should take you to the registration page or if already logged in, just make the notebook available
- Scan a random QR code/barcode and you should get an error message
 
## Checklist


api/src/api/routes.ts
```diff
@@ -84,11 +85,12 @@ api.post('/initialise/', async (req, res) => {
  * Handle info requests, basic identifying information for this server
  */
 api.get('/info', async (req, res) => {
```
PeterBaker0, line 87: propose api.get<{}, ListingsObject> - moving forward good to type the methods.

api/src/auth_routes.ts
```diff
@@ -63,9 +64,35 @@ export function determine_callback_urls(provider_name: string): {
...
+  } else if (
+    redirect.startsWith('/') ||
+    redirect.startsWith('org.fedarch.faims3://')
+  ) {
+    return redirect;
+  } else {
+    return '/';
+  }
+}
+
```
PeterBaker0, line 87: Could you explain in the doc string what makes a redirect valid? E.g. what is org.fedarch.faims3? Is that something to do with the app?

api/src/auth_routes.ts
```diff
@@ -63,9 +64,35 @@ export function determine_callback_urls(provider_name: string): {
...
 export function add_auth_routes(app: any, handlers: any) {
```
PeterBaker0, line 88: Could you add a doc string for this function and type the inputs ideally?

api/src/auth_routes.ts
```diff
@@ -63,9 +64,35 @@ export function determine_callback_urls(provider_name: string): {
...
-  app.get('/auth/', (req: any, res: any) => {
-    // Allow the user to decide what auth mechanism to use
+  app.get('/auth/', async (req: any, res: any) => {
+    const redirect = validateRedirect(req.query?.redirect || '/');
+
+    if (req.user) {
+      await redirect_with_token(res, req.user, redirect);
+    }
+
     const available_provider_info = [];
```
PeterBaker0, line 96: some inline comments in this function would be nice - pretty hard to understand what is going on

api/src/auth_routes.ts
```diff
@@ -77,37 +104,87 @@ export function add_auth_routes(app: any, handlers: any) {
...
 
-  // handle local login post request
-  app.post(
-    '/auth/local',
-    passport.authenticate('local', {
-      successRedirect: '/send-token',
-      failureRedirect: '/auth',
-    })
-  );
+  app.get('/logout/', (req: any, res: any, next: any) => {
```
PeterBaker0, line 111: can we type these inputs and doc string.

api/src/auth_routes.ts
```diff
@@ -77,37 +104,87 @@ export function add_auth_routes(app: any, handlers: any) {
...
+    }
+    res.redirect(redirect);
+  });
+
+  const authenticate_return = (
+    req: any,
+    res: any,
+    next: any,
+    redirect: string
+  ) => {
```
PeterBaker0, line 129: typing here

PeterBaker0: and doc string too

api/src/auth_routes.ts
```diff
@@ -77,37 +104,87 @@ export function add_auth_routes(app: any, handlers: any) {
...
+      req.login(user, async (loginErr: any) => {
+        if (loginErr) {
+          return next(loginErr);
+        }
+        return redirect_with_token(res, user, redirect);
+      });
+    };
+  };
+
+  const redirect_with_token = async (
```
PeterBaker0, line 148: doc string/types

api/src/auth_routes.ts
```diff
@@ -77,37 +104,87 @@ export function add_auth_routes(app: any, handlers: any) {
...
+    user: Express.User,
+    redirect: string
+  ) => {
+    // Generate a token
+    const token = await generateUserToken(user);
+    // Append the token to the redirect URL
+    const redirectUrlWithToken = `${redirect}?token=${token.token}`;
+
+    // Redirect to the app with the token
+    return res.redirect(redirectUrlWithToken);
```
PeterBaker0, line 159: Is this a recommended way to do this re: embedding the token into the redirect as a query string - I guess it works but is it best practise? Probably okay but thought cookies might be used

api/src/auth_routes.ts
```diff
@@ -77,37 +104,87 @@ export function add_auth_routes(app: any, handlers: any) {
...
+  };
+
+  app.post('/auth/local', (req: any, res: any, next: any) => {
+    const redirect = validateRedirect(req.query?.redirect || '/');
+    passport.authenticate(
+      'local',
```
PeterBaker0, line 165: doc strings/types etc for this function - and what is the significance of local here?

app/android/app/src/main/assets/capacitor.config.json
```diff
@@ -6,10 +6,14 @@
...
 			"launchShowDuration": 0
+		},
+		"CapacitorHttp": {
+			"enabled": false
 		}
 	},
 	"server": {
-		"allowNavigation": []
+		"allowNavigation": [],
+		"iosScheme": "org.fedarch.faims3"
```
PeterBaker0, line 16: I see this value again - how do we manage this for other orgs/deployments?

PeterBaker0 (changes requested):
New functionality sounds good. 

I'm out of time to fully review - just noting that the authentication flow logic in the API is hard to follow due to a lack of docstrings/types/documentation. 

Could we add doc strings, inline comments explaining the steps, and ideally even a diagram explaining how the various  login flows/redirects work.

Will test when I'm back next week.

api/src/auth_routes.ts
stevecassidy: This is the standard way to pass a token. Cookies won't work as we're redirecting back to another app/domain. Need to send the token with the request URL.

app/android/app/src/main/assets/capacitor.config.json
stevecassidy: It occurs in code and in config files so we would need to work out how to get the right values in place. Will raise a ticket for this.

stevecassidy: https://jira.csiro.au/browse/BSS-344

PeterBaker0:
image
This is what I see

"RoleShort Code" ? Maybe don't show existing invites if there are none? Or placeholder.

PeterBaker0:
I noticed when I click "Copy URL" it copied "http://localhost:8080/register/DEV-4NRR33" which is my conductor instance - is this supposed to redirect to the app instead? Or am I misunderstanding?

PeterBaker0:
> I noticed when I click "Copy URL" it copied "http://localhost:8080/register/DEV-4NRR33" which is my conductor instance - is this supposed to redirect to the app instead? Or am I misunderstanding?

Oh I see - if you're not logged in it takes you to register.

PeterBaker0:
image

When I land on the workspace page - sometimes there is no workspace shown until I refresh page. In this condition the above renders - i.e. there is no where to input code?

PeterBaker0:
This will be a common UX issue
image

If the user includes the prefix, it doesn't work as prefix is duplicated - can we strip this out if prefix matches.

PeterBaker0:
image
Copy short code would be nice here.

PeterBaker0:
After adding a new survey from short code for existing user I got here

image

Notice 2 available, only one in the list - I think this might be due to @luke-mcfarlane-rocketlab stuff with context?

api/public/swagger.json
PeterBaker0, line 1: Nice - thanks for updating

api/src/auth_routes.ts
```diff
@@ -28,6 +28,19 @@ import {registerLocalUser} from './auth_providers/local';
...
+interface RequestQueryRedirect {
+  redirect: string;
+}
+interface PostRegisterRequestBody {
+  username: string;
+  password: string;
+  email: string;
+  repeat: string;
+  name: string;
+}
```
PeterBaker0, line 44: If these are returns/inputs to API - should go in data model package with the other api.ts interfaces?

api/src/auth_routes.ts
```diff
@@ -63,9 +76,57 @@ export function determine_callback_urls(provider_name: string): {
...
+  // In the following generic type signatures for routes the
+  // order of types is:
+  //  <Params, ResBody, ReqBody, ReqQuery, Locals>
+  // Params: Route parameters
+  // ResBody: Response body
+  // ReqBody: Request body
+  // ReqQuery: Query string parameters
+  // Locals: Response local variables
+
+  app.get<{}, {}, {}, RequestQueryRedirect>('/auth/', async (req, res) => {
```
PeterBaker0, line 123: Once we merge new survey work, will be a task to come back and align the validation and typing approach - but appreciate adding the doc strings and interfaces. 👍

stevecassidy:
I think some of these are due to refresh issues hopefully fixed in Luke's PR.  Will add the copy short code link and do some validation on entered short codes.

api/src/auth_routes.ts
```diff
@@ -77,69 +138,164 @@ export function add_auth_routes(app: any, handlers: any) {
...
+        );
+        res.render('register', {
+          invite: invite_id,
+          loginURL: `/auth?redirect=${encodedRedirect}`,
+          providers: available_provider_info,
+          redirect: redirect,
+          localAuth: true, // maybe make this configurable?
+          messages: req.flash(),
         });
       }
```
PeterBaker0, line 280: Just checking that it's possible to login from the register page. The workflow here would be 1) open app 2) app is logged out for some reason 3) accept code 4) redirected to register, but user already has account.

api/src/auth_routes.ts
```diff
@@ -169,44 +327,49 @@ export function add_auth_routes(app: any, handlers: any) {
...
-            // res.redirect('https://auth.datacentral.org.au/cas/logout');
-            //throw err ?? Error('Authentication failed (next, no error)');
-          }
-        );
+        passport.authenticate(
+          handler + '-validate',
+          authenticate_return(req, res, next, redirect)
+          //          HANDLER_OPTIONS[handler]
+        )(req, res, next);
+        console.log('TODO: may need to insert:', HANDLER_OPTIONS[handler]);
```
PeterBaker0, line 372: is this okay?

api/src/authkeys/create.ts
```diff
@@ -29,6 +30,7 @@ export async function createAuthKey(
   const jwt = await new SignJWT({
     '_couchdb.roles': user.roles ?? [],
     name: user.name,
+    server: CONDUCTOR_PUBLIC_URL,
```
PeterBaker0, line 33: do we use issuer yet - that's a more standard way to describe the server that issued a token - see https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-token-claims and https://www.iana.org/assignments/jwt/jwt.xhtml#claims

api/src/authkeys/create.ts
```diff
@@ -42,3 +44,18 @@ export async function createAuthKey(
     .sign(signingKey.privateKey);
   return jwt;
 }
+
+export async function generateUserToken(user: Express.User) {
+  const signingKey = await KEY_SERVICE.getSigningKey();
+  if (signingKey === null || signingKey === undefined) {
+    throw new Error('No signing key is available, check configuration');
+  } else {
```
PeterBaker0, line 52: no need for else here.

api/src/authkeys/create.ts
```diff
@@ -42,3 +44,18 @@ export async function createAuthKey(
...
+    const token = await createAuthKey(user, signingKey);
+
+    return {
+      token: token,
+      pubkey: signingKey.publicKeyString,
+      pubalg: signingKey.alg,
+    };
```
PeterBaker0, line 59: agree with eventually removing the alg and pub key from this payload

api/src/core.ts
```diff
@@ -53,16 +53,16 @@ import markdownit from 'markdown-it';
...
+if (process.env.NODE_ENV !== 'test') {
+  // set up rate limiter: maximum of 30 requests per minute
+  const limiter = RateLimit({
+    windowMs: 1 * 60 * 1000, // 1 minute
+    max: 30,
+    validate: true,
+  });
+  app.use(limiter);
+  console.log('Rate limiter enabled');
+}
```
PeterBaker0, line 65: I did hit this while testing the feature

api/src/couchdb/invites.ts
```diff
@@ -21,30 +21,74 @@
...
-    await invite_db.put(invite);
+    let done = false;
+    while (!done) {
+      try {
+        await invite_db.put(invite);
+        done = true;
+      } catch {
+        invite._id = generateId();
+      }
+    }
```
PeterBaker0, line 90: invite_db.put could fail for any reason - this could easily infinite loop if something else goes wrong - would recommend catching the specific error and putting a retry limit on this.

PeterBaker0: Even better - separate this into phases 1) determine if code is unique by getting from DB 2) put into DB. Relying on property of put to throw error if it exists and _rev is not provided seems a bit risky.

api/src/routes.ts
```diff
@@ -251,17 +231,11 @@ app.get('/send-token/', (req, res) => {
 
 app.get('/get-token/', async (req, res) => {
   if (req.user) {
-    const signingKey = await KEY_SERVICE.getSigningKey();
-    if (signingKey === null || signingKey === undefined) {
+    try {
+      const token = await generateUserToken(req.user);
+      res.send(token);
+    } catch {
       res.status(500).send('Signing key not set up');
```
PeterBaker0, line 238: does this error still make sense - could be other reasons generateUserToken fails?

app/ios/App/App/Info.plist
```diff
@@ -30,7 +30,7 @@
 		</dict>
 	</array>
 	<key>CFBundleVersion</key>
-	<string>202407170719</string>
+	<string>202409110011</string>
```
PeterBaker0, line 33: Is this covered in that ticket about hardcoded values?

app/src/gui/components/authentication/auth_return.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const navigate = useNavigate();
+
+  useEffect(() => {
+    const storeToken = async (token: string) => {
+      const token_obj = decodeJwt(decodeURIComponent(token));
+
+      console.log('decoded', token_obj);
+
+      const listing_id = await getListingForConductorUrl(
+        token_obj.server as string
```
PeterBaker0, line 57: Okay - I see this being picked up here

app/src/native_hooks.ts
```diff
@@ -19,81 +19,36 @@
...
+    CapacitorApp.addListener('appUrlOpen', (event: URLOpenListenerEvent) => {
+      const url = new URL(event.url);
+      console.log('Event URL', url);
+      // grab the 'pathname' part of the URL, note that url.pathname
+      // is not correct on Safari so we go the long way around
+      const redirect = url.href.substring(url.protocol.length + 1);
+      Browser.close();
+      console.log('navigating from app url to', redirect);
+      navigate(redirect);
+    });
```
PeterBaker0, line 40: Should this useEffect return a cleanup function which detaches this listener?

PeterBaker0 (approved):
Functionality works well. Couple of minor front-end comments and a few quirks. 

Some changes to make to code for safety/clarity. 

Thanks - this is a big improvement to how the auth flow works and is noticeably simpler/smoother to use. 

I **have not** tested on phone - will require some work to get that going. I know you have so I think we merge this now and sort out phone testing in general soon.

PR #1142, opened 2024-09-09
feat: custom hook for offline/online status and fallback component

PeterBaker0:
# feat: custom hook for offline/online status and fallback component

## JIRA Ticket

[BSS-321](https://jira.csiro.com/browse/BSS-321)

## Description

Adds a custom hook for offline/online status and fallback component which can be rendered to display connectivity issues. 

Also adds example component which can be added to pages to test functionality. Will need to remove before merging.

## How to Test

Currently the home page will show the offline/online as a test component. You can then use the browser dev mode to simulate different network conditions and make sure the component responds as intended. 

image

This is the fallback component - note that refresh just checks the online status again with window navigator, and return home goes to index. The refresh button will always show a short load to reassure user. 

@luke-mcfarlane-rocketlab  @stevecassidy  if you could test this works similarly on the android app that would be great.

## Additional Information

Currently can't get the useStyles to pickup the theme properly. That's why I have marked this as draft. I'd like to refactor the fallback styling to use the theme palette etc once this is working.

## Checklist


PeterBaker0:
This is ready now for testing - will remove the test component before merging. Theming is not utilising the ThemeProvider properly but is okay for now.

luke-mcfarlane-rocketlab (changes requested):
Tested on web, iOS, Android.

- works find on web/Android.
- issues on iOS simulator though it seems to be caused by the fact that although there is no internet connection, the phone is still connected to a network. So we might need to have an extra check check for the case where the phone is connected to a network but with no internet.

Requesting changes for the removal of the test component.

luke-mcfarlane-rocketlab (approved):
LGTM

PR #1143, opened 2024-09-10
dev: new survey API mock interfaces

PeterBaker0:
# dev: new survey API mock interfaces

Proposed set of interfaces / endpoints for new functions in the new survey API.

See `library/data-model/src/types.ts` for the new interfaces. 

```typescript 
// ================== MOCKS FOR NEW SURVEY API ===============
// TODO cleanup

// Template database item

// These are the fields used to instantiate the item
type UiSpecificationType = object;

export interface TemplateDbDocumentDetails {
  name: string;
  ui_specification: UiSpecificationType;
}

// Once it's in the DB we'll have a _id field as well
export interface TemplateDbDocument extends TemplateDbDocumentDetails {
  _id: string;
}

// Conductor CRUD API

// Create new template

// To create a new template
// POST /templates

// Expects JSON Payload
interface PostCreateTemplateInput extends TemplateDbDocumentDetails {}
interface PostCreateTemplateResponse extends TemplateDbDocument {}

// To update an existing template
// POST /templates/:id

// Expects JSON Payload
interface PostUpdateTemplateInput extends TemplateDbDocumentDetails {}
interface PostUpdateTemplateResponse extends TemplateDbDocument {}

// To get all templates
// GET /templates'
interface GetListTemplatesResponse {
  templates: Array<TemplateDbDocument>;
}

// To get a specific template by _id
// GET /templates/:id'
interface GetTemplateByIdResponse extends TemplateDbDocument {}

// To delete a specific template by _id
// POST /templates/:id/delete'
// 200 OK ack

// To create a new survey from a template by ID
/**
 * POST to /notebooks/template/:id to create a new notebook from a template ID
 */

type NotebookMetadataType = object;

interface PostCreateNotebookFromTemplate {
  template_id: string;
  project_name: string;
  metadata: NotebookMetadataType;
}

```

### DIFF
--- a/library/data-model/src/types.ts
+++ b/library/data-model/src/types.ts
@@ -678,3 +678,66 @@ export interface InitialMergeDetails {
   initial_head: RevisionID;
   initial_head_data: RecordMergeInformation;
 }
+
+// ================== MOCKS FOR NEW SURVEY API ===============
+// TODO cleanup
+
+// Template database item
+
+// These are the fields used to instantiate the item
+type UiSpecificationType = object;
+
+export interface TemplateDbDocumentDetails {
+  name: string;
+  ui_specification: UiSpecificationType;
+}
+
+// Once it's in the DB we'll have a _id field as well
+export interface TemplateDbDocument extends TemplateDbDocumentDetails {
+  _id: string;
+}
+
+// Conductor CRUD API
+
+// Create new template
+
+// To create a new template
+// POST /templates
+
+// Expects JSON Payload
+interface PostCreateTemplateInput extends TemplateDbDocumentDetails {}
+interface PostCreateTemplateResponse extends TemplateDbDocument {}
+
+// To update an existing template
+// POST /templates/:id
+
+// Expects JSON Payload
+interface PostUpdateTemplateInput extends TemplateDbDocumentDetails {}
+interface PostUpdateTemplateResponse extends TemplateDbDocument {}
+
+// To get all templates
+// GET /templates'
+interface GetListTemplatesResponse {
+  templates: Array<TemplateDbDocument>;
+}
+
+// To get a specific template by _id
+// GET /templates/:id'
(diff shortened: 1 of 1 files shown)


ranisa-gupta16:
@PeterBaker0   I've reviewed the PR, and the **TemplateDbDocument** and related interfaces (**TemplateDbDocumentDetails**, **GetListTemplatesResponse**, etc.) make sense and seem well-structured. Thanks for adding comments, I believe this is more than enough for me to finish up with this workflow. :-)

Just a query, `For **GetListTemplatesResponse**, I am assuming there will be scenarios where the list might be empty or fail to load due to an API error? If so, Should we display a fallback UI or retry mechanism if the template list fetch fails?`. This can be setup in advance while the API development is in progress.

PeterBaker0:
> @PeterBaker0 I've reviewed the PR, and the **TemplateDbDocument** and related interfaces (**TemplateDbDocumentDetails**, **GetListTemplatesResponse**, etc.) make sense and seem well-structured. Thanks for adding comments, I believe this is more than enough for me to finish up with this workflow. :-)
> 
> Just a query, `For **GetListTemplatesResponse**, I am assuming there will be scenarios where the list might be empty or fail to load due to an API error? If so, Should we display a fallback UI or retry mechanism if the template list fetch fails?`. This can be setup in advance while the API development is in progress.

Hi @ranisa-gupta16 - I think the UI should use a standard retry logic. An empty list of templates is not really an 'error' - this is a valid state - but maybe it shows a special message e.g. "It seems like there are no templates available! Contact your system administrator and try again later." 

As for retrying - this is a broader question than this particular API - usually an API framework like react-query handles retrying up to some threshold or for particular error codes. I would default to providing a fallback when the API errors describing the error code. But worth more discussion.

PeterBaker0:
Going to merge this as it's out of date. See #1144

PR #1145, opened 2024-09-11
feat: project db and context overhaul

luke-mcfarlane-rocketlab:
## JIRA Ticket

Main Ticket: [BSS-317](https://jira.csiro.au/browse/BSS-317)

## Description

Added context provider for projects and new local database to store state for offline use.

### Projects Context Provider
- created projects context provider which initialises from the projects API and local database, triggers re-renders whenever projects are updated and keeps state synced with local storage for offline or poor internet use
- created functions for updating project state
- created functions for getting remote projects
- wrapped app with projects context provider
- updated existing components to use the new context provider

### Projects Local DB
- create new local db for storing projects in the same format as received from the API
- create helper functions for updating the new projects local db

### Other
- remove a bunch of console logs from around the project so we better monitor the console for errors and warnings
- add the test environment rate limiter back (I believe one of my previous merges removed it)
- removed some loading states to be refined at a later point (this state can be provided using the context provider)

## How to Test

1. View projects list
2. Try to activate a project
3. Try to toggle sync of a project
4. Open a project to view and/or update

## Additional Information

I would like to eventually implement loading skeletons with loading state passed by the context provider so that the components do not jump around on load.

## Checklist


stevecassidy (commented):
I think this looks really good overall.  Probably missing something that we'll catch later but I think we can go with this. 
I've added one change which is to include project_id (which is ${listing_id}||${_id}) in the Project type so that it's accessible to callers and we don't need to make up the id string everywhere.  Tried to catch all of the places where that was happening.

stevecassidy (approved):
Sorry, meant to approve.

PR #1148, opened 2024-09-16
feat: increasing usage of Zod for API routes, improving error handling, misc refactoring

PeterBaker0:
# feat: increasing usage of Zod for API routes, improving error handling, misc refactoring

## Description

Makes the following changes

- Moves the hard-coded role names for cluster admin and notebook group to data model - currently the env variable is ignored in the build config so this hardcoded value may as well be available to the front-end so it can reason about user auth - happy to chat more on this though
- adds --exit flag to mocha tests since it was hanging on some presumably unclosed async stuff, tried to fix this but couldn't find any other way around it
- changes approach to error handling in API routes to address a few factors
  - ensure all errors are caught and passed through to next(e) to avoid hang-ups which were previously present all the way through to the frontend 
  - this is now being handled by https://www.npmjs.com/package/express-async-errors which wraps all routes automatically in `try {...} catch (e) {next(e)}` for us. 
  - use exceptions which extend a base custom exception to handle errors more consistently and concisely e.g. `throw new Exceptions.ItemNotFoundException("Could not find notebook")` rather than repeatedly reimplementing the error logic `res.status(401).json({error: ..})` etc 
- add more zod types for notebook and other routes which specify input/output interfaces where feasible
- add more type annotations for existing typescript interfaces within zod models - this is a pathway to getting full validation for example ui-spec, but it won't actually validate yet
- changing error handling pattern from 

```
const value = await something();
if (value){
...
}
else {
throw new Exception 
}
```
to 
```
const value = await something();
if (!value) {
throw new Exception 
}
... // do stuff
```

this avoids nested if statements and also provides a more logical flow since the error logic is closer in proximity to the value.

- change the delete notebook route to no longer return a redirect - instead it returns a 200OK if successful or throws errors if something went wrong
- use more efficient type annotation approach - for example:

```typescript
/**
 * PUT update existing template
 * Updates an existing template. The payload is validated by Zod before reaching this
 * function. Expects a document as the response JSON. Requires cluster admin
 * privileges.
 */
api.put(
  '/:id',
  // these will be parsed at runtime and fully type the req.body and req.params - this can be for query args too with query: ... 
  processRequest({
    params: z.object({id: z.string()}),
    body: PutUpdateTemplateInputSchema,
  }),

  requireAuthenticationAPI,
 // annotating the response here forces typescript to alert you if returned res.json() does not conform to this interface.
  async (req, res: Response<PutUpdateTemplateResponse>) => {
    //...
  }
);
```
this avoids any repetition of annotations. 

- remove various any response types in utility functions
- add couch db type annotations
- add various doc strings and comments explaining logic
- remove redundant DB calls for project metadata etc by modularising logic to get metadata for project by ID or by projectDoc, improve error handling throughout
- replaces `indexOf > 0` with `includes` in various places
- avoids returning empty lists which obscure errors in various places - this is happening in the wrong place. Consider getUsers() - previously if the user DB could not be found it would return an empty list - this is obscuring a critical system fault and should instead throw an exception. If a route depending on this function wants to mask it with [], then it should catch the error and explicitly justify this decision.
- builds models for notebook users info as zod models and moves to data models instead of coding into api above function

## Note on testing workaround and --exit flag

The delete notebook route must have had a bug in it. It was returning a redirect even though the route wasn't actually deleting the notebook so the test had no idea it was failing. When I fixed it, the mocked databaseList object still retained a reference to the database for that notebook but it was actually destroyed. This then meant that the cleanup function would hang on db.destroy() presumably because the db was already destroyed. I think it's throwing an error but it's not getting handled so it just hangs, like we saw with api routes that didn't handle their exceptions properly. 
 
I have it all working again, the only weirdness I'm getting now is all the tests pass but the process hangs after saying all successful - I presume there is some async thing going on or db connection sitting around, adding --exit to the mocha command fixes it.
 
I added a workaround where the notebook delete test just manually cleans up the reference to that DB in the mocked database list so that the cleanup function doesn't try to do anything with the dead link - not the nicest but I presume this isn't an issue in a real deployment since there is no "database list" persay for the data DBs

## How to Test

Run unit tests to ensure functionality is not regressive. 

## Checklist


api/docker-compose.yml
```diff
@@ -49,9 +49,6 @@ services:
 
       # https://faimsproject.atlassian.net/wiki/spaces/FAIMS3/pages/151846960/FAIMS3+Conductor+build+run+environment+variables
 
-      # optional
-      - CLUSTER_ADMIN_GROUP_NAME
-
```
PeterBaker0, line 54: This is not used anywhere in the app - I've moved this to data models in a new file. This will be important for the front end to be able to reason about what parts of the application the user should be able to see as we move functionality into the main app.

api/src/api/notebooks.ts
```diff
@@ -58,19 +65,26 @@ import {
 import * as Exceptions from '../exceptions';
 import {requireAuthenticationAPI} from '../middleware';
 
+// See https://github.com/davidbanham/express-async-errors - this patches
+// express to handle async errors without hanging or needing an explicit try
+// catch block
+require('express-async-errors');
```
stevecassidy, line 71: Why is this 'require' rather than 'import'?

api/src/couchdb/index.ts
```diff
@@ -145,53 +152,116 @@ export const getInvitesDB = (): PouchDB.Database | undefined => {
   return _invitesDB;
 };
 
-export const getProjectMetaDB = async (
+/**
+ * Gets a database client object for the metadata DB for a given project document
+ * @param projectDoc The project document to use as a connection reference for the metadata DB
+ * @returns The connected instantiated database
+ */
+export const getMetadataDbFromProjectDocument = async (
```
stevecassidy, line 160: You've introduced this new export to use it only in the delete notebook method which could get the dbs from the project id instead.  I think it would be simpler just to access the dbs through the project id as before. Introducing another way just means you need longer funciton names to disambiguate them. I don't see any improvement by adding this.

api/src/couchdb/index.ts
```diff
@@ -145,53 +152,116 @@ export const getInvitesDB = (): PouchDB.Database | undefined => {
...
+  return new PouchDB(dbUrl, pouch_options);
+};
+
+/**
+ * Returns the metadata DB for a given project - involves fetching the project
+ * doc and then fetching the corresponding metadata db
+ * @param projectID The project Id to use
+ * @returns The metadata DB for this project
+ */
+export const getMetadataDbFromProjectId = async (
```
stevecassidy, line 187: Long Java-esque names not needed if you always get the db from the project id!

api/test_output
```diff
@@ -0,0 +1,4 @@
+
+> @faims3/api@0.1.0 test
+> env-cmd cross-env NODE_ENV=test TS_NODE_PROJECT='./tsconfig.json' mocha
```
stevecassidy, line 3: Should this be here?

stevecassidy (changes requested):
Only minor changes needed if you agree with my aversion to those new methods.  Pushed a change to fix the failing test when DEVELOPER_MODE = true - test never worked so whoever wrote it should be sacked!

api/src/api/notebooks.ts
PeterBaker0: Just using their docs example - but I'll see if the behaviour works with import.

api/src/couchdb/index.ts
PeterBaker0: The benefit was just for performance - reduces the round-trips to DB. There were flows in the code which were

- get project doc
- validate something about it 
- get metadata db (which gets project doc again)

With this change you can just do 

- get project doc
- validate something about it 
- get metadata db from project doc

Minor improvement though

api/src/api/notebooks.ts
PeterBaker0: Just tested and as discussed there is no typing on the patch module - require is a cleaner way to indicate/hide this.

api/src/couchdb/index.ts
PeterBaker0: Reverted to previous version, retaining additional documentation.

PR #1162, opened 2024-09-27
Fix: Field validation timing

stevecassidy:
# Fix Field Validation Timing

## JIRA Ticket

[BSS-352](https://jira.csiro.com/browse/BSS-352)

## Description

If a photo field is required we still show an error message after adding a photo in some cases (rapid impact prototype).   This seems to be due to when validation runs on the form - it doesn't run until a change is triggered and adding a photo doesn't trigger a change.

## Proposed Changes

It turns out that validation was not being triggered when we set field values because a third argument to `setFieldValue` was missing.  If this is true, then validation is triggered when the field is set.   All calls have been updated.  In a few cases, notably select, I had to do more work to ensure that it was called at the right time. 

Some fields are directly implemented via the Formik text field and so I've not changed them. TextField seems to work and I think the others do but in some cases you have to leave them to see the value (date field).

## How to Test

[Field-Sampler.json](https://github.com/user-attachments/files/17158913/Field-Sampler.json)
 is a sample notebook with lots of required fields.  You should see the error messages go away as you fill in values.

## Checklist

### DIFF
--- a/app/src/gui/components/record/fields.tsx
+++ b/app/src/gui/components/record/fields.tsx
@@ -61,7 +61,7 @@ export function getComponentFromFieldConfig(
       onWheel={(event: any) => event.target.blur()}
       onChange={(event: any) => {
         formProps.handleChange(event);
-        formProps.setFieldValue('updateField', fieldName);
+        formProps.setFieldValue('updateField', fieldName, true);
       }}
       disabled={disabled}
     />
@@ -79,7 +79,7 @@ export function getComponentFromFieldConfig(
       onWheel={(event: any) => event.target.blur()}
       onChange={(event: any) => {
         formProps.handleChange(event);
-        formProps.setFieldValue('updateField', fieldName);
+        formProps.setFieldValue('updateField', fieldName, true);
       }}
       issyncing={isSyncing}
       disabled={disabled}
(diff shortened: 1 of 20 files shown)


PeterBaker0 (approved):
Looks good - noting our discussion offline re: the weirdness of observed behaviour of explicitly setting validate to True vs not providing it, see https://formik.org/docs/api/formik#formik-render-methods-and-props - notably the setFieldValue comment "Set the value of a field imperatively. field should match the key of values you wish to update. Useful for creating custom input change handlers. Calling this will trigger validation to run if validateOnChange is set to true (which it is by default). You can also explicitly prevent/skip validation by passing a third argument as false."

Regardless, tried it out and photos seem to validate on time now.

stevecassidy:
Just had another look and it looks like `validateOnChange` is explicitly false (in `RecordForm`), the commit that did that says "Dropping `onChange` reduces the lag as we're not checking each keypress.".  Might need to look into this a bit more to make sure we're not killing performance - but also can more easily implement the change by setting `validateOnChange`.

stevecassidy:
Will merge as is. I think changing the global `validateOnChange` might give us a bit less control about when validation happens, this way we can force it when we want which might be an advantage.  We'll just have to be on the lookout for UI lag due to frequent validation.

PR #1163, opened 2024-09-27
Fix draft deletion

stevecassidy:
# Fix - draft deletion doesn't work

## JIRA Ticket

[BSS-399](https://jira.csiro.com/browse/BSS-399)

## Description

Deleting draft records was not working, the dialog stayed open after clicking delete and the record stayed in the list.

## Proposed Changes

The bug was with the refresh of the record lists. Deletion was happening but a  previous PR had changed the callback
function that was supposed to trigger refresh.   The solution was actually to move the refresh down to the record
table component.

The same problem existed with deleting records. This is also now fixed.

I've also removed a few bits of debugging output and fixed some warnings that showed up during testing. 

- TakePoint fields were passing the helperText property on to the Button component causing a  warning. This was because all properties from the configuration were just passed in blindly. I changed this to pass none in so we control the look of the button in code rather than via the notebook.
- Made the Map Input button full width to match the new default for TakePoint

## How to Test

Create some records (and drafts) and then try to delete them.

Use a notebook with a TakePoint and Map field to see that the buttons render appropriately.

## Additional Information

I investigated the extra draft creation problem and found that it was due to drafts being created in useEffect inside the DraftCreate component.  The double rendering of React.StrictMode means that this is called twice and makes two drafts each time.   This won't be a problem in production but points to a poor implementation.  I've created BSS-400 to address this at some point.

## Checklist

### DIFF
--- a/app/src/drafts.ts
+++ b/app/src/drafts.ts
@@ -40,7 +40,7 @@ export function listenDrafts(
       live: true,
     })
     .on('change', info => {
-      console.log(info);
+      console.log('change in drafts database', info);
       if (info.doc!.project_id === project_id) {
         runCallback();
       }
--- a/app/src/gui/components/notebook/delete.tsx
+++ b/app/src/gui/components/notebook/delete.tsx
@@ -73,7 +73,6 @@ async function deleteFromDB(
     );
     await deleteDraftsForRecord(project_id, record_id);
   }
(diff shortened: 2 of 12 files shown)


PeterBaker0 (approved):
Made some changes as discussed to remove added state variable, tested functionality and works as described.

PR #1164, opened 2024-09-30
Fix and update csv export

stevecassidy:
# Feat/Fix lat/long in CSV export and add new fields

## JIRA Ticket

[BSS-380](https://jira.csiro.com/browse/BSS-380)

## Description

CSV export extracts latitude and longitude for GPS locations as separate columns as well as including the JSON fragment.  These values were switched in the current implementation.  

## Proposed Changes

- Fix the order of lat/long in GPS points
- Add similar export code for map locations as long as they are points
- Add GPS accuracy as a separate column (user request)
- Add 'created' and 'created_by' fields in export
- Fix a bug that hangs CSV export when there are no records (or all records are deleted)

Note that the late of these involved making a change to an index used to access the database (the recordRevisions index) to 
include the created_by field (created was already there).   This means that we won't see created_by for any existing notebooks  unless we can update the indexes (not currently implemented).    It will show for any new notebooks

## How to Test

Create a new notebook (to get the new index) that includes GPS location and/or Map fields. Create some records in the app. Then go to the notebook page on Conductor and export the data as CSV.  The resulting CSV file should include created, created_by and the lat/long/accuracy fields for  your GPS points.

Try exporting from a notebook with no records, should get a CSV with just a header line.

## Checklist


api/src/couchdb/notebooks.ts
```diff
@@ -623,16 +624,35 @@ const csvFormatValue = (
   if (fieldType === 'faims-pos::Location') {
     if (value instanceof Object && 'geometry' in value) {
       result[fieldName] = value;
-      result[fieldName + '_latitude'] = value.geometry.coordinates[0];
-      result[fieldName + '_longitude'] = value.geometry.coordinates[1];
+      result[fieldName + '_latitude'] = value.geometry.coordinates[1];
+      result[fieldName + '_longitude'] = value.geometry.coordinates[0];
```
PeterBaker0, line 628: Not sure on level of 'trust' in the validity of the data at this point - what if geometry object does not have coordinates field or the array index does not exist? Should this fail gracefully, throw an exception? Go to next record?

library/data-model/src/data_storage/index.ts
```diff
@@ -615,6 +616,7 @@ export const notebookRecordIterator = async (
       }
       if (record) {
         const data = await hydrateRecord(project_id, record);
+        console.log('returning hydrated record', record.created_by);
```
PeterBaker0, line 619: want to keep this?

PeterBaker0 (approved):
See the changes and note the update to the data model. Lgtm.

library/data-model/src/data_storage/index.ts
stevecassidy: removed

api/src/couchdb/notebooks.ts
stevecassidy: Added some guards to this and the other geometry export.

PR #1165, opened 2024-10-01
feat: survey form layout

luke-mcfarlane-rocketlab:
## JIRA Ticket

[BSS-382](https://jira.csiro.au/browse/BSS-382)

## Description

Changed the layout of the create site record form to be in a single long view.

## Proposed Changes

- updated form component to show list with headings rather than tabs
- moved errors to the end of the form

## How to Test

This was tested using the Rapid Impact Survey.

1. Navigate to a survey
2. Add a new record
3. Verify layout and form functionality
4. (optional) Open an already created record and verify layout and functionality

## Additional Information

This is an update rather than configurable. Happy to discuss the option of being configurable but it seems currently there are a few bugs in the form code and if we were to maintain both we will need to address these in each version of the component.

## Checklist


stevecassidy (commented):
The layout looks ok but we need to have a configuration option to select between this and the old layout.

stevecassidy:
Thinking about this, the option to use inline vs tabs for sections should be a configuration on the form in the uiSpec.  Then I could if I wanted have one form with tabs and another with inline sections.  Need to add in a configuration option to the Designer.

luke-mcfarlane-rocketlab:
@stevecassidy issues above have been addressed

stevecassidy (changes requested):
Generally ok, see my small comments.  We need to document this change to the notebook format but I'll make a separate ticket for that and include the other recent change.

app/src/gui/components/record/form.tsx
```diff
@@ -1185,29 +1167,169 @@ class RecordForm extends React.Component<
...
+                                fontWeight: 'bold',
+                                padding: '10px',
+                              }}
+                            >
+                              {ui_specification.views[view].label}
+                            </div>
+                            <Form>
+                              {description !== '' && (
+                                <Box
+                                  bgcolor={'#fafafa'}
```
ranisa-gupta16, line 1207: Can you use theme specific color variables?

luke-mcfarlane-rocketlab:
@stevecassidy @ranisa-gupta16 comments actioned

PR #1171, opened 2024-10-09
feat: record progress bar

luke-mcfarlane-rocketlab:
## JIRA Ticket

[BSS-383](https://jira.csiro.au/browse/BSS-383)

## Description

Add progress bar to top of record edit form to track completion of required fields.

## Proposed Changes

- add progress bar component
- update current form screens to include new component
- pass state updater to form to update progress
- create helper functions for measuring completion

## How to Test

1. Navigate to any survey
2. Click "+ NEW RECORD" button
3. Input values into required fields an check that progress bar reflects these changes
4. (optional) Publish and close record
5. (optional) Open published record and verify completion is retained

## Additional Information

The survey form seems reasonably unreliable and might be worth considering a rebuild at some point.
- the form re-renders around 11 times on load and another 5 times every few seconds (even without any changes)
- in testing I was able to get into a broken state where I would need to kill the tab and reload to resolve
- the code uses class components and is thousands of lines long across multiple files

## Checklist


app/src/lib/survey.ts
PeterBaker0, line 1: Would propose this file is named more descriptively - e.g. formUtils ?

app/src/lib/survey.ts
```diff
@@ -0,0 +1,43 @@
...
+    .map(([key]) => key);
+
+/**
+ * Determines whether a given value is considered "empty".
+ * A value is considered empty if it is an empty array, an empty string, or explicitly set to null.
+ *
+ * @param {any} value - The value to be checked. It can be of any type.
+ * @returns {boolean} - Returns true if the value is considered empty, otherwise false.
+ */
+const emptyValue = (value: any): boolean => {
```
PeterBaker0, line 21: Boolean fields default to 'false' - this means they are reported as complete even when user has never interacted with them. This is out of scope of this PR though.

PeterBaker0 (approved):
Seems to work well - noting the issue with required boolean field progress as discussed.

PR #1174, opened 2024-10-14
docs: configuration architecture and approach

PeterBaker0:
Adds `docs/Configuration.md` which describes the forms of configuration in the FAIMS3 application, and a process to determine where configuration should occur.

### DIFF
--- a/docs/developer/Configuration.md
+++ b/docs/developer/Configuration.md
@@ -1,0 +1,114 @@
+# FAIMS3 Configuration
+
+This document describes the configuration architecture of the FAIMS3 application. The configuration consists of three main components:
+
+1. Notebook Specification: The most comprehensive form of configuration, defining what to render in a notebook, and how. It includes metadata, field definitions, forms, and views/sections. This is used for notebook-specific configurations and can be modified by users through the Designer application.
+
+2. App UI Theme: Global style customization for the mobile application, set through an environment variable. It affects all notebooks and is not user-changeable.
+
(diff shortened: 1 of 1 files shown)


docs/developer/Configuration.md
```diff
@@ -0,0 +1,114 @@
...
+- metadata: a set of key value pairs which contain
+  - important notebook metadata such as name, description, author etc
+  - configuration switches such as QR code behaviour (TODO document this further)
+  - arbitrary user metadata key-value pairs
+- field definitions: a set of fields with a unique name, this includes detailed specifications on how to render the field, it's allowable returned values etc
+- forms: top level forms which display as discrete forms in the application e.g. Site, Building - these are a collection of views, which are a collection of fields
+- views/sections: a sub-section within a form, which contain a collection of Fields identified from the field definitions list by their unique identifier
+
+**When should this be used**: The notebook specification is a powerful form of configuration for the behaviour of a _specific_ notebook. When the behaviour to be configured applies to a) a specific notebook b) individual fields/forms/sections c) some notebooks but not all d) should be configurable by the user within a given deployment, then the notebook specification should be used.
+
```
stevecassidy, line 52: Note that making a change to the notebook format should be done in a backward compatible way if possible - code should work if the option is missing or has an older style value.  Any change to the format will generally also require an update to the Designer to allow users to customise that option.

stevecassidy (changes requested):
Good apart from these couple of changes.

stevecassidy (approved):

PR #1180, opened 2024-10-15
fix(aws): jwt auth backwards compatibility (exp claim) and chttpd_auth security improvements

PeterBaker0:
# fix(aws): jwt auth backwards compatibility (exp claim) and chttpd_auth security improvements

## Description

- adds configurable couch version tag and pins to 3.3.3 by default
- the above provides a workaround to the `jwt_auth` breaking change where 'exp' is a mandatory claim - see JIRA ticket to implement and integrate token refresh. Issue here (https://github.com/apache/couchdb/issues/5046)
- updates the user data couch configuration to inject the cookie secret from the FAIMS_COOKIE_SECRET shared AWS Secret Manager, rather than hardcoding. The dist file was copied which included a placeholder secret value which was replaced at build time with a secret.

## How to Test

AWS deployment is up to date with these changes including re-creating a new instance. 

## Additional Information

It might be worth extending the initdb API - however we would need to ensure that the chttpd_auth secret configuration is not required to enable API connection - this could be chicken and egg problem. This approach is secure for the AWS use case but does mean that dynamically updating the secret requires manual intervention.

## Checklist

### DIFF
--- a/api/couchdb/local.ini.dist
+++ b/api/couchdb/local.ini.dist
@@ -10,7 +10,6 @@ port = 5984
 bind_address = 0.0.0.0
 authentication_handlers = {chttpd_auth, cookie_authentication_handler}, {chttpd_auth, jwt_authentication_handler}, {chttpd_auth, default_authentication_handler}
 
-
 [httpd]
 enable_cors = true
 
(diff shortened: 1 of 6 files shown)


stevecassidy (approved):
Can't see any issues.

PR #1182, opened 2024-10-16
fix: fix express async issue, return API unit tests, add turbo config

PeterBaker0:
# fix: fix express async issue, return API unit tests, add turbo config

- changes require('express-async-errors') to instead hardcode the actual patch into a file, and export a function - this resolves the issue where the require was being cached causing improper patching, regressing the async timeout on API unit tests
- fixes merge issue which removed the conductor short code prefix from the listings object in info 
- adds the API unit tests as a turbo config, the other tests run but don't currently pass due to existing issues. To run unit tests, ensure `.env` file is valid and then run `turbo run test --filter=@faims3/api`
- updates github action to run this on PR

### DIFF
--- a/.github/workflows/build-lint.yml
+++ b/.github/workflows/build-lint.yml
@@ -59,4 +59,12 @@ jobs:
         
       - name: Build
         run: npm run build
+
+      - name: API Unit tests
+        run: |
+          cd api
+          cp .env.dist .env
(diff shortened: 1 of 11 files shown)


stevecassidy (approved):
Looks good, nice to have the tests back in CI.

PR #1183, opened 2024-10-16
Add Database Validation Utility

stevecassidy:
# Database validation utility and more robust record retrieval

## JIRA Ticket

[BSS-479](https://jira.csiro.au/browse/BSS-479)

## Description

Fix a crashing bug where a badly formed database will cause data export to crash. 
Adds developer tools to do some validation of the CouchDB database so that we can try 
to diagnose why data is getting messed up.

## Proposed Changes

- Fix up record retrieval in the export code to be robust to missing documents - revisions and avps
- Some minor enhancements to export like using the form name instead of ID to display and name files
- Adds a basic validation report on the database so we can see what documents are missing

## How to Test

Data export in a broken database should now complete and export all records that are not broken.  Eg. randomly
delete some revisions or avps in your CouchDB database and it should still export the rest.

Filenames for export of CSV and Zip will be based on the form name not the form ID.  Export list
in the notebook page also uses the form name. 

When developer tools are enabled in conductor (`DEVELOPER_MODE=true`) a button to validate 
the database will be shown on the notebook page. It should give you a report on the missing 
documents.

## Checklist


PeterBaker0:
Confirming I deleted an avp and got this result

image

PeterBaker0 (approved):
lgtm - only question is why the missing AVP causes the whole record to be skipped. AVPs are one value (right?) so couldn't it put missing for specific column value then include the rest. I might be misunderstanding. Either way it's improvement from current state. 

Tested diagnostic tool and correctly noted missing AVP. Tested CSV export and was durable against this.

PR #1185, opened 2024-10-17
Update Android Deployment

stevecassidy:
# Update and Parameterise Android Deployment

## JIRA Ticket

- [BSS-405](https://jira.csiro.com/browse/BSS-405)
- [BSS-344](https://jira.csiro.com/browse/BSS-344)

## Description

To allow us to depoy the BSS app we need to be able to configure the build and
deploy process to make it more flexible.   

## Proposed Changes

- Refactors the Android build and deployment pipeline to add ENV variables that allow us to customse the application deployed to Google Play
- Documents the Android deployment process for future reference
- Removes some old deployment related code that is no longer relevant

## How to Test

We can test by running the Nightly Android Test Build workflow on Github. This should
work as before on the main FAIMS repo and with the right configuration, publish to
the new BSS app from the BSS-NBIC repo.

## Checklist


app/android/fastlane/Appfile
```diff
@@ -1,10 +1,3 @@
-json_key_file(ENV["JSON_KEY_FILE"]) # Path to the json secret file - Follow https://docs.fastlane.tools/actions/supply/#setup to get one
-package_name("au.edu.faims.fieldmark") # e.g. com.krausefx.app
+json_key_file(ENV["ANDROID_JSON_KEY_FILE"]) # Path to the json secret file - Follow https://docs.fastlane.tools/actions/supply/#setup to get one
+package_name(ENV["ANDROID_APP_ID"])
```
PeterBaker0, line 2: In one place you provided a fall back default value - is that necessary as I note it was not provided here?

PeterBaker0: Maybe we want it to fail if the app ID is not supplied?

PeterBaker0 (approved):
Looks good - appreciate the documentation and removal on unused steps in the workflow.

app/android/fastlane/Appfile
stevecassidy: It does fail if package_name is not set with a suitable warning.

stevecassidy: On the default, I guess it doesn't hurt. The build would succeed if it is not set but the upload would fail.

PR #1188, opened 2024-10-18
feat: implement token refresh and on-demand refresh in app upon created new survey

PeterBaker0:
# feat: implement token refresh and on-demand refresh in app upon created new survey

## JIRA Ticket

https://jira.csiro.au/browse/BSS-438
https://jira.csiro.au/browse/BSS-439

## Description

Implements a backend API stateful token refresh workflow. 

Adds logic in front end to refresh when a new notebook is created so that the updated token has the new roles, enabling the user to see the updated list. 

## Proposed Changes

This PR touched a **lot** of files as I was chasing through some unnecessarily complex/redundant handling of the tokens in the front end. Apologies for the huge diff. I made some improvements along the way, notably

- fixed initialisation bug which was causing listings to intermittently not appear on fresh load until refreshed
- added the parsed token contents into the local auth db so that you don't need to repeatedly parse the token contents
- new approach for the new Auth DB which makes use of native javascript code which is serialised into the DB upon init 
- removed a lot of `useState(token)` - this is unnecessary as the token is available from the local auth db

## TODOs before merge

- unit tests for API 
- developer documentation on refresh logic

## How to Test

Clear your local storage, logout, login. Your local storage will now have a refresh token. 

Then setup a user which has the create notebooks permission but not admin. Try creating a new survey. The survey list should be updated to reflect the new notebook, showing the token has the correct permissions/roles. 

## Checklist


api/src/couchdb/index.ts
```diff
@@ -314,3 +328,64 @@ export const closeDatabases = async () => {
...
+  }
+  return _authDB;
+};
+
+/**
+ * Initialises the Auth DB by injecting a security document, and design
+ * documents from the data model.
+ *
+ * NOTE: This is currently an upsert so that DBs can be maintained, but this
+ * might be exploitable since the endpoint is not authenticated.
```
PeterBaker0, line 356: @stevecassidy  this is not in alignment with the other init methods. Let me know what you think about it upserting - this will make it easier to 'migrate' a database programmatically, but really should need an authenticated route.

app/src/constants/privateRouter.tsx
```diff
@@ -2,16 +2,32 @@ import {Navigate} from 'react-router-dom';
...
   // tslint:disable-next-line:no-any
   children: React.ReactElement;
-  allowed: boolean;
 }
 
+/**
+ * Conditional auth route
+ * @param props children to render
+ * @returns Conditionally renders children if token is present (not necessarily
+ * valid) for the first listing entry in the listings db
```
PeterBaker0, line 17: @stevecassidy this is an example of where the app has no concept of which token to protect routes against - the idea we discussed of an 'active' profile will solve this problem as we have a definitive token to check for.

app/src/context/functions.tsx
```diff
@@ -61,7 +84,15 @@ export const getRemoteProjects = async () => {
     if (!_id || !conductor_url) continue;
 
     const token = await getToken(_id);
-    const response = await getProjects(conductor_url, token);
+
+    if (!token) {
+      // TODO consider implications of this
```
PeterBaker0, line 89: @luke-mcfarlane-rocketlab ... worth your view on this behaviour

app/src/context/store.tsx
```diff
@@ -71,6 +72,8 @@ const {Provider} = store;
 
 const StateProvider = (props: any) => {
   const [initialized, setInitialized] = useState(false);
+  const startedInitialisation = useRef(false);
```
PeterBaker0, line 75: @stevecassidy  this records whether we have started the initialisation instead of relying on global let variable - more idiomatic for react.

app/src/gui/components/authentication/auth_return.tsx
```diff
@@ -40,51 +39,100 @@ async function getListingForConductorUrl(conductor_url: string) {
...
+      try {
+        await setTokenForCluster(
+          decodedToken,
+          parsedToken,
+          decodedRefreshToken,
+          listing_id
+        );
+      } catch (e) {
+        return setErrorAndReturnHome(
+          'Auth return route attempted to store token in local auth DB but encountered an error. ' +
```
stevecassidy, line 92: Seeing this error whenever I try to login: 
Error: Auth return route attempted to store token in local auth DB but encountered an error. TypeError: current_doc.available_tokens.get is not a function. You will be returned to home in 10 seconds.

stevecassidy: Sorry, it was the old db structure as you said.

stevecassidy: However, can we code around this at all? Would be a big black mark if users can't login after an update.

api/src/couchdb/index.ts
stevecassidy: Firstly this should probably go in initialise.ts with the other init code.  Making it an upsert makes sense if we have a migration strategy but without one it just makes this db different to all the others which is probably not a good idea. I'd say write it like the others for now and make a ticket to develop a migration scheme for databases in future.

app/src/gui/components/authentication/auth_return.tsx
PeterBaker0: This has been resolved as discussed by reverting to plain JS object but providing a wrapper ObjectMap class which provides a map interface over the object.

PeterBaker0: In the future version tagging with migration strategy would be ideal.

luke-mcfarlane-rocketlab (approved):
Other than the build errors this looks good to me.

Desired behaviour is working for me and none of the context/state management around projects appears to be effected.

PR #1189, opened 2024-10-23
feat: progress bar in my sites

luke-mcfarlane-rocketlab:
## JIRA Ticket

[BSS-375](https://jira.csiro.au/browse/BSS-375)

## Description

Added progress bar to sites table in mobile view.

## Proposed Changes

- created queries to get 

## Checklist


app/src/gui/components/notebook/record_table.tsx
```diff
@@ -72,6 +77,11 @@ type RecordsBrowseTableProps = {
 function RecordsTable(props: RecordsTableProps) {
   const {project_id, maxRows, rows, loading} = props;
 
+  const {data: uiSpec} = useQuery({
+    queryKey: ['uiSpec', project_id],
+    queryFn: () => getUiSpecForProject(project_id),
+  });
```
PeterBaker0, line 83: I think this could be helpful in other areas so would prefer you make a simple custom hook e.g. useGetUiSpec which wraps the useQuery.

app/src/gui/components/notebook/record_table.tsx
```diff
@@ -243,6 +253,14 @@ function RecordsTable(props: RecordsTableProps) {
           minWidth: 150,
           filterable: true,
           renderCell: (params: GridCellParams) => {
+            const {project_id, record_id, revision_id} = params.row;
+
+            const {data: record} = useQuery({
+              queryKey: ['recordData', project_id, record_id, revision_id],
+              queryFn: () =>
+                getFullRecordData(project_id, record_id, revision_id),
+            });
```
PeterBaker0, line 262: Surely you can't put a useQuery in a render function - won't you get a bunch of React errors with hooks being used in different orders depending on the contents of the list?

PeterBaker0 (changes requested):
Doesn't work currently as useQuery in the render function causes React error e.g. 

image

luke-mcfarlane-rocketlab:
@PeterBaker0 issues above resolved

app/src/gui/components/notebook/record_table.tsx
```diff
@@ -448,29 +439,45 @@ function RecordsTable(props: RecordsTableProps) {
...
+            record.revision_id
           );
-          setPouchData(ra);
+
+          if (!data) continue;
+
+          record.record = data;
         }
+
+        setPouchData(records);
```
PeterBaker0, line 491: Is this is probably going to conflict with my pending PR #1199 which moves this into a react query - but that's okay we can fix it up

app/src/gui/components/notebook/record_table.tsx
```diff
@@ -448,29 +439,45 @@ function RecordsTable(props: RecordsTableProps) {
...
+                props.filter_deleted
+              )
+        ) as RecordMetaDataExtended[];
+
+        for (const record of records) {
+          const data = await getFullRecordData(
+            record.project_id,
+            record.record_id,
+            record.revision_id
           );
```
PeterBaker0, line 484: Performance concerns here possibly - but we'll see.

app/src/lib/queries.ts
PeterBaker0, line 1: cool - I think there is a space for hooks already though

PeterBaker0 (commented):
Mildly concerned about performance implications of fetching all data for all entries in the record list on every remount. Also not sure on layout efficiency - want to wait until Ranisa's PRs are resolved to see how it could fit in updated layout.

app/src/gui/components/notebook/record_table.tsx
stevecassidy: For the PR to display summary fields in the record table we'll be grabbing the whole record data so this would not be necessary. A few interacting changes going on here. Can this be deferred?

stevecassidy:
Not what you're doing here but your earlier progress bar breaks the rules by updating the parent component (Record) while rendering the child RecordForm.  The call to setProgress in RecordForm.render is the problem, it needs to move out of render and into one of the other methods.  See the console when you render a form for the warning.

PR #1194, opened 2024-10-24
Customise Deployment of Android App

stevecassidy:
# Customise Deployment of Android App

## JIRA Ticket

[BSS-405](https://jira.csiro.com/browse/BSS-405)
[BSS-344](https://jira.csiro.com/browse/BSS-344)

## Description

Makes a few more things customisable in the app build to facilitate deployment to the Google Play Store as a separate app. 

## Proposed Changes

- Adds environment variables for some previously fixed values relating to app store deployment
- Adds documentation of the Android deployment workflow for future reference
- Automatically generate app icons based on configured theme
- Automatically update app name and id based on configured values

Note. tested on Android deployment but also touches some parts of Apple deployment which will need testing and fixing later.

## How to Test

Hard to test independantly but nothing here should break the build. Note that there is a new task run before starting the app to generate icons so one immediate change you see will be the main app icon will change with theme changes (need to re-run `npm start` since it is a pre-start step). 

## Checklist


.github/workflows/deploy-aws-cdk.yml
PeterBaker0, line 1: this shouldn't be in this PR - was for nbic-bss only

.github/workflows/publish-conductor.yml
```diff
@@ -5,13 +5,14 @@ on:
...
     types: [published]
+  workflow_dispatch:
 
 # Defines two custom environment variables for the workflow. These are
 # used for the Container registry domain, and a name for the Docker
 # image that this workflow builds.
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAME: FAIMS/faims3-api
+  IMAGE_NAME: nbic-bss/faims3-api
```
PeterBaker0, line 15: this also shouldn't be here

PeterBaker0 (approved):
If you can remove those two files/changes as commented from the PR which are BSS only changes, that would be good. Not sure the best way to manage "bss" only stuff - typically we'd want to only PR from upstream -> fork.

.github/workflows/publish-conductor.yml
stevecassidy: We should make this customisable via variables.

stevecassidy: Customised via ${{github.repository_owner}}.

.github/workflows/deploy-aws-cdk.yml
stevecassidy: We can include the workflow and disable it for FAIMS.

PR #1201, opened 2024-11-01
feat: adding permission handling into create new survey workflow views

PeterBaker0:
# feat: adding permission handling into create new survey workflow views

## JIRA Ticket

[BSS-465](https://jira.csiro.au/browse/BSS-465)

## Description

Updates the create new survey workflows to respect authorisation. Keeping in mind this is not currently well defined because there is no sense of 'active listing'. This is the behaviour proposed and implemented.

- on the surveys page, if there is ANY listing in which the active user for that listing has creation permissions, show the button
- on the create new <survey> page, filter the listings to only show listings for which the user has permission

## How to Test

Create a user with permission, and one without, and check that things render as expected. Noting that you will need to manually refresh tokens using the Workspace refresh button to get the updated tokens. 

## Checklist

### DIFF
--- a/app/src/gui/components/workspace/CreateNewSurvey.tsx
+++ b/app/src/gui/components/workspace/CreateNewSurvey.tsx
@@ -1,17 +1,22 @@
 import DescriptionIcon from '@mui/icons-material/Description';
 import {
-  Grid,
-  useMediaQuery,
-  Typography,
-  useTheme,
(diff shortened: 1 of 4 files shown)


luke-mcfarlane-rocketlab (approved):
LGTM - tested with both admin/non-admin users

PR #1202, opened 2024-11-01
feat: short code UI/UX refinements

PeterBaker0:
Refining short code by removing prefix automatically, updating to correct prefix when detected, styling improvements, layout improvements, popup improvements, and fixing token bug for multiple listings.

https://jira.csiro.au/browse/BSS-457


luke-mcfarlane-rocketlab (approved):
Works great now, good to merge - though we might want to increase the width of the prefix input box at some point.

Also, matter of preference but we could just show the prefix in the left box and place the dash between the two.

PR #1204, opened 2024-11-01
chore: changing button label to 'Add new' instead of 'New'

PeterBaker0:
Changing button label to 'Add new' instead of 'New'

https://jira.csiro.au/browse/BSS-471


luke-mcfarlane-rocketlab (approved):
LGTM

PR #1209, opened 2024-11-04
feat: initialise surveys

luke-mcfarlane-rocketlab:
## JIRA Ticket

[BSS-451](https://jira.csiro.au/browse/BSS-451)

## Description

Initialise surveys on page load and button to trigger remote sync manually.

## Proposed Changes

- added init/sync projects functions to context provider
- added init project to run on login
- added sync projects button
- made tab default to 2 if no activated projects

## How to Test

1. Log out of app
2. Log into app
3. Check that surveys are visible without refresh
4. (optional) Repeat 1-3 but clearing cache after logout
5. Add survey in conductor
6. Hit refresh button and ensure it becomes visible

## Checklist

### DIFF
--- a/app/src/context/projects-context.tsx
+++ b/app/src/context/projects-context.tsx
@@ -3,11 +3,10 @@ import {
   activateProjectDB,
   getProjectsDB,
   setSyncProjectDB,
-  updateProjectsDB,
 } from '../dbs/projects-db';
 import {activate_project} from '../sync/process-initialization';
 import {ProjectExtended} from '../types/project';
-import {getRemoteProjects} from './functions';
+import {getProjectMap, getRemoteProjects} from './functions';
 
 export const ProjectsContext = createContext<{
   projects: ProjectExtended[];
@@ -46,37 +45,48 @@ export function ProjectsProvider({children}: {children: ReactNode}) {
    */
   const initProjects = async () => {
     const localProjects = await getProjectsDB();
-
-    const newProjectsMap = new Map(
-      localProjects.map(project => [project._id, project])
-    );
-
     const remoteProjects = await getRemoteProjects();
 
+    const localProjectsMap = getProjectMap(localProjects);
+    const newProjectsMap = getProjectMap(localProjects);
+
     for (const remoteProject of remoteProjects) {
-      newProjectsMap.set(remoteProject._id, remoteProject);
+      const activated =
+        localProjectsMap.get(remoteProject._id)?.activated ?? false;
+
+      newProjectsMap.set(remoteProject._id, {
+        ...remoteProject,
+        activated,
+      });
     }
 
     setProjects([...newProjectsMap.values()]);
   };
 
+  /**
+   * Synchronizes the list of projects with remote projects, updating the local project list
+   * to include any remote project updates while preserving the activation state.
+   *
+   * @async
+   * @function syncProjects
+   * @returns {Promise<void>} Resolves when the project synchronization is complete.
+   */
   const syncProjects = async () => {
-    const newProjectsMap = new Map(
-      projects.map(project => [project._id, project])
-    );
-
     const remoteProjects = await getRemoteProjects();
 
+    const projectsMap = getProjectMap(projects);
+    const newProjectsMap = getProjectMap(projects);
+
     for (const remoteProject of remoteProjects) {
-      newProjectsMap.set(remoteProject._id, remoteProject);
+      const activated = projectsMap.get(remoteProject._id)?.activated ?? false;
+
+      newProjectsMap.set(remoteProject._id, {
+        ...remoteProject,
+        activated,
+      });
     }
 
-    for (const project of projects) {
-      newProjectsMap.set(
-        project._id,
-        newProjectsMap.get(project._id) ?? project
-      );
-    }
+    setProjects([...newProjectsMap.values()]);
   };
 
   /**
--- a/app/src/context/projects-context.tsx
+++ b/app/src/context/projects-context.tsx
@@ -13,10 +13,14 @@ export const ProjectsContext = createContext<{
   projects: ProjectExtended[];
   activateProject: (id: string, listing: string) => void;
   setProjectSync: (id: string, listing: string, sync: boolean) => void;
+  syncRemoteProjects: () => Promise<void>;
+  initProjects: () => Promise<void>;
 }>({
   projects: [],
   activateProject: () => {},
   setProjectSync: () => {},
+  syncRemoteProjects: () => new Promise(() => {}),
+  initProjects: () => new Promise(() => {}),
 });
 
 /**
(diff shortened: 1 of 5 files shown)


app/src/context/projects-context.tsx
```diff
@@ -28,34 +32,67 @@ export function ProjectsProvider({children}: {children: ReactNode}) {
...
+  const syncRemoteProjects = async () => {
+    const remoteProjects = await getRemoteProjects();
 
-      const remoteProjects = await getRemoteProjects();
+    if (!remoteProjects.length) return;
 
-      if (!remoteProjects.length) return;
+    const currentProjectsMap = new Map(
+      projects.map(project => [project._id, project])
+    );
```
PeterBaker0, line 68: This might be a common op - maybe make into helper function/class method? E.g. getProjectMap(this): Map<string, Project>

app/src/context/projects-context.tsx
```diff
@@ -28,34 +32,67 @@ export function ProjectsProvider({children}: {children: ReactNode}) {
...
 
-      const localProjectsMap = new Map(
-        localProjects.map(project => [project._id, project])
-      );
+    const newProjects = remoteProjects.map(project => {
+      return {
+        ...project,
+        activated: !!currentProjectsMap.get(project._id)?.activated,
```
PeterBaker0, line 73: Can you add a comment - truthy hackery is a bit hard to read e.g. 'default to false, use existing value if available'.

app/src/context/projects-context.tsx
```diff
@@ -28,34 +32,67 @@ export function ProjectsProvider({children}: {children: ReactNode}) {
...
 
-      if (!remoteProjects.length) return;
+    const currentProjectsMap = new Map(
+      projects.map(project => [project._id, project])
+    );
 
-      const localProjectsMap = new Map(
-        localProjects.map(project => [project._id, project])
-      );
+    const newProjects = remoteProjects.map(project => {
```
PeterBaker0, line 70: so you are mapping only across the remote projects, this means that any projects which were removed online will be no longer available on the device? Is this desired behaviour. What if unsynced changes?

app/src/context/projects-context.tsx
```diff
@@ -28,34 +32,67 @@ export function ProjectsProvider({children}: {children: ReactNode}) {
...
+   * Initializes the application by synchronizing local and remote projects.
+   * First, it fetches projects from the local database and sets them in the project state,
+   * then it synchronizes with remote projects to update the local state and database as needed.
+   *
+   * @async
+   * @function init
+   * @returns {Promise<void>} Resolves when initialization is complete.
+   */
+  const initProjects = async () => {
+    await syncLocalProjects();
```
PeterBaker0, line 48: Aren't you going to have race conditions since you are using `useState` setters inside these async functions. 

E.g. you might get 

- syncLocalProjects
- setProjects(local result) queued
- syncRemoteProjects
- setProjects(remote result) queued 
- setProjects(remote result) executed
- setProjects(local result) executed

the result being the local would overwrite the remote, or vice versa.

Might need to instead have the async functions return results, then do a deterministic merge between them, and set it once.

app/src/gui/components/authentication/auth_return.tsx
```diff
@@ -96,6 +99,7 @@ export function AuthReturn() {
 
       // this requires the token
       update_directory();
+      initProjects();
```
PeterBaker0, line 102: How does React handle the async ops prior to a navigate - does it assure they are executed? I guess navigate isn't a refresh so the execution should continue in the background?

PeterBaker0: Not really a huge fan of this being added in here - it feels a bit of a hammer solution but that's okay I guess it works.

app/src/gui/components/workspace/notebooks.tsx
```diff
@@ -194,17 +198,35 @@ export default function NoteBooks() {
...
+            <Button
+              variant="contained"
+              onClick={() => history(ROUTES.CREATE_NEW_SURVEY)}
+              sx={{mb: 3, mt: 3, backgroundColor: theme.palette.primary.main}}
+              startIcon={<AddCircleSharpIcon />}
+            >
+              Create New {NOTEBOOK_NAME}
+            </Button>
+          ) : (
+            <div />
```
PeterBaker0, line 214: why are you rendering an empty div?

app/src/gui/components/workspace/notebooks.tsx
```diff
@@ -194,17 +198,35 @@ export default function NoteBooks() {
...
-            onClick={() => history(ROUTES.CREATE_NEW_SURVEY)}
+            disabled={refresh}
             sx={{mb: 3, mt: 3, backgroundColor: theme.palette.primary.main}}
-            startIcon={<AddCircleSharpIcon />}
+            startIcon={<RefreshOutlined />}
+            onClick={async () => {
+              setRefresh(true);
+              await syncRemoteProjects();
+              setRefresh(false);
+            }}
```
PeterBaker0, line 225: sounds like a useMutate opportunity?

app/src/gui/components/workspace/notebooks.tsx
```diff
@@ -194,17 +198,35 @@ export default function NoteBooks() {
...
           >
-            Create New {NOTEBOOK_NAME}
+            Sync {NOTEBOOK_NAME}s
```
PeterBaker0, line 227: Sync is not the right term here - overloaded with data sync - refresh would be better.

app/src/gui/components/workspace/notebooks.tsx
```diff
@@ -194,17 +198,35 @@ export default function NoteBooks() {
...
+          )}
           <Button
             variant="contained"
-            color="primary"
-            onClick={() => history(ROUTES.CREATE_NEW_SURVEY)}
+            disabled={refresh}
             sx={{mb: 3, mt: 3, backgroundColor: theme.palette.primary.main}}
-            startIcon={<AddCircleSharpIcon />}
+            startIcon={<RefreshOutlined />}
+            onClick={async () => {
```
PeterBaker0, line 221: this op is very fast and gives no visible indication to user that it is occurring, can you do a debounced/min duration loading icon or similar? Or even use the global popup provider to flash a little success alert like "{NOTEBOOK_NAME}s refreshed"

PeterBaker0 (changes requested):
Fix seems to work, however have some concerns on implementation details which could prove to be unpredictable/unreliable.

app/src/gui/components/workspace/notebooks.tsx
luke-mcfarlane-rocketlab: empty div pushes the other button to the right in the flex box due to `justifyContent: 'space-between'`. Could change to `justifySelf: 'end'` on the child but I generally default to putting in the parent.

app/src/gui/components/workspace/notebooks.tsx
luke-mcfarlane-rocketlab: I like the idea of adding a success alert as it doesn't arbitrarily delay the user refreshing.  Will update

PeterBaker0 (approved):
lgtm - I think @stevecassidy  should review too

stevecassidy (approved):
Looks good

PR #1212, opened 2024-11-04
Remove slotprops to fix CI build

stevecassidy:
# Still trying to fix CI build of Designer

## JIRA Ticket

None

## Description

Still getting a build error in CI. This time remove slotProps which it claims does not
belong on Accordian contrary to the documentation and examples. 

Change means that each field will still be mounted on being closed, should not be 
an issue. 

## How to Test

Merge into main and hope that the CI build works

## Checklist

### DIFF
--- a/designer/src/components/field-editor.tsx
+++ b/designer/src/components/field-editor.tsx
@@ -123,9 +123,6 @@ export const FieldEditor = ({
           display: 'none',
         },
       }}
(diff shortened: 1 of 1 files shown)


PeterBaker0 (approved):
If necessary - sure - concerned about chasing CI build issues we can't replicate locally. Would rather address the build environment.

PR #1214, opened 2024-11-04
Fix display and toggle of user roles

stevecassidy:
# Fix display and toggle of user roles

## JIRA Ticket

[BSS-467](https://jira.csiro.com/browse/BSS-467)

## Description

In Conductor, changing the roles of a user on the /users page didn't seem to work properly. Here we
fix this.

## Proposed Changes

The fix is to the display of roles, existing code did not properly display whether a user had these roles
or not.  Toggling the roles worked but since the page wasn't displaying them, it was confusing and
unuseable. 

## How to Test

Visit the /users page on Conductor where you have more than one user, you should be able to add roles
and refresh the page to see that they persist.

Note, there is a quirk in that 'cluster-admin'  implies notebook creation so if you toggle admin for a user
and refresh the page, create notebooks will be checked.  I decided not to write code to be any fancier 
than this but added a note into the page.

## Checklist

### DIFF
--- a/api/src/routes.ts
+++ b/api/src/routes.ts
@@ -303,6 +303,7 @@ app.get('/notebooks/:id/users', requireClusterAdmin, async (req, res) => {
     const notebook = await getNotebookMetadata(project_id);
 
     const userList = await getUserInfoForNotebook(project_id);
+
     res.render('users', {
       roles: userList.roles,
       users: userList.users,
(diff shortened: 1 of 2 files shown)


PeterBaker0 (approved):
Seems to work fine - noting the added message regarding implies create notebooks.

PR #1215, opened 2024-11-05
feat: improved local development workflow

PeterBaker0:
# feat: improved local development workflow

## Description

Optimises the local docker deployment for rapid testing by 

- unifying base layers between app and api to share npm install + utilise build cache 
- optimising install steps by only copying package files before doing npm i 
- sharing base Dockerfile but exposing convenient targets for each service
- simplified docker-compose file which utilises build-time env variables from .env.dist at top level, then does not duplicate stuff from .env within each service
- docker-compose file refers to targets in the multi-stage docker build
- exposing the app as well as api, both live reloading from local mounts of relevant files

Updated scripts etc to utilise these updates. Updated README. 

Recommended 'fresh start' procedure is: 

```
docker compose down && ./scripts/purge.sh . && ./localdev.sh
```

## A note on compose env variables

The only thing that tripped me up is the not very well documented handling of env files with docker compose. Turns out there's two categories of env files/variables. 1) container env variables 2) compose env variables. The container env variables are defined by the env_file: command, however these are not available at compose build time so you can't do things like port: {conductor_internal_port} in the compose file. However, by default compose looks in .env (in the location of the compose file) for compose level env variables - this was how this was previously working. Because the compose file was sitting in the api folder, it was using ./api/.env to fill in the compose env variables. So there were two options left to fix this 1) expect user to run docker compose --env-file ./api/.env 2) make a .env file in root which just has necessary build time variables. I liked the second option more hence the new .env.dist and the cp .env.dist .env in the localdev script.

## How to Test

Clone repo, run above command. Test live reloading. Check app functions (noting you may need to refresh local storage given key refresh in the scripts.


luke-mcfarlane-rocketlab (approved):
Works really well - nice one

PR #1216, opened 2024-11-05
Fix horizontal scrolling issue due to bigger username, enhanced username button display

ranisa-gupta16:
## JIRA Ticket

(https://jira.csiro.au/browse/BSS-460)
https://jira.csiro.au/browse/BSS-483

## Description

Fixed the horizontal scrolling issue due to bigger username

## Proposed Changes

The display of the username button has been updated. On mobile, the username is truncated with ellipses at the end to indicate a longer name, wrapping the content within the button and preventing any horizontal scroll caused by longer usernames. In desktop mode, the full username is displayed without truncation.

## How to Test

Check the app bar header with different lengths of usernames.

Screenshots: 
Updated User profile , tooltip for showing username 
image
image
image
image

Mobile:
image-2024-11-06-01-43-45-626
image-2024-11-06-01-44-11-542
---------------------------------------------------------------------------------------------------------------------------------------
Desktop:
image-2024-11-06-01-40-07-937

## Checklist

- [y ] I have confirmed all commits have been signed.
- [y ] I have added JSDoc style comments to any new functions or classes.
- [y ] Relevant documentation such as READMEs, guides, and class comments are updated.


PeterBaker0:
I agree this is an improvement in that the layout doesn't overflow. 

Just not sure about the design overall - like I don't think many apps actually display the username in the button in the top right. 

image

I also note the alignment on the arrows above and below cloud have gone off. 

I would propose we need to rethink the design a bit to not be layout sensitive to the size of the username - like what about just a profile icon? Then if you click on it it opens a little popup with your email, first name, last name, quick operations e.g. logout, refresh token, and a link to the home page thing with the listings

PeterBaker0 (changes requested):
As discussed I can see that this stops overflow issue, but not a huge fan of the design overall. Happy to either just fix the alignment on the cloud and leave this as is for now, then do a nicer version later, or can have a go at doing a revised design in this PR.

PeterBaker0 (approved):
Fixed sign in styling to be more similar to before, and tightened up display a bit.

PR #1218, opened 2024-11-06
Fix Federated Login Redirect

stevecassidy:
# Fix Federated Login Redirect

## JIRA Ticket

None

## Description

Federated login wasn't returning the token to the app

## How to Test

Login to the app with Google.

## Checklist

### DIFF
--- a/api/src/auth_routes.ts
+++ b/api/src/auth_routes.ts
@@ -380,9 +380,7 @@ export function add_auth_routes(app: Router, handlers: string[]) {
         passport.authenticate(
           handler + '-validate',
           authenticate_return(req, res, next, redirect)
-          //          HANDLER_OPTIONS[handler]
         )(req, res, next);
-        console.log('TODO: may need to insert:', HANDLER_OPTIONS[handler]);
       } else {
         throw Error(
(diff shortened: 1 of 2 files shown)


api/src/auth_routes.ts
```diff
@@ -395,12 +393,10 @@ export function add_auth_routes(app: Router, handlers: string[]) {
...
         const redirect = req.session['redirect'] || '/send-token';
-        passport.authenticate(handler + '-validate', {
-          successRedirect: redirect,
-          failureRedirect: '/auth',
-          failureFlash: true,
-          successFlash: 'Welcome!',
-        })(req, res, next);
+        passport.authenticate(
+          handler + '-validate',
+          authenticate_return(req, res, next, redirect)
```
PeterBaker0, line 398: I note that this authenticate_Return function does not distinguish between successful and unsuccessful redirects where as the previous version does. I also note the error messages in the authenticate_return function don't make sense in the context of a social sign in e.g. "invalid email or password" - this could be confusing to the user.

PeterBaker0 (approved):
Would be good to a) explain what this change does b) make a ticket to document and cleanup the implementation later.

PR #1219, opened 2024-11-07
Fix form progress bar count

stevecassidy:
# Fix form progress bar count

## JIRA Ticket

None

## Description

The record completion progress bar was not working. It was counting
the percentiage of all fields in the uiSpec rather than those in 
the current form. 

Also fixed a warning about keys in the image list.

## How to Test

Progress bar should now show 100% completion when all required fields
are filled in.

## Additional Information

Note, I've left the update of the progress bar inside the render method since
there is no other place to put it that would work but it is updating the parent
state inside the render method which is bad form - get a warning if you refresh
the page when on the form but not in normal flow. 

## Checklist

### DIFF
--- a/app/src/gui/components/record/form.tsx
+++ b/app/src/gui/components/record/form.tsx
@@ -1164,13 +1164,6 @@ class RecordForm extends React.Component<
               }}
             >
               {formProps => {
-                this.props.setProgress?.(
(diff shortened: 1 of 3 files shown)


PeterBaker0 (approved):
Changes seem reasonable to me

PR #1220, opened 2024-11-08
fix: only show prefix selector if more than one listing

PeterBaker0:
# fix: only show prefix selector if more than one listing

We were having issues on iOS with a single listing of the prefix selector obscuring input on small/weird scaling displays. This is a **temporary fix** to facilitate user testing which conditionally renders the dropdown only when there is actually multiple listings to choose from. This allows the input field to scale to fill the whole available width.

Reasonable scaling example

image

Exaggerated large scaling example

image


luke-mcfarlane-rocketlab (approved):
Seems fine for temporary fix - small nitpick I would put a space before the "-" character.

PR #1221, opened 2024-11-08
feat: make photo fields more prominent

PeterBaker0:
# feat: make photo fields more prominent

Moves the 'label' text from the button itself to a <\h3\> component. Moves the helper text into a \<p\> element which helps with size and spacing. The button is adorned with a camera icon, and uses a contained style which makes it more visible. 

@stevecassidy worth considering if this change will negatively impact any existing notebooks which rely on the button text being the label - I didn't really personally think this was very helpful as the button always does the same thing, in any notebook, it 'takes a photo'. 

image

## Checklist

### DIFF
--- a/app/src/gui/fields/TakePhoto.tsx
+++ b/app/src/gui/fields/TakePhoto.tsx
@@ -22,13 +22,14 @@ import React from 'react';
 import {FieldProps} from 'formik';
 import Button, {ButtonProps} from '@mui/material/Button';
 import {Camera, CameraResultType, Photo} from '@capacitor/camera';
+import CameraAltIcon from '@mui/icons-material/CameraAlt';
 
 // import ImageList from '@mui/material/ImageList';
 import ImageListItem from '@mui/material/ImageListItem';
 import ImageListItemBar from '@mui/material/ImageListItemBar';
 import IconButton from '@mui/material/IconButton';
 import ImageIcon from '@mui/icons-material/Image';
-import {Typography} from '@mui/material';
+import {Box, InputAdornment, Typography, useMediaQuery} from '@mui/material';
 import DeleteIcon from '@mui/icons-material/Delete';
 import {createTheme, styled} from '@mui/material/styles';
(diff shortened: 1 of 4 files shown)


app/src/gui/fields/TakePhoto.tsx
```diff
@@ -287,49 +300,57 @@ export class TakePhoto extends React.Component<
...
+            <Button
+              variant="contained"
+              color="primary"
+              fullWidth={isMobile ? true : false}
+              onClick={this.takePhoto}
+            >
+              Take photo
+              <span style={{width: 10}} />
+              <CameraAltIcon />
+            </Button>
```
PeterBaker0, line 324: This is the only part that actually changed - FYI

app/src/gui/fields/TakePhoto.tsx
```diff
@@ -18,23 +18,36 @@
...
+ */
+function MediaQueryWrapper({
+  children,
+}: {
+  children: (isMobile: boolean) => React.ReactNode;
+}) {
+  const isMobile = useMediaQuery(theme.breakpoints.down('sm'));
+  return children(isMobile);
+}
+
```
stevecassidy, line 50: Is it worth defining this component here just for a single use, seems a lot of overhead for a boolean flag, line 47 is all you need.

app/src/gui/fields/TakePhoto.tsx
stevecassidy: Happy with the layout change.

stevecassidy (approved):
Ok but question the need for introducing a new component.

luke-mcfarlane-rocketlab (approved):
LGTM

app/src/gui/fields/TakePhoto.tsx
PeterBaker0: As discussed - cannot use custom hooks in class fields. Rewrite later.

PR #1228, opened 2024-11-11
feat: improved permission handling for MapFormField

PeterBaker0:
# feat: improved permission handling for MapFormField

This PR enables a center to be defined while still requesting and gracefully handling a current location call (which prompts permission check). This is to allow user testing with the following workflow. 

This might be temporary since in reality we probably don't need to ask for the user's location if we don't intend to actually use it. 

Technical changes include:
- useEffect with useRef to only run once
- use const variables instead of modifying readonly props
- only store get location result if center is not defined in props
- add alerts for warning conditions such as not being able to access location
- add alerts for using default map center location (no current location available AND no configured center location) - this defaults to Sydney

## How to Test

Try different permutations of permissions yes, no, ask, as well as providing and not providing a center in the notebook definition. Check it always respects the center provided, shows errors where appropriate, and notifications work. 

## Checklist

### DIFF
--- a/app/src/gui/fields/maps/MapFormField.tsx
+++ b/app/src/gui/fields/maps/MapFormField.tsx
@@ -29,9 +29,10 @@ import {FieldProps} from 'formik';
 import {Alert} from '@mui/material';
 import {Capacitor} from '@capacitor/core';
 import {APP_NAME} from '../../../buildconfig';
+import {useNotification} from '../../../context/popup';
 
 // If no center is available - pass this through
-const FALLBACK_CENTER = [0, 0];
+const FALLBACK_CENTER = [151.2093, -33.8688];
 
 export interface MapFieldProps extends FieldProps {
   label?: string;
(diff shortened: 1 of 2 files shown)


app/src/gui/fields/maps/MapFormField.tsx
```diff
@@ -114,13 +133,12 @@ export function MapFormField({
     <div>
       <div>
         <MapWrapper
-          label={
-            props.label ? props.label : 'Get ' + props.featureType + ' from Map'
-          }
-          featureType={props.featureType}
+          label={label}
```
stevecassidy, line 137: This will always be the default label, not the supplied one.

app/src/gui/fields/maps/MapFormField.tsx
```diff
@@ -44,52 +49,66 @@ export function MapFormField({
...
+          .catch(() => {
+            notify.showWarning(
+              'We were unable to access your current location. Map fields may not work as expected.'
+            );
+            setNoPermission(true);
+          });
+      }
+    };
+    getCoords();
+  }, []);
```
stevecassidy, line 112: Ok with this since we're providing an alert as well.

stevecassidy (approved):
A reasonable solution but I think you've lost the supplied label.

app/src/gui/fields/maps/MapFormField.tsx
PeterBaker0: Fixed - good catch

PR #1231, opened 2024-11-11
feat: simplified new user onboarding for single listing special case

PeterBaker0:
# feat: simplified new user onboarding for single listing special case

## Description

During user testing it became apparent most users are struggling with the onboarding due to a cluttered / unclear workflow to get started. This PR introduces a special case component which triggers under these conditions

- not signed in to ANY listing
- there is only one listing
- the token is not being loaded from local storage (to try and minimise flickering)
- QR scanning logic is reused, button props are passed through to enable styling of the QR buttons - only enabled on mobile platforms

Note: the implementation is not very high quality - I copy and pasted other components to try and not risk disturbing any existing stuff. I would like to clean this up later, probably when we handle the 'active persona' work later. 

### Not signed in, one listing (no qr)

image

### Not signed in, clicked access code

image

### Not signed in, one listing with QR

image

### Not signed in, multiple listings (old view)

image

### Signed in (multiple or single listings) (old view)

image

## How to Test

@stevecassidy  if you could, would be great if you could check that the QR scanning works on Android. 

Otherwise just log in, log out, check it works as expected.

## Checklist

### DIFF
--- a/app/src/gui/components/authentication/oneServerLanding.tsx
+++ b/app/src/gui/components/authentication/oneServerLanding.tsx
@@ -44,7 +44,7 @@ const OnboardingComponent = ({
         elevation={2}
         sx={{
           width: '100%',
-          maxWidth: 380,
+          maxWidth: 420,
           padding: '32px 24px',
           borderRadius: '28px',
           display: 'flex',
--- a/app/src/gui/components/authentication/oneServerLanding.tsx
+++ b/app/src/gui/components/authentication/oneServerLanding.tsx
@@ -1,0 +1,149 @@
+/*
+This is a special login landing page for the most common use case i.e. one
+listing, user not logged in.
+*/
+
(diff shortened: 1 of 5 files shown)


app/src/gui/components/authentication/oneServerLanding.tsx
```diff
@@ -0,0 +1,149 @@
...
+        </Typography>
+
+        {/* Sign In Button */}
+        <Button
+          variant="outlined"
+          fullWidth
+          startIcon={
+            <LoginOutlinedIcon sx={{color: theme.palette.primary.main}} />
+          }
+          onClick={async () => {
```
stevecassidy, line 76: Can you refactor this into a function that can be used in both places rather than copying here.

stevecassidy: scratch that, will do it in a later PR

stevecassidy (approved):

PR #1232, opened 2024-11-12
Improve Camera Permissions on Android

stevecassidy:
# Improve Camera Permissions on Android

## JIRA Ticket

None

## Description

Android app was not asking for permission to use the camera. 

## Proposed Changes

Add an explicit check for camera permission and more useful alerts if it is not available that include instructions on how to grant permission.

## How to Test

Requires Android build, try turning off camera permission.

## Additional Information

## Checklist

### DIFF
--- a/app/src/gui/fields/TakePhoto.tsx
+++ b/app/src/gui/fields/TakePhoto.tsx
@@ -27,26 +27,15 @@ import React from 'react';
 // import ImageList from '@mui/material/ImageList';
 import DeleteIcon from '@mui/icons-material/Delete';
 import ImageIcon from '@mui/icons-material/Image';
-import {List, ListItem, Typography, useMediaQuery} from '@mui/material';
+import {Alert, List, ListItem, Typography, useMediaQuery} from '@mui/material';
 import IconButton from '@mui/material/IconButton';
 import ImageListItem from '@mui/material/ImageListItem';
 import ImageListItemBar from '@mui/material/ImageListItemBar';
 import {createTheme, styled} from '@mui/material/styles';
 import {logError} from '../../logging';
 import FaimsAttachmentManagerDialog from '../components/ui/Faims_Attachment_Manager_Dialog';
-
-/**
- * This functional component provides a isMobile flag which can be used in class
- * components to dynamically style the component.
- */
-function MediaQueryWrapper({
-  children,
-}: {
-  children: (isMobile: boolean) => React.ReactNode;
-}) {
-  const isMobile = useMediaQuery(theme.breakpoints.down('sm'));
-  return children(isMobile);
-}
+import {Capacitor} from '@capacitor/core';
+import {APP_NAME} from '../../buildconfig';
 
 function base64image_to_blob(image: Photo): Blob {
   if (image.base64String === undefined) {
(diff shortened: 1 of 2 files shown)


app/src/gui/fields/TakePhoto.tsx
```diff
@@ -220,38 +200,53 @@ const FAIMSImageList = (props: ImageListProps) => {
...
+  );
+
+  const isMobile = useMediaQuery(theme.breakpoints.down('sm'));
+
+  React.useEffect(() => {
+    const value = props.form.values[props.field.name];
+    if (value !== null && value !== undefined) {
+      setImages(value);
+    }
+  }, [props.form.values[props.field.name]]);
```
PeterBaker0, line 227: Bit of an anti-pattern - if the only purpose of the images state is to remain synchronised with the props.form.values value, just declare a const of that value and use it.

app/src/gui/fields/TakePhoto.tsx
```diff
@@ -220,38 +200,53 @@ const FAIMSImageList = (props: ImageListProps) => {
...
+    } else {
+      const permissions = await Camera.requestPermissions({
+        permissions: ['camera'],
+      });
+
+      if (permissions.camera === 'denied') {
+        setNoPermission(true);
+        return;
+      }
+    }
```
PeterBaker0, line 248: So you can't use the Camera.request permissions on web? Interesting

app/src/gui/fields/TakePhoto.tsx
```diff
@@ -262,100 +257,90 @@ export class TakePhoto extends React.Component<
...
-}
+  return (
+    <div>
+      {title && <h3>{title}</h3>}
+      {helperText && <p>{helperText}</p>}
+      <Button
+        variant="contained"
+        color="primary"
+        fullWidth={isMobile ? true : false}
+        onClick={takePhoto}
```
PeterBaker0, line 287: This change looks similar to the one that broke it before which you fixed - are you sure just takePhoto is okay, you don't need async () => {await takePhoto()}; ?

app/src/gui/fields/TakePhoto.tsx
```diff
@@ -262,100 +257,90 @@ export class TakePhoto extends React.Component<
...
+      )}
+      <FAIMSImageList
+        images={images}
+        setopen={(path: string) => {
+          setOpen(true);
+          setPhotoPath(path);
+        }}
+        setimage={(newfiles: Array<any>) => {
+          props.form.setFieldValue(props.field.name, newfiles, true);
+        }}
```
PeterBaker0, line 326: Yeah I don't think we need the images useState.

app/src/gui/fields/qrcode/QRCodeFormField.tsx
```diff
@@ -153,12 +153,17 @@ export function QRCodeButton(props: QRCodeButtonProps): JSX.Element {
   };
 
   const startScan = async () => {
-    const permissions = await BarcodeScanner.checkPermissions();
+    // Ask for permission to use the camera, the explicit call here
+    // allows us to provide a more useful message to the user if they
+    // say no or have said no before
+    const permissions = await Camera.requestPermissions({
+      permissions: ['camera'],
+    });
```
PeterBaker0, line 161: so this will block until user interacts with popup?

PeterBaker0 (commented):
I don't think the useState with the images is necessary - and would be good to merge main into this branch to ensure the changes with recent login component are merged well.

app/src/gui/fields/TakePhoto.tsx
stevecassidy: That was because it was an object based component and it forgot what `this` was because of the call, here we're just calling a fn which doesn't have this so no bug.

app/src/gui/fields/TakePhoto.tsx
stevecassidy: Dropped the variable all together, just use the prop value directly.

PR #1233, opened 2024-11-12
Fix Android Logout

stevecassidy:
# Fix Android Logout

## JIRA Ticket

[BSS-484](https://jira.csiro.com/browse/BSS-484)

## Description

Logout didn't work on Android, well it almost worked but the UI still thought we were logged in.

## Proposed Changes

Changed the use of useQuery in relation to user tokens, it was being too aggressive in caching and handed back an old token after logout (only on Android, suspect a timing issue). 

## How to Test

Try logging in and logging out again on web and Android.

## Additional Information

Bug was because getAnyToken returned undefined when there was no token which useQuery treats as an error.   Replaced this with an object value and we're sweet as.

## Checklist


app/src/utils/tokenHooks.tsx
```diff
@@ -63,11 +63,22 @@ export const useGetAnyToken = () => {
...
+      }
     },
-    // The query key, which includes the listing ID to ensure proper caching
+    // we don't want this to be cached since then logout would break
     queryKey: ['get any token'],
+    refetchOnWindowFocus: true,
     refetchOnMount: true,
+    gcTime: 0, // don't cache it
+    staleTime: 0, // no really, please don't
+    enabled: true, // always refetch
```
PeterBaker0, line 81: I think just refetchOnMount/WindowFocus should work? But all good if this is working

Enabled isn't required for sure since it's default to true anyway.

PeterBaker0 (approved):
Changes make sense - lesson learned re: returning undefined. I did see console warnings, should have actioned. Cheers.

PR #1241, opened 2024-11-27
Improve Auto-incrementers

stevecassidy:
# Fix: Improve Auto-incrementers

## JIRA Ticket

None

## Description

Auto increment fields are used to generate human readable IDs for records.  There 
was a bug that caused them to not work if the initial value of the field was 
set to "" as it was in the designer. 

These fields require that a range of numbers is set for the device before records can
be properly created.  The past behaviour was to notice when they weren't set up but
not really help the user set them.

## Proposed Changes

- Deals with the case when the initial value is ""
- Turn the range setting form into a popup dialog
- trigger the popup if a user tries to create a record without the ranges being set

## How to Test

Open a notebook with an auto increment field configured, eg. [Campus Survey](https://github.com/FAIMS/FAIMS3-Notebook-Campus-Survey-Demo).  Try to create
a record after activating the notebook, you should see a popup to allow you to set
the ranges.

Same popup is accessible via the Settings tab on the notebook.

## Checklist

### DIFF
--- a/app/src/gui/components/autoincrement/edit-form.test.tsx
+++ b/app/src/gui/components/autoincrement/edit-form.test.tsx
@@ -19,7 +19,7 @@
  */
 
 import {fireEvent, render, screen, waitFor} from '@testing-library/react';
-import BasicAutoIncrementer from './edit-form';
+import {AutoIncrementEditForm} from './edit-form';
 import {expect, describe, it} from 'vitest';
 
 const props = {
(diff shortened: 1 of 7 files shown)


luke-mcfarlane-rocketlab:
I get the following error:
Screenshot 2024-11-28 at 3 57 43 PM

As a side note, could we create an auto-incrementer that doesn't require any user input?

stevecassidy:
Thanks for picking that up. I'll look at that.   The whole idea of incrementers is that the user gets to set the range, if not, then you get clashes between different users and the number is meaningless.

stevecassidy:
Following up on that error message. It's due to some old notebooks having initialValue=null for the field which is invalid according to the validation schema.  I can't get around that in code, we need to update those notebooks. I've added a migration in the notebook editor to automatically fix this for any notebook that is edited.  

Also rewrote the edit form to be bit nicer and use useQuery :-)

PeterBaker0 (approved):
code lgtm - see use of react query with cache invalidation - makes sense in this case I think

PR #1250, opened 2024-12-06
feat: active user state management and token refreshing

PeterBaker0:
# feat: active user state management and token refreshing

## Description

Implements improved state management for active users aligned with our proposed design of having an 'active user' which is defined as a combination of a server ID and a username in that server. The token, parsed token, refresh token, and expiry (to be determined once we embed into the token) are stored for all combinations of user/server which we know about/are logged in.

This is persisted in a [zustand](https://github.com/pmndrs/zustand) store, so that UI elements can subscribe to only relevant parts of the store. 

I have also updated the notebook list page to only view the active user's available notebooks. I have updated the cluster card, and login profile button to reflect these changes, adding an active user selector when the user is logged into more than one. 

Some issues/things to be discussed

- the biggest issue I had with implementation was that at a lot of points the app tries to perform operations which apply across **all** users, typically as part of initialisation/update operations on the local DBs - in these cases the pattern is usually something like 

```
get all listings
for each listing
get token(listing id)
do stuff e.g. fetch metadata
end
```

There's a few challenges here a) the app was not previously handling multiple users very well in this context, it was using the active user for the given listing/server - however who knows if this user is the best one to be making this query. Now, it just takes the first (and typically only) username available for that listing to do those operations. 

I think it would make more sense to only do update/sync operations over things 'relevant' to the current user/server combination - especially since once we turn on token expiry, all the other tokens are going to expire anyway - making any operations pointless/bound to fail. 

- what happens if refreshes fail e.g. if the user is offline - in this PR currently it will set the isAuthenticated to false, which will mean the private route will re-render, and the user will be prompted to log back in, which is not good for offline use cases - I think we need to discuss ideal behaviour here for the app to be tolerant to inconsistent online conditions. E.g. we could display some banner in the app?

Probably more I can't think of.

## How to Test

Setup multiple conductor URLs (I use a local one and the BSS deployment). Try logging in, logging out, of one at a time, both, etc, ensure the active user is updated properly. Try changing the active user - you can even do this while looking at a view such as the notebook list and see it refresh in real time. Make sure the app works otherwise - noting I'm not convinced it will work offline yet.

## Checklist


PeterBaker0:
> My first reaction is that you're introducing a new dependency in zustand. Not something I've come across. We already use Redux in the designer so adding another state management solution is a concern. We could discuss whether this is the right solution and it or Redux may be useful more broadly but I'd rather not just add it in to a PR like this.

Yeah worth discussing more - just wanted to give it a go as I was confident I could get it working quickly as it's simple to use. I have never used Redux before and this looked easier to get into.

I did investigate the bundle size - Zustand is super minimal, like 1kb minified, half that zipped - so not like bundle size is an issue - so the risk of introduction of the dependency would be in maintainability/exposure to bugs/incompatibilities. 

I guess firstly a) do we like the idea of using a state management instead of the local auth DB? I'm thinking this is a nicer approach because the additional complexity of managing the pouchDB instance locally has no benefit unless we are syncing, which we aren't here - so I feel it's just overkill. b) if so, do we want to use Zustand or something else - I imagine as long as redux has similar hooks we could replace it if it's really that big of a problem. 

I know @luke-mcfarlane-rocketlab has used Zustand a lot before - it will probably be a strong candidate for being used in the new conductor - in which case we could share this store between those apps which would save a fair bit of work.

stevecassidy:
At first glance, your implementation here doesn't seem to be much simpler than the use of Redux in Designer. See designer/src/state for how that looks.  Redux provides a very clean model of state management. Would have to look further into zustand  but I'm cautious about a less mainstream dependency.

PeterBaker0:
This has been replaced with #1260

PR #1251, opened 2024-12-09
Minor fixes before making an official V1.0 release

stevecassidy:
# Minor fixes before making an official V1.0 release

## JIRA Ticket

None

## Description

Just fixing up any minor things before finalising the 1.0.0 release.

## Proposed Changes

- Auto-incrementer field editor was too strict in checking values and didn't allow deleting of current value
- Auto-incrementer sometimes didn't get the existing ranges, I think because of the initial value in useQuery

## How to Test

Check that all of the above are working well on mobile.

## Checklist

### DIFF
--- a/app/src/gui/components/autoincrement/edit-form.tsx
+++ b/app/src/gui/components/autoincrement/edit-form.tsx
@@ -80,7 +80,6 @@ export const AutoIncrementEditForm = ({
       );
       return ranges;
     },
-    initialData: [],
     enabled: true,
   });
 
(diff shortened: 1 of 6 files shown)


app/src/gui/components/autoincrement/edit-form.tsx
```diff
@@ -207,20 +208,25 @@ type IncremenenterRangeProps = {
  * @param props component props
  */
 const IncrementerRange = (props: IncremenenterRangeProps) => {
-  const [start, setStart] = useState(props.range.start);
-  const [stop, setStop] = useState(props.range.stop);
+  const [start, setStart] = useState<number | string>(props.range.start);
+  const [stop, setStop] = useState<number | string>(props.range.stop);
```
PeterBaker0, line 212: do they really need to be a number or a string? Might be better to on each render parse the value as a number - just choose one?

PeterBaker0 (approved):
Looks minor - I haven't tested but not too concerned about auto-incrementers at this stage. Pretty opposed to the number | string setup but it's minor.

app/src/gui/components/autoincrement/edit-form.tsx
stevecassidy: This is to allow for the empty string, if I don't, you can't delete the number that's there to enter a new one.  Note this is only for the field value, not for the final data structure which is a number.

PR #1253, opened 2024-12-11
feat: map location 🗺️ 

luke-mcfarlane-rocketlab:
## JIRA Ticket

[BSS-469](https://jira.csiro.au/browse/BSS-469)

## Description

Update the map form fields to show the users live current location as a stylised blue dot (I have also added this to the survey map view)

## Proposed Changes

- Created functions to add the following to the map
    1. Center control component
    2. Current location dot
- Added function to get current location
- Updated survey map view and map form fields to use these

## How to Test

1. Navigate to any survey
2. Check that the centre control and current location dot are functioning correctly
3. Create a new record
4. Click "Map Location" and verify the centre control and current location dot are functioning correctly

## Checklist


PeterBaker0 (changes requested):
Two glitches I can see with a quick test

1. when you open a take point map form field, the first time you select a location, it renders as a blue dot very similar to the current location indicator, but then when you pan it reverts to the green dot
2. when I opened the overview map, it had two 'current location' blue dot indicators, one in my correct location, and one where it previously thought I was
3. this is not a 'live' location feature - it only checks location at initial map render right? That's okay for now but we should clarify this and make a separate ticket to make it update live.

PR #1254, opened 2024-12-11
fix: offline projects list :bug:

luke-mcfarlane-rocketlab:
## JIRA Ticket

[BSS-548](https://jira.csiro.au/browse/BSS-548)

## Description

Projects list not working while offline.

## Proposed Changes

- Added function to update local database with projects list on both sync and init

## How to Test

1. Ensure the projects list works without internet

## Additional Information

I believe the context provider was already pulling in the local projects it was just empty as the update local function was lost in an update.

## Checklist


stevecassidy (approved):
I still got no projects with your change, added a catch in getProjects to guard against being offline and it works ok for me now when offline. Thanks.

luke-mcfarlane-rocketlab:
> I still got no projects with your change, added a catch in getProjects to guard against being offline and it works ok for me now when offline. Thanks.

Thanks for applying the fix!

PR #1257, opened 2024-12-18
feat: new conductor :rocket:

luke-mcfarlane-rocketlab:
### Initial Setup

[Initial Environment Setup](https://jira.csiro.au/browse/BSS-598)
[Navigation Bar](https://jira.csiro.au/browse/BSS-576)

### Auth

[Auth API Routes](https://jira.csiro.au/browse/BSS-603)
[Login Form](https://jira.csiro.au/browse/BSS-601)
[Signup Form](https://jira.csiro.au/browse/BSS-602)
[Auth Context Provider](https://jira.csiro.au/browse/BSS-599)
[Protected Routes Setup](https://jira.csiro.au/browse/BSS-600)

### Surveys

[Survey List Page](https://jira.csiro.au/browse/BSS-577)
[Survey Details Page](https://jira.csiro.au/browse/BSS-578)

### Templates

[Templates List Page](https://jira.csiro.au/browse/BSS-585)
[Templates Details Page](https://jira.csiro.au/browse/BSS-587)

## Description

Initial setup of the new conductor web platform.

## Proposed Changes

- Initial environment setup
- Navigation bar
- Auth setup including API route, login & register forms, context provider, protected routes
- Surveys pages including list page, various details pages
- Templates pages including list page, various details pages
- Reusable table and form components

## How to Test

1. Install libraries
```
pnpm i
```
2. Run web server
```
pnpm run dev
```

## Additional Information

Progress on other pages but missing functionality
- Survey tabs other than details
- Templates tabs other than details
- User list page

## Checklist


web/src/auth.tsx
PeterBaker0, line 1: This is nice. 

Can you add the token refresh compatibility? Token refresh has now been merged to main so I think it's high priority to support that. Login gives you access token + refresh token. Refresh token can be exchanged for fresh access token. Access token expires every 5 mins (configurable). I'd recommend setting up a scheduled refresh a minute before the token expires. If you fail to refresh - unauthenticate the user to promote a fresh login.

web/src/components/data-table/column-header.tsx
PeterBaker0, line 1: Can you document in the README the file structure - i.e. what goes where and why? Notably distinguishing shadcn components from custom components, routes etc.

web/src/components/data-table/data-table.tsx
PeterBaker0, line 1: nice

web/src/components/forms/create-survey-from-template-form.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {useRouter} from '@tanstack/react-router';
+import {z} from 'zod';
+
+export const fields = [
+  {
+    name: 'name',
+    label: 'Survey Name',
+    schema: z
+      .string()
+      .min(3, {message: 'Survey name must be at least 3 characters.'}),
```
PeterBaker0, line 13: I think this is 5 actually.

web/src/components/logo.tsx
PeterBaker0, line 1: This will need to be configurable as part of theme

web/src/components/side-bar/team-switcher.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * TeamSwitcher component renders a dropdown menu within the sidebar that allows users
+ * to switch between different teams. It displays the active team's logo, name, and plan,
+ * and provides options to select other teams or add a new team.
+ *
+ * @param {TeamSwitcherProps} props - The properties object.
+ * @param {Team[]} props.teams - The array of teams to be displayed in the switcher.
+ * @returns {JSX.Element} The rendered TeamSwitcher component.
+ */
```
PeterBaker0, line 37: I'm guessing this is somewhat boilerplate from shadcn - could you update the nomenclature/docs to reflect our use case - i.e. we don't have a team here right?

web/src/components/tables/example.tsx
```diff
@@ -0,0 +1,35 @@
...
+import {DataTableColumnHeader} from '../data-table/column-header';
+
+export type Payment = {
+  id: string;
+  amount: number;
+  status: 'pending' | 'processing' | 'success' | 'failed';
+  email: string;
+};
+
+export const columns: ColumnDef<Payment>[] = [
```
PeterBaker0, line 11: Example component, remove?

web/src/components/tables/survey-users.tsx
```diff
@@ -0,0 +1,51 @@
...
+import {DataTableColumnHeader} from '../data-table/column-header';
+import {Trash} from 'lucide-react';
+import {Button} from '../ui/button';
+
+export type Payment = {
+  id: string;
+  amount: number;
+  status: 'pending' | 'processing' | 'success' | 'failed';
+  email: string;
+};
```
PeterBaker0, line 11: Payment?

web/src/components/tables/users.tsx
```diff
@@ -0,0 +1,73 @@
...
+import {DataTableColumnHeader} from '../data-table/column-header';
+import {Pencil, Trash} from 'lucide-react';
+import {Button} from '../ui/button';
+
+export type Payment = {
+  id: string;
+  amount: number;
+  status: 'pending' | 'processing' | 'success' | 'failed';
+  email: string;
+};
```
PeterBaker0, line 11: Payment?

web/src/components/theme-provider.tsx
PeterBaker0, line 1: how will this relate with BSS vs Fieldmark themes - what's the plan here - I see system vs dark but that's probably a different theme than a build time env variable switch like we have with the main app

web/src/components/ui/alert.tsx
PeterBaker0, line 1: this is shadcn isn't it? I'll ignore all these /ui components. E.g. what the hell is cva? haha

web/src/hooks/use-mobile.tsx
PeterBaker0, line 1: cool

web/src/index.css
PeterBaker0, line 1: Could you add a class comment here explaining when/why someone should edit these?

web/src/lib/queries.ts
PeterBaker0, line 1: Can you ensure that the response model is typed in this client - i.e. type generics for the get method or whatever which types the response model. I don't want to have to type cast the useQuery results anywhere I use it.

web/src/routes/_auth/surveys/$surveyId.tsx
PeterBaker0, line 1: Could you add some text to the README explaining how/where a developer should add new routes - and basic principles of how to do this with Tanstack router - okay if you just link to their docs for main steps.

web/src/routes/_auth/templates/$templateId.tsx
PeterBaker0, line 1: One pattern I've found very user friendly is having the tab as a query string argument which is automatically tracked so that deep links go straight to the tab - is this something tanstack router can do? Otherwise I just added a very nice custom hook in the main FAIMS3 code base we could reuse which provides a nice typed query string interface.

web/src/routes/_auth/users.tsx
PeterBaker0, line 1: Overall the routes file structure is a little confusing to me - so its routes/_auth/<all protected routes>

At first glance, you would expect that _auth/ would contain only auth related routes like login/register etc - rather than all the protected routes in the app - is there a way to separate these things so you could have like /_auth and also /templates /notebooks? 

Not urgent but I find the structure a bit unintuitive at first glance.

PeterBaker0 (changes requested):
Overall it's looking great. 

I think the code is high quality and well structured/documented. 

Some changes to make it ready to merge.

I have left comments on specific aspects, but also noted down some broader concerns. 

- (nice to have) add a clickable cursor on the choose file menu in the new survey form rather than the normal cursor
- (nice to have) highlight the active/current survey/template in the left hand panel and expand if active (e.g. if you go to template -> surveys -> click on the survey, the surveys side bar does not  expand to show that you are there)
- add turbo integration (e.g. build, lint, fix) ensuring proper dependencies for data model etc
- ensure npm is used instead of pnpm for consistency with turbo build and rest of project

**A broader note on auth:**

I'm concerned about the proliferation of auth routes - we probably need to come up with a unified auth strategy which is used by all platforms rather than having special web routes/approaches. This might be okay for now - but I'd like to discuss with the team about having register not require an invite code as is implied by your changes.

web/src/auth.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+export interface User {
+  user: {
+    id: string;
+    name: string;
+    email: string;
+  };
+  token: string;
+}
+
```
stevecassidy, line 11: We need to support the full range of login options that the backend supports.  It can be configured to support Google login etc.  Best option might be to redirect to the API login page for actual login entry as the app does, then you don't need to duplicate here.

web/src/auth.tsx
```diff
@@ -0,0 +1,150 @@
...
+  };
+
+  /**
+   * Logs in a user using the provided username and password.
+   * @param {string} username - The username of the user.
+   * @param {string} password - The password of the user.
+   * @returns {Promise<{status: string; message: string}>} The result of the login operation.
+   */
+  const login = async (username: string, password: string) => {
+    const response = await fetch('http://localhost:8080/web/login/local', {
```
stevecassidy, line 71: This needs to use the configured API URL but see my other comment about not doing it this way.

web/src/components/tabs/survey/survey-actions.tsx
```diff
@@ -0,0 +1,63 @@
+import {Alert, AlertTitle, AlertDescription} from '@/components/ui/alert';
```
stevecassidy, line 1: You've used 'survey' to name these modules! Try the neutral 'project' as per the app sourcecode.

web/src/components/tabs/survey/survey-users.tsx
```diff
@@ -0,0 +1,42 @@
...
+const SurveyUsers = ({surveyId}: {surveyId: string}) => {
+  const {user} = useAuth();
+
+  const {data, isPending} = useQuery({
+    queryKey: ['survey-users', surveyId],
+    queryFn: async () => {
+      if (!user) return [];
+
+      const data = await fetch(
+        `http://localhost:8080/api/notebooks/${surveyId}/users`,
```
stevecassidy, line 22: Hard coded API endpoint

web/src/components/tables/survey-users.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const columns: ColumnDef<Payment>[] = [
+  {
+    accessorKey: 'name',
+    header: ({column}) => (
+      <DataTableColumnHeader column={column} title="Name" />
+    ),
+  },
+  {
+    accessorKey: 'email',
```
stevecassidy, line 21: Email is not showing in my deployment.

stevecassidy (commented):
Generally good and just what we need but some language problems with excessive use of 'survey' that needs to be 'project' and needs a customisation option as per the app configuration.  In general borrow as much configuration from the app as possible.  

Login needs to redirect to the API auth endpoint to get other configured login options.  

Haven't looked deeply yet but looks very neat and just what we need.

api/src/api/auth.ts
```diff
@@ -0,0 +1,89 @@
...
+      return res.status(200).json({
+        user: {
+          id: user._id,
+          name: user.name,
+          email: user.emails[0],
+        },
+        token,
+      });
+    }
+  );
```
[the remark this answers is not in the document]
stevecassidy: These aren't needed since we should be doing login from the one page served by the API and getting the token on redirect back to the front end.

web/src/routes/_auth/users.tsx
luke-mcfarlane-rocketlab: _route folders add the layout and apply the functions in the _route.tsx file but do not add to the URL (rather than being used for organising pages). E.g. in this instance, for any page in "_auth/" the _auth.tsx has a before load that checks if the user is authenticated and adds the sidebar.

You are able to add what is called "route groups" to organise pages using (group) syntax but I find it confusing when using file based routing if there are folders without function.

The intuition for the name was that anything within _auth/ would require authentication but happy to change this if you can think of a more intuitive name.

web/src/routes/_auth/templates/$templateId.tsx
luke-mcfarlane-rocketlab: Definitely agree with the pattern and is something that tanstack router supports. The current behaviour was the default for shadcn tabs and so I left it as I did not think this would be a priority.

web/index.html
```diff
@@ -0,0 +1,14 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/vite.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Fieldmark Manager</title>
```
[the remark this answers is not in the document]
stevecassidy: System name can come from /api/info on the server.   Retain Conductor unless we can come up with a better name (I'm not wedded to it).

api/src/api/auth.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    req.logout((err: any) => {
+      if (err) logoutError = err;
+    });
+
+    if (logoutError) return res.status(400).send(logoutError);
+
+    return res.status(200).send('Logged out');
+  });
+}
```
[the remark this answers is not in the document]
luke-mcfarlane-rocketlab: Good shout - I forgot about this one

api/src/api/auth.ts
```diff
@@ -0,0 +1,89 @@
...
+      return res.status(200).json({
+        user: {
+          id: user._id,
+          name: user.name,
+        },
+      });
+    } catch (e) {
+      return res.status(400).send(e);
+    }
+  });
```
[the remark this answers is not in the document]
luke-mcfarlane-rocketlab: Agree, we can lump this in with the broader auth discussion.

api/src/api/auth.ts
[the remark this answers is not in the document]
luke-mcfarlane-rocketlab: Happy to use either approach but might be worth a discussion as this approach removed the need for the user to navigate between platforms to log in.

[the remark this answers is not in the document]
stevecassidy: No need to navigate between platforms, API just provides the login page that we redirect to from here, then redirects back. Same way as the app.   Seamless for the user.

web/src/components/ui/alert.tsx
luke-mcfarlane-rocketlab: Yeh thats right, you can ignore all the shadcn components in `ui/`. `cva` just provides a way to make typesafe tailwind classes with variants. If you look at the alertVariants function it is just taking the variant and returning the corresponding classNames as well as some classnames for all variants. Happy to go over any of the shadcn components, for the most part they are just divs/radix components with tailwind classes

web/src/index.css
luke-mcfarlane-rocketlab: Will also include in the README as this is how one would update the colour's of the application.

web/src/auth.tsx
luke-mcfarlane-rocketlab: This might take some more thought and time around design and implementation - will create a ticket and push in separate PR

web/src/components/logo.tsx
luke-mcfarlane-rocketlab: Will add a separate ticket/PR for this one

web/src/components/theme-provider.tsx
luke-mcfarlane-rocketlab: This is mainly a theme setup script - the actual colours can be modified in the `src/index.css` which will differ between the two deployments.

web/src/routes/_auth/templates/$templateId.tsx
luke-mcfarlane-rocketlab: I can add this as a low prio ticket.

web/src/auth.tsx
luke-mcfarlane-rocketlab: Removed the auth endpoints in this PR

web/src/auth.tsx
luke-mcfarlane-rocketlab: Removed the auth endpoints in this PR

web/src/components/forms/create-project-from-template.tsx
```diff
@@ -1,37 +1,38 @@
...
 
   /**
-   * Handles form submission for creating a survey.
+   * Handles form submission for creating a project.
    * @param {{name: string}} params - The submitted form values.
    * @returns {Promise<{type: string; message: string}>} The result of the form submission.
    */
   const onSubmit = async ({name}: {name: string}) => {
-    const response = await fetch(`http://localhost:8080/api/notebooks`, {
+    const response = await fetch('http://localhost:8080/api/notebooks', {
```
PeterBaker0, line 35: localhost!

web/src/components/forms/create-project-from-template.tsx
```diff
@@ -1,37 +1,38 @@
...
 export const fields = [
   {
     name: 'name',
-    label: 'Survey Name',
-    schema: z
-      .string()
-      .min(3, {message: 'Survey name must be at least 3 characters.'}),
+    label: `${NOTEBOOK_NAME_CAPITALIZED} Name`,
+    schema: z.string().min(5, {
+      message: `${NOTEBOOK_NAME_CAPITALIZED} name must be at least 3 characters.`,
```
PeterBaker0, line 13: 5 characters*

web/src/context/auth-provider.tsx
```diff
@@ -0,0 +1,155 @@
...
+  /**
+   * Logs out the user by removing the stored user object and setting the user to null.
+   */
+  const logout = () => {
+    setStoredUser(null);
+    setUser(null);
+  };
+
+  /**
+   * Logs in the user with a token and refresh token.
```
PeterBaker0, line 63: This is kind of a weird phrasing of this - you are not logging in the user right? You are just retrieving fresh user details based on the token (which implies the user was already logged in?)

web/src/context/auth-provider.tsx
```diff
@@ -0,0 +1,155 @@
...
+        body: JSON.stringify({refreshToken: user.refreshToken}),
+      }
+    );
+
+    if (!response.ok) {
+      setStoredUser(null);
+      setUser(null);
+
+      return {status: 'error', message: 'Refresh token failed'};
+    }
```
PeterBaker0, line 116: Some issues here I think 

a) only try and refresh the token if the current access token is expiring soon (e.g. <1 minute) (saves network traffic)
b) if the refresh fails, don't immediately boot the user out, instead prompt a refresh of the isAuthenticated state based on the existing access token - if it's not expired yet (or within some margin) then let the user continue working and try the refresh again on the next interval

web/src/context/auth-provider.tsx
```diff
@@ -0,0 +1,155 @@
...
+    setUser({...user, token});
+
+    return {status: 'success', message: ''};
+  };
+
+  useEffect(() => {
+    if (isAuthenticated) {
+      refreshToken();
+
+      const intervalId = setInterval(refreshToken, 5 * 60 * 1000);
```
PeterBaker0, line 130: These intervals are configurable in the /app - can you use the same env variable/build config approach for conductor see VITE_REFRESH_TOKEN_INTERVAL etc

PeterBaker0 (approved):
Pending couple of minor comments, lgtm 🚀

web/src/routes/users.tsx
```diff
@@ -0,0 +1,29 @@
...
+ */
+function RouteComponent() {
+  const {user} = useAuth();
+
+  const {data, isPending} = useGetUsers(user);
+
+  return (
+    <DataTable
+      columns={columns}
+      data={data?.map((user: any) => ({...user, email: user.emails[0]}))}
```
stevecassidy, line 25: Getting a crash here for an unprivileged user, 'data?.map is not a function'.  Also, 'user' here is shadowing 'user' from useAuth().

web/src/routes/index.tsx
```diff
@@ -0,0 +1,16 @@
...
+});
+
+/**
+ * RouteComponent component renders the home page.
+ * It displays a card with the home page content.
+ *
+ * @returns {JSX.Element} The rendered RouteComponent component.
+ */
+function RouteComponent() {
+  return <Card>Home</Card>;
```
stevecassidy, line 15: Home should have some content. I created a new home page for old Conductor with links to the web app etc.  Also need to have an option to copy a user token for scripts somewhere - currently on the old conductor home page - that could be in the user profile.

web/src/routes/profile.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * RouteComponent component renders the profile page.
+ * It displays a card with the user's profile information.
+ *
+ * @returns {JSX.Element} The rendered RouteComponent component.
+ */
+function RouteComponent() {
+  const auth = useAuth();
+
```
stevecassidy, line 17: Not sure we need both 'Account' and 'Profile' but you could usefully list the global roles that the user has here (cluster-admin, notebook-creator).  Maybe also what auth method they have used (local password, google).

stevecassidy (commented):
Looking good! See comments about package versions but happy to merge once that is dealt with.

PR #1259, opened 2025-01-17
Support for Offline Maps

stevecassidy:
# Support for Offline Maps

## JIRA Ticket

[BSS-143](https://jira.csiro.com/browse/BSS-143)
[BSS-630](https://jira.csiro.com/browse/BSS-630)

## Description

We need to support offline data collection with maps.  This PR adds support for pre-loading
and using map tiles to display maps rather than requiring an online connection. 

## Proposed Changes

- [x] Add a custom tile source that can cache tiles once downloaded
- [x] Add methods to download a given region of the map for offline use
- [x] Add methods to monitor and manage the storage space used by maps
- [x] Add a user interface to view stored maps and select regions to download

OSM as a tile source does not allow bulk downloads so we switch to a commercial provider (MapTiler). 

Current state seems to be workable since switching to vector tiles.   Downloads are smaller and complete in reasonable time.  Eg. all of Canberra is around 13Mb and downloads in around 30s.  

Alternate map styles are possible, I've included a few options.  All maps in the app now use the same map component.

## How to Test

Configuration needed, ask @stevecassidy for a maptiler key or make your own free account.

```bash
# Set to one of 'osm', 'maptiler', defaults to 'osm'
VITE_MAP_SOURCE=maptiler
# Optional key for the map tile service, eg. maptiler
VITE_MAP_SOURCE_KEY=<your api key>
# Set to true to enable offline map downloads
# note that some services, notably OSM, don't allow map tile downloads
VITE_OFFLINE_MAPS=true
# map style can be 'basic' (default), 'osm-bright', 'openstreetmap', 'toner'
VITE_MAP_STYLE=basic
```

In the app, go to the hamburger menu and select "Offline Maps" This provides the interface
to download regions for offline use.   Pan/Zoom the map to a region of interest and click
Download.   Note the Estimate Size button.

Downloaded regions are listed below the form and will update as tiles are downloaded. 
You should be able to leave the page and return and see the download continue.  

Click on "Show Download Area" for a region to jump to it. 

Go offline (browser tools > Network) and you should still be able to zoom to your downloaded
area and zoom out to see a bigger area. Note that zooming out might show you only the downloaded region at some zoom levels.  Zooming right out should give you a globe map if that were of interest but that is only there for zoom levels 0-2.

Now go to a survey and view a map field (it will need to be centered close to the map you downloaded). You should
be able to browse the map that was downloaded while offline.  Similarly the overview map for a notebook should work if all of the data is within the download region.

## Checklist


app/src/gui/components/map/tile-source.ts
```diff
@@ -0,0 +1,588 @@
...
+      'https://api.maptiler.com/maps/outdoor-v2/{z}/{x}/{y}.png?key={key}',
+    vector: 'https://api.maptiler.com/tiles/v3/{z}/{x}/{y}.pbf?key={key}',
+  },
+};
+
+interface StoredTile {
+  url: string;
+  data: Blob;
+  sets: string[];
+}
```
PeterBaker0, line 56: Would be nice to have this class well commented e.g. what do each of these fields do/refer to?

app/src/gui/components/map/tile-source.ts
```diff
@@ -0,0 +1,588 @@
...
+      }
+      const transaction = this.db.transaction(this.dbName, 'readonly');
+      const store = transaction.objectStore(this.dbName);
+      const request = store.get(query);
+      request.onsuccess = () => {
+        resolve(request.result);
+      };
+      request.onerror = () => reject(request.error);
+    });
+  }
```
PeterBaker0, line 104: Generally this class needs to be well documented given some non-obvious indexedDB stuff going on

app/src/gui/components/map/tile-source.ts
```diff
@@ -0,0 +1,588 @@
...
+  tileDB!: IDB<StoredTile>;
+  tileSetDB!: IDB<StoredTileSet>;
+
+  constructor() {
+    this.initDB();
+  }
+
+  // initDB is called from the constructor but clients may want
+  // to call it directly and wait for it to resolve if they
+  // need to ensure that the databases are ready
```
PeterBaker0, line 175: Can you tie in initDB into the initialisation logic which already exists in the app? Then it can be awaited as the app is loading

app/src/gui/components/map/tile-source.ts
```diff
@@ -0,0 +1,588 @@
...
+   * @param {number} maxZoom The maximum zoom level to estimate the size of.
+   * @return {Promise<number>} The estimated size of the region in MB.
+   */
+  async estimateSizeForRegion(
+    extent: Extent,
+    minZoom: number,
+    maxZoom: number
+  ) {
+    const tileGrid = this.getTileGrid();
+    const average_size = 100; // kb
```
PeterBaker0, line 283: averageSize*

app/src/gui/components/map/tile-source.ts
```diff
@@ -0,0 +1,588 @@
...
+      source: this.source,
+      background: 'hsl(40, 26%, 93%)',
+    });
+    applyStyle(this.tileLayer, getMapStylesheet(MAP_STYLE), {
+      transformRequest: (url: string) => {
+        // TODO: cache these requests...
+        //console.log('transformRequest', url.replace('{key}', MAP_SOURCE_KEY));
+        return url.replace('{key}', MAP_SOURCE_KEY);
+      },
+    });
```
PeterBaker0, line 540: Not sure how this relates to what you were describing re: using JSON stylesheet?

PeterBaker0: i.e. do we still need to do funky caching of the style requests

PeterBaker0 (commented):
As discussed privately - seems to be some reliability issues but on the right track. The download worked well - was fast enough for the region - though the UI could definitely be improved. 

The actual offline functionality as tested in browser with offline dev tools against locally served static files (trying best to replicate a deployed build), would be best described as a bit flaky. It worked sometimes, but sometimes the styles just wouldn't load (default style), other times I was getting console errors related to fetching sprites, there was also a repeatable but not fully consistent bug where after zooming too far out (and map disappears), then rezooming to downloaded region, the map is 'fuzzy' and won't re-load the higher res available offline vectors. This state seems to persist even after going back online. I have sent you a video exhibiting that behaviour.

app/src/gui/components/map/tile-source.ts
```diff
@@ -0,0 +1,588 @@
...
+import VectorTile from 'ol/VectorTile';
+import {MAP_SOURCE, MAP_SOURCE_KEY, MAP_STYLE} from '../../../buildconfig';
+import {applyStyle} from 'ol-mapbox-style';
+import {getMapStylesheet} from './styles';
+import Tile from 'ol/Tile';
+
+// Table of map tile sources for raster and vector tiles
+// based on configuration settings we select which of these to use
+//
+const TILE_URL_MAP: {[key: string]: {[key: string]: string}} = {
```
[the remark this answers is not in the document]
stevecassidy: Object literal keys can only be strings :-(  Since the value comes from the environment we wouldn't catch any more errors that way anyway I don't think.

luke-mcfarlane-rocketlab (commented):
I have encountered a bug where the map renders as a solid white component. This occurs with any notebook (tested a couple of rapid impact surveys) and on chrome.

In order to replicate I:
1. Navigate to a survey
2. Refresh page
3. Throttle internet as offline
4. Open map

If I un-throttle the internet and allow the map to load it uses cache and loads properly even with internet throttled until the next page refresh.

ranisa-gupta16 (commented):
Looking great @stevecassidy , thanks for all the work on this feature.  :-)
Based on review, I was able to successfully download multiple regions for offline use; downloads & the estimated download sizes appear fair and scaled reasonably based on region size and tile count :-)


 - Verified that downloaded maps persisted:
after page refresh, navigating away and back, ssigning out and back in and even after clearing browser cache :-)

- Went offline via DevTools and navigated to the “Offline Maps” section., Initially, previously downloaded map regions were visible. however, upon zooming in/out, the map turned completely white and stopped responding, switching to other downloaded maps also did not load anything further (screenshot below) :-)  

I came back online and then was able to view the maps view again  but it got freezed during offline for me. :-)

Screenshot 2025-03-26 214947

- The current UI doesn't clearly indicate when a download is in progress.
Definitely needs work on UI. Just adding few notes which we can fix later :-)  The "Offline Maps" icon in the sidebar could be more visually representative (e.g., a small map/pin icon) and also observed an UI alignment issue with the support email and policy links at the bottom right (see screenshot)

Screenshot 2025-03-26 214423
Screenshot 2025-03-26 213827

- Another one to consider is I downloaded couple of maps, and went to a survey form, the map location section showed offline wanring (note:  this is all during online not offline mode) 
Not sure if we want to discuss and display more informative warning to the user in map location section of the survey form  .
It stated: "The interactive map is not available while offline and there is no downloaded map covering this location."

See screenshot
Screenshot 2025-03-26 220821
Screenshot 2025-03-26 214604

Overall, I believe this is definitely a complex feature.  There are a few areas for polish around UX feedback, offline zoom handling, and better error recovery/rendering. :-)

app/src/App.tsx
```diff
@@ -200,6 +210,12 @@ export default function App() {
                         }
                       />
                       <Route path={ROUTES.ABOUT_BUILD} Component={AboutBuild} />
+                      {OFFLINE_MAPS && (
+                        <Route
+                          path={ROUTES.OFFLINE_MAPS}
```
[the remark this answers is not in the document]
stevecassidy: Replaced with a map icon.

stevecassidy:
Have limited the minimum zoom level when offline to mean that you hopefully can't zoom out to a blank map.  Fixed the page layout to put the footer in the right place. 

I think if the downloaded map covers a reasonable area (eg. all of Sydney) then this should work better.  Hopefully that's a realistic scenario but needs more testing.

PeterBaker0 (approved):
**Approved pending a few minor things - see below.**

image

Seeing this console error on the offline maps page. 

Also upon refreshing I appear to lose my downloaded maps. I then downloaded another one, and it then showed the other downloaded maps.

_baseline is confusing to the user - don't show it in the downloads if deleting it automatically re downloads it. 

image

Seem to be able to intermittently end up in fuzzy map situations - particularly when jumping around. Resetting the map seems to help. 

When you have multiple maps downloaded it feels confusing to me as a user regarding which map is limiting the zoom. Not sure on ideal behaviour here.

"The interactive map is not available while offlineand there is no downloaded map covering this location. Use the button below to submit your current GPS location." **Missing a space here.**

Seems to work okay overall - just noticing the flakiness of map fuzziness. 

I think a) add a 'Beta' chip or similar to the feature so users know it's in testing b) fix the bug where downloaded maps on the offline maps page don't show until you download a new map c) fix the missing space in the offline map notification d) do a code tidy up as you see fit. 

Then let's merge this to unblock follow up work - it's an improvement from nothing. Thanks - this has been a challenging/frustrating task.

PR #1264, opened 2025-01-21
fix: stopping key file path reading when key source is AWS_KM

PeterBaker0:
# fix: stopping key file path reading when key source is AWS_KM and CSP changes

Previously the function was being called which threw an error if the file path was not specified. Given the file path does not always need to be specified this only calls the underlying function in the key source generation. This happens anyway during build config setup due to the creation of the key source service. This prevents an error in the latest nbic deployment.

Also updating some CSP settings for AWS deployment for designer fixing a bug with notebook upload.

### DIFF
--- a/api/src/services/keyService.ts
+++ b/api/src/services/keyService.ts
@@ -3,9 +3,9 @@ import {importPKCS8, importSPKI, KeyLike} from 'jose';
 import {
   CONDUCTOR_INSTANCE_NAME,
   CONDUCTOR_KEY_ID,
-  CONDUCTOR_PUBLIC_KEY_PATH,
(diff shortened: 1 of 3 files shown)


api/src/services/keyService.ts
```diff
@@ -288,8 +288,9 @@ export function createKeyService(
   switch (keySource) {
```
stevecassidy, line 288: This is fine but just noting that createKeyService is exported but only ever used in this module.

stevecassidy (approved):

api/src/services/keyService.ts
PeterBaker0: ah true - I'll fix that

PR #1268, opened 2025-01-22
chore: developer docs

PeterBaker0:
# chore: developer docs

BSS workshop hackathon - finishing up an idea to restructure docs into user vs developer, and deploy as static webpage. 

@stevecassidy I have tested the existing Fieldmark docs in docs/user still build and work at least locally - wasn't sure of existing workflow to deploy those (couldn't see any action file). 

Not sure on deployment for docs - currently put a simple github pages build, unsure if we are using that already for something else. Let me know what you think.


stevecassidy (approved):
Thanks for this.

PeterBaker0:
Just want to sort out actions before merging

PR #1269, opened 2025-01-23
feat: multi select conditions

PeterBaker0:
# feat: multi select conditions

## JIRA Ticket

[634](https://jira.csiro.au/browse/BSS-634)

## Description

Implements the following designer conditions - necessary for the updated rapid impact (as well as subsequent detailed surveys). 

- contains (the list contains this value)
- does not contain (the list does not contain this value/ it is not selected)
- contains regex (contains at least one selection matching this value)
- does not contain regex (the list does not contain any selection matching this value)

The values in the list and the specified filter value are white space stripped, but I haven't touched any existing work to ensure I don't effect legacy notebooks relying on this non-sanitised behaviour.

## How to Test

I implemented unit tests but they won't run due to broader issues in the app tests. However you can specifically run them with. 

```bash
cd app 
npm i
npx vitest src/conditionals.test.ts
```

I have made a test notebook which I'll attach which demonstrates these features. Everything appears to be working well. I'll also link a video showing it.

## Checklist

### DIFF
--- a/app/src/conditionals.ts
+++ b/app/src/conditionals.ts
@@ -184,7 +184,7 @@ const sanitizeComparisonInput = (input: string): string => {
  * @param value The value to check
  * @returns True iff is an array of strings
  */
-const isStringArray = (value: unknown): value is string[] => 
+export const isStringArray = (value: unknown): value is string[] => 
   Array.isArray(value) && value.every(item => typeof item === 'string');
 
 /**
--- a/app/src/conditionals.ts
+++ b/app/src/conditionals.ts
@@ -169,3 +169,222 @@ registerCompiler('and', (expression: ConditionalExpression) => {
     return () => true;
   }
 });
+
+/**
+ * Trims whitespace in order to sanitize comparisons
+ * @param input The input string to sanitize
+ * @returns The output with whitespace trimmed
+ */
+const sanitizeComparisonInput = (input: string): string => {
+  return input.trim();
+};
+
+/**
+ * This condition checks that the list contains the targeted entry. If an error
+ * occurs or other edge case, returns false to reflect 'not' containing.
+ */
+registerCompiler('contains', (expression: ConditionalExpression) => {
+  return (values: RecordValues) => {
+    if (expression.field && expression.field in values) {
+      try {
+        // cast the types into their desired values but be 'careful'
+        const valuePresent = values[expression.field]! as
+          | string[]
+          | undefined
+          | null;
+        const target = expression.value as string | undefined | null;
+
+        // empty string is okay - be explicit
+        if (target === undefined || target === null) {
+          // The condition has been setup incorrectly - this should be defined
+          return false;
+        }
+
+        // If the list undefined or empty, return false since contains will
+        // always be false
+        if (!valuePresent) {
+          return false;
+        }
+
+        // Try seeing if it contains it
+        if (
+          valuePresent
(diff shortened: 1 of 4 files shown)


PeterBaker0:
[test-contains.json](https://github.com/user-attachments/files/18515367/test-contains.json)

The above is a suitable test notebook for showing the contains logic.

PeterBaker0:
https://github.com/user-attachments/assets/2db0ead9-9fc8-41b2-8dd9-7ca42373aba0

A short video showing this in operation.

app/src/conditionals.ts
```diff
@@ -169,3 +169,230 @@ registerCompiler('and', (expression: ConditionalExpression) => {
...
+ * @returns True iff is an array of strings
+ */
+const isStringArray = (value: unknown): value is string[] => 
+  Array.isArray(value) && value.every(item => typeof item === 'string');
+
+/**
+ * This condition checks that the list contains the targeted entry. If an error
+ * occurs or other edge case, returns false to reflect 'not' containing.
+ */
+registerCompiler('contains', (expression: ConditionalExpression) => {
```
ranisa-gupta16, line 194: Everything looks great overall.
I notice that the` contains` and `does-not-contain` function currently have error handling that silently returnss false or true when unexpected inputs are present.
Will it be helpful to log or throw an error may be for example when non array is passed as value? just thinking it through for production env.

app/src/conditionals.ts
```diff
@@ -169,3 +169,230 @@ registerCompiler('and', (expression: ConditionalExpression) => {
...
+const sanitizeComparisonInput = (input: string): string => {
+  return input.trim();
+};
+
+/**
+ * Checks if the item is a string array
+ * @param value The value to check
+ * @returns True iff is an array of strings
+ */
+const isStringArray = (value: unknown): value is string[] =>
```
ranisa-gupta16, line 187: Do you think it would be helpful to add additional checks for some edge cases in isStringArray like handling empty arrays or unexpected types like `null`, `undefined` within thearray? 
`  -   **Array.isArray(value) && value.every(item => typeof item === 'string' && item != null);**
`

ranisa-gupta16 (commented):
I appreciate how you've provided  thorough test cases.. and test notebook, easy to test. I added a testExpressions functions too to check few consoles and they all worked fine based on logic. It was really helpful to see how thoroughly you’ve thought through the possible conditions
Based on the logic it works fine for me. Just left some clarifying comments . :-)

app/src/conditionals.ts
PeterBaker0: I've added unit tests, these conditions are already covered as the unit tests indicate.

app/src/conditionals.ts
PeterBaker0: As discussed in standup - this would be better handled at a higher level. Right now you can put this condition on any data type. If the user does this, throwing an error results in lots and lots of errors in the console which will get reported in bugsnag. I have silently failed for this reason.

stevecassidy (approved):
We should maybe work out what to do with exceptions in condition evaluation. What you have here is fine for now.  In general I guess we don't want exceptions in condition evaluation to result in hidden fields or any runtime exceptions.  Writing the condition functions to carefully catch errors will probably result in lots of duplicated code so we could perhaps build something into the compiler?

PR #1277, opened 2025-01-28
feat: Designer protected fields 🔒

ryankontos:
# feat: Designer protected fields 🔒

## JIRA Ticket

[<JIRA_TICKET>](https://jira.csiro.au/browse/BSS-641)

---

## Description

This PR introduces support for **protected fields** and **hidden fields** in the Designer.

---

### UI/UX Enhancements:
- **Hidden Fields**:  
  - Users can now mark fields as hidden from the survey.
  - Hidden fields are visually moved to a separate section at the bottom of the field list.
- **Field Protection**:  
  - Fields can now be protected to prevent deletion/modification. Protection states include:
    - **Unprotected**: The field can be freely modified or deleted.
    - **Protected (hiding allowed)**: The field's configuration is locked, but it can be hidden.
    - **Fully Protected**: The field cannot be modified or hidden.
  - A protected field's configuration UI is greyed out, and an explanatory message is displayed above.
- **Required Fields**:  
  - These cannot be hidden or set to the "hiding allowed" protection state.

### Code Changes:
- Added a `_protection_` property to the `FieldType` in the notebook schema. Possible values are:
  - `'none'`
  - `'allow-hiding'`
  - `'protected'`
- Added a `hidden` property (boolean) to the `FieldType` in the notebook schema.
- Updated the `uiSpec` reducer:
  - Added handlers for `toggleFieldProtection` and `toggleFieldHidden`.
  - Prevented the deletion of protected fields.

---

## How to Test

1. **Run the Designer**.
2. **Load a Notebook** and open the **Design** tab (e.g., `FAIMS3-Beta-Demo-Notebook.json` works well).
3. Test the following scenarios:
   - Hide/unhide fields.
   - Toggle field protection states.
   - Attempt to delete/modify fields in various protection states.
   - Ensure required fields cannot be hidden or set to "hiding allowed."

---

## Additional Information

- **Current Limitations**:
  - All users can modify a field's protection settings.  
  - While the popup displays a warning message:  
    _"Only change this if you are sure you have permission"_, the current design assumes user trust.  
  - Future Improvement: Implement template derivation detection to hide protection options where necessary.

---

## Checklist

- [x] All commits are signed.
- [x] Added JSDoc comments for new functions and classes.
- [x] Updated relevant documentation (e.g., READMEs, guides, class comments).


stevecassidy (commented):
Code looks good but the behaviour is not quite what we need.  

Required fields can be marked as hidden - as in we can set allow-hiding on a required field if we want.

In this first iteration, we are just implementing setting these properties, not actually restricting the editing/hiding.  So your hidden toggle is premature.   There will be two modes of editing - here we are editing a template that might be used later to derive a new template.   We set protections on fields here that will apply later when a new template is derived. 

I'd say we move forward from what you have here to implement BSS-642 as well. That adds a property in the template that turns on these editing controls.

PR #1278, opened 2025-01-29
Fix: Designer bug when switching forms

HazeyWazy:
# Fix: Designer bug when switching forms

## JIRA Ticket

[BSS-640](https://jira.csiro.au/browse/BSS-640)

## Description

This PR fixes a bug in the designer where switching from the second section of a form to a form with only one section incorrectly displays an empty form with the message, *"Form has been created. Add a section to get started."*

## Proposed Changes

- **Reset activeStep:**
  - Added a `useEffect` hook to reset `activeStep` to `0` when `viewSetId` changes
- **Conditional Rendering Fix:**  
  - Updated the rendering logic to check `sections.length === 0` instead of relying on `activeStep` to prevent incorrect state interpretation when switching tabs.
  - If `sections.length === 0`, the message *"Form has been created. Add a section to get started."* is displayed.  
  - Otherwise, the `SectionEditor` component renders the active section correctly. 

## How to Test

1. Run Designer.
2. Open a notebook with multiple forms where some have multiple sections.  
3. Go to the *Design* tab, select a form with multiple sections and switch to second section.  
4. Switch to a form with only one section.  
5. Verify that:  
   - The correct section is displayed.  
   - If no sections exist, the appropriate message appears.  
   - No crash occurs when switching forms.  

## Additional Information

The previous condition `activeStep === viewSet.views.length` checks if the active step matches the number of sections in the form. When switching tabs, if `activeStep` isn't reset, it might carry over from the previous form. Since the previous form had more sections than the new one, `activeStep` could be greater than or equal to the number of sections, leading to the incorrect identification of the form as having no sections. This causes the "*Form has been created. Add a section to get started*" message to display, even if the new form contains a section.

## Checklist

### DIFF
--- a/designer/src/components/design-panel.tsx
+++ b/designer/src/components/design-panel.tsx
@@ -55,6 +55,7 @@ export const DesignPanel = () => {
   const maxKeys = Object.keys(viewSets).length;
 
   const handleTabChange = (_event: React.SyntheticEvent, newValue: number) => {
+    console.log('Switching to tab', newValue);
     setTabIndex(newValue.toString());
   };
 
--- a/designer/src/components/design-panel.tsx
+++ b/designer/src/components/design-panel.tsx
@@ -55,7 +55,6 @@ export const DesignPanel = () => {
   const maxKeys = Object.keys(viewSets).length;
 
   const handleTabChange = (_event: React.SyntheticEvent, newValue: number) => {
-    console.log('Switching to tab', newValue);
     setTabIndex(newValue.toString());
   };
 
--- a/designer/src/components/section-editor.tsx
+++ b/designer/src/components/section-editor.tsx
@@ -67,7 +67,7 @@ export const SectionEditor = ({
   );
   const dispatch = useAppDispatch();
 
-  console.log('SectionEditor', viewId, viewSet);
+  console.log('SectionEditor', { viewId, viewSet, fView });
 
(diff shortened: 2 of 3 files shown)


designer/src/components/design-panel.tsx
```diff
@@ -55,6 +55,7 @@ export const DesignPanel = () => {
   const maxKeys = Object.keys(viewSets).length;
 
   const handleTabChange = (_event: React.SyntheticEvent, newValue: number) => {
+    console.log('Switching to tab', newValue);
```
stevecassidy, line 58: Remove this debug output

designer/src/components/section-editor.tsx
```diff
@@ -67,7 +67,7 @@ export const SectionEditor = ({
   );
   const dispatch = useAppDispatch();
 
-  console.log('SectionEditor', viewId, viewSet);
+  console.log('SectionEditor', { viewId, viewSet, fView });
```
stevecassidy, line 70: At some point we should remove this debug output but since it's there for all the main components we'll leave it in for now.

stevecassidy (approved):
Very neat fix. Thanks Htet!  Just one console.log to remove and we can merge.

HazeyWazy:
Removed the debug output that Steve mentioned.

PR #1276, opened 2025-01-28
feat: photo form field redesign

PeterBaker0:
# feat: photo form field redesign

## JIRA Ticket

[BSS-528](https://jira.csiro.au/browse/BSS-528)

[BSS-527](https://jira.csiro.au/browse/BSS-527)

## Description

Redesigns the photo form field visually to 

- When there are no photos attached make it clearer that the user needs to add photos and make the button more prominent
- When there are photos attached, improve the styling and scaling of the photo gallery, refining the on click popup preview with improved button visibility and image scaling
- Make the delete image button more prominent
- Make the add a new photo button when there are existing images more prominent (rather than take photo button)
- Improve un-downloaded attachment user help

And makes some targeted refactor to the actual code i.e. removes the deprecated `atob` usage.

Also improves un-downloaded attachment user help by providing a deep link to the settings tab. Introduces a new fully typed custom hook for managing query string params. E.g. 

```typescript
const {params, setParam} = useQueryParams<{tab: TabIndex}>({
    tab: {
      key: ROUTES.INDIVIDUAL_NOTEBOOK_ROUTE_TAB_Q,
      defaultValue: 'records',
    },
  });
```

In the above you are given a type safe interface into params with a useState style interface. This automatically updates the query string upon change. 

## How to Test

Add photos, remove them, click to preview them. Publish record with photos, then login with different user and browser, and see the un-downloaded attachment behaviour.

## Additional Information

Testing this exposed an existing bug around draft saving and record publishing when editing forms with existing un-downloaded attachments. @stevecassidy has created a JIRA ticket to track this issue.

## Checklist

- [x] I have confirmed all commits have been signed.
- [x] I have added JSDoc style comments to any new functions or classes.
- [x] Relevant documentation such as READMEs, guides, and class comments are updated.

PR #1280, opened 2025-01-29
Highpriority ux form fields improvements

ranisa-gupta16:
## JIRA Ticket

[BSS-650
](https://jira.csiro.au/browse/BSS-650)

## Description
New FAIMSTextField Component:

- Created a custom MUI TextField integrated with Formik.
- Uses FieldWrapper for consistent label & helper text styling.
- Handles text & number inputs dynamically.
- Ensures no undefined values are passed.

FieldWrapper Enhancements:

- Standardized labels, helper text, and required field indicators across form components.
- Added support for required fields with a red asterisk (*
- Will allow scalability for other input components in future

Updated Components to Use FieldWrapper:

- MultiSelect, Select, RadioGroup, TakePhoto rtc
- Integrated FieldWrapper into these components for consistent form layouts.
- Preserved existing functionalities without changes to behavior.
- Improved Required Field Handling:

Updated all relevant components to pass required prop correctly.

- Ensured required fields display an asterisk and validation feedback instantly.
- Prevents users from missing mandatory fields.

- Theme Customization for Help Text:
- Added helpText color to theme palette for better visual distinction.
- Updated theme to use alertcolor for required fields (*)
- Made consitstent styling across all fields touched.

## How to Test

Verify that all form fields in the Restructured Rapid Impact Survey follow a consistent structure in this format
- Heading which is the field name or label 
      - Required asterisk to indicate that this is a  required field
- Subheading which is the helpText
- Input field
Ensure that the newly registered FAIMSTextField is used instead of formik-material-ui TextField.
Validate that required fields display a red asterisk.
The JSON needs to be checked for the new textfield that's registered and updated in designer
For example: 
the below two component namespace and name will have these values fortextfield instead of formik-textfield

  **"component-namespace": "faims-custom",
        "component-name": "FAIMSTextField",**

` "Notes": {
        "component-namespace": "faims-custom",
        "component-name": "FAIMSTextField",
        "type-returned": "faims-core::String",
        "component-parameters": {
          "label": "Site Notes",
          "fullWidth": true,
          "helperText": "Enter any additional comments about this site, if necessary.",
          "variant": "outlined",
          "required": false,
          "InputProps": {
            "type": "text"
          },
          "name": "Notes"
        },
        "validationSchema": [
          [
            "yup.string"
          ]
        ],
        "initialValue": "",
        "meta": {
          "annotation": {
            "include": false,
            "label": "annotation"
          },
          "uncertainty": {
            "include": false,
            "label": "uncertainty"
          }
        },
        "condition": null,
        "persistent": false,
        "displayParent": false
      },
      "Photos": {
        "component-namespace": "faims-custom",
        "component-name": "TakePhoto",
        "type-returned": "faims-attachment::Files",
        "component-parameters": {
          "name": "Photos",
          "helperText": "Capture images relevant to the assessment or observation.",
          "label": "Photos",
          "required": true
        },
        "validationSchema": [
          [
            "yup.array"
          ],
          [
            "yup.of",
            [
              [
                "yup.object"
              ],
              [
                "yup.nullable"
              ]
            ]
          ],
          [
            "yup.nullable"
          ]
        ],
        "initialValue": null,
        "meta": {
          "annotation": {
            "include": false,
            "label": "Comment"
          },
          "uncertainty": {
            "include": false,
            "label": "uncertainty"
          }
        },
        "condition": null,
        "persistent": false,
        "displayParent": false
      },`
      
      
Please refer to screenshots below:  (I have intentionally made few fields required for testing purpose, updated different helptext for testing purpose. )

image
image
image

## Checklist

- [ Y ] I have confirmed all commits have been signed.
- [ y ] I have added JSDoc style comments to any new functions or classes.
- [ y ] Relevant documentation such as READMEs, guides, and class comments are updated.


PeterBaker0 (changes requested):
Code looks great, couple of minor changes requested

Could we ensure the indentation is the same for dropdown multi select and list multi select

image

The location field is not updated yet - but I presume that is covered by an upcoming PR which will update it's display approach so fine with me to leave out of scope. 

image

I think the blue annotation symbol is not displaying in a very nice spot - is this interacting with your field wrapper?

image

Could you make the number field use the wrapper? This is used in one spot in our survey - when it asks for stories of houses in residential dwelling sub type. 

image

We will need to integrate the updated photo field from #1276 into this PR - but trust that will be achievable. Noting that the photo field shows required twice - with the nice red astericks can probably remove the bottom required indicator.

image

ranisa-gupta16:
> Code looks great, couple of minor changes requested
> 
> Could we ensure the indentation is the same for dropdown multi select and list multi select
> 
> image
> 
> The location field is not updated yet - but I presume that is covered by an upcoming PR which will update it's display approach so fine with me to leave out of scope.
> 
> image
> 
> I think the blue annotation symbol is not displaying in a very nice spot - is this interacting with your field wrapper?
> 
> image
> 
> Could you make the number field use the wrapper? This is used in one spot in our survey - when it asks for stories of houses in residential dwelling sub type.
> 
> image
> 
> We will need to integrate the updated photo field from #1276 into this PR - but trust that will be achievable. Noting that the photo field shows required twice - with the nice red astericks can probably remove the bottom required indicator.
> 
> image

ranisa-gupta16:
> @PeterBaker0 - Yes for the location related changes, I'll cover them in location ticket and it's PR.
> 
> * Takephoto component has been now integrated to use fieldwrapper, removed photos required field text below as asterisk is fine for that.
> * Would need to explore this  I think the blue annotation symbol is not displaying in a very nice spot - is this interacting with your field wrapper?) - Not sure how this is getting displayed in UI
> * I'll have a look in the dropdowns indentation
> * Yes you can use the wrapper for the number field too, i have already taken care of that in fieldwrapper logic :-) when I tested rapid survey with this new field it will work just fine.
>   image

@PeterBaker0  - Looked at the annotatation icon and why it's placed like that is because the annotation icon is currently placed in a separate Grid item with xs={6}. This may cause it to misalign with the fields, especially if the field takes full width. Will have a look into this.

designer/src/fields.tsx
```diff
@@ -441,6 +441,23 @@ const fields: {[key: string]: FieldType} = {
     },
     validationSchema: [['yup.object'], ['yup.nullable']],
   },
+  FAIMSTextField: {
+    'component-namespace': 'faims-custom',
+    'component-name': 'FAIMSTextField',
+    'type-returned': 'faims-core::String',
+    'component-parameters': {
```
stevecassidy, line 448: So I'd suggest we simplify the props here to include just those that specify the field content. Remove eg. 'variant' and 'InputProps' as options, those should be decided by our internal stylesheets.  Then also add a Number field that will render as a TextField with number input (not sure if we need that for BSS).

stevecassidy (commented):
Looks really good, like the FieldWrapper. Just one suggestion about the text field.

app/src/gui/component_registry/bundle_components.ts
```diff
@@ -293,6 +294,15 @@ registerField(
   FAIMSTextField
 );
 
+registerField(
+  'faims-custom',
+  'NumberField',
+  'Number Input Field',
+  'Allows users to input numeric values',
+  'Text',
```
PeterBaker0, line 302: Should the 'category' be something else maybe 'Number'?

app/src/gui/fields/NumberField.tsx
PeterBaker0, line 1: Nice - looks clean - will this support default values for the number field?

designer/src/fields.tsx
```diff
@@ -451,13 +451,24 @@ const fields: {[key: string]: FieldType} = {
...
+    'component-namespace': 'faims-custom',
+    'component-name': 'NumberField',
+    'type-returned': 'faims-core::Number',
+    'component-parameters': {
+      label: 'Number Input',
+      fullWidth: true,
+      helperText: 'Enter a valid number',
+      required: false,
+    },
+    validationSchema: [['yup.number'], ['yup.min', 0], ['yup.max', 100]],
```
PeterBaker0, line 469: Why is 100 the max and 0 the min? These should probably be configurable or not provided if not yet supported?

PeterBaker0 (approved):
Looks good - just couple of minor changes to clarify the new number field behaviour e.g. how is a default specified, and loosening the default yup validation.

app/src/gui/component_registry/bundle_components.ts
PeterBaker0: Maybe check with Steve what the intent of this 'category' is

stevecassidy: I think it was to group the fields into similar types but not used at present.  More relevant in Designer where this info is not available (it has it's own list of fields).

designer/src/fields.tsx
stevecassidy: Agreed, just leave the validation as yup.number, although we could use this schema and ensure that the settings are syncronised with the configuration for the input when the field is created in designer.

stevecassidy: Include a min/max option in component-parameters and ensure that they are the same as the values in the validationSchema.  Use them in the number field component to configure the TextField input.

ranisa-gupta16: @PeterBaker0 @stevecassidy 
Thanks for your comments.
I've removed hardcoded min and max lvaues, number field will validate as a number with optional dynamic min/max support if needed in future.
Updated the category, overlooked it earlier. Thanks for the catch. :-)

PR #1281, opened 2025-01-29
Increment version to 1.1.0, remove ios specific version string

stevecassidy:
# Increment version to 1.1.0, remove ios specific version string

## JIRA Ticket

None

## Description

Increments the version number to 1.1.0 for future builds.  

Removes a special case version number for IOS,  now all apps will be on the same version number.

The rationale for doing this now is to avoid confusion with nightly builds pushed to the app stores. These are now clearly pre-release 1.1.0 builds rather than post release 1.0.0. 

## Checklist

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@faims3/api",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "private": true,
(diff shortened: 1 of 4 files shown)


PeterBaker0 (approved):
lgtm

PR #1282, opened 2025-01-30
feat: exclusionary multi-select and options select upgrades

PeterBaker0:
# feat: exclusionary multi-select and options select upgrades

## JIRA Ticket

[BSS-684](https://jira.csiro.au/browse/BSS-684)

## Description

Implements a new feature for multi selects. This allows the designer to specify special option(s) in the multi-select which, when selected, automatically and visually exclude all other selections. A use case for this is:

```
1. Are there any hazards? If so, please select them below 

[gas]
[fuel]
[battery]
...
[unknown]
[none]
```

Selecting unknown or none, should not allow the user to select other responses. For example, this response is inconsistent: 'There are the following hazards: [None, gas]' - which is true? Is there no hazard, or is there gas? 

Previously, we needed two questions 1. Are there hazards? Yes/no/unknown 2. (conditionally if yes) Which hazards. so this is is a streamlined user experience compared to the two questions. 

While in the options editor, I also had some fun redesigning the layout to make it simpler, and also added drag and drop reordering! 

@stevecassidy  I also made the arrow icons consistent.

## Implementation approach

Following suit from the expandedChecklist example, I added a new flag to show this control in the options selector, and added it as an optional field (backwards compatible) to the ElementProps. 

The designer has a column in the new table you can see below

image

The tooltip provides explanation as to what this property does. 

In the client side/app - the exclusive options are now considered/enforced for both the dropdown and expanded check list variants, disabling other options, and applying the behaviour as described.

The drag and drop is implemented using dnd-kit which provides a nice set of hooks and attributes to attach to the table body and draggable entries. This gives css transforms which you can directly put into the sx styling for the attributes.

## How to Test

Use the new designer to build a notebook which has two multi select fields. One which is an expanded check list with some options (mixture of exclusive and non exclusive). One which is the same but non expanded. Ensure functionality works as anticipated in both variants. 

## Checklist


stevecassidy (approved):
LGTM

PR #1286, opened 2025-02-03
feat: hrid redesign, designer settings, template editor and TemplatedStringField changes

PeterBaker0:
# feat: hrid redesign, designer settings, template editor and TemplatedStringField changes

## JIRA Ticket

[BSS-398](https://jira.csiro.au/browse/BSS-398)

## Description

This PR addresses a number of limitations in the existing HRID implementation. Changes included

- implementation of new 'Form settings' module at the top of designer for forms which includes summary fields, layout settings and,
- new `hridField` selection at Form level which allows specific selection of any required text type field as a HRID
- removal of hrid prefix system from designer, and migration for old notebooks to adopt the new style (app is backwards compatible with old hrid prefix fields)
- updates to HRID methods in the app to look for the new HRID first, then fall back to previous logic or `record_id` as the last resort
- upgrades to the TemplatedStringField capability including
  - system context variables including created by and created at which can be injected in the editor and are populated when available at run time
  - (designer) whole new interactive template editor which includes variables, text, if, and unless blocks
  - (designer) template preview capability allowing you to test template resolution with different variable values
- addition of `hidden` prop which allows you to hide a field from display (this is equivalent to a conditionFn excluding it) @stevecassidy this may have implications for export but we can refine this - addition of a panel for designer to toggle this, only enabled on templated string field currently

## How to Test

Test designer by

- creating a text field
- creating a TemplatedStringField which references this field and/or system variables 
- do this manually and using the visual editor - test out the preview
- select this field as the HRID
- test out this works in the app

I'd also check the summary fields - make sure these work + layouts. 

Probably worth checking backwards compatibility with old HRID, and that the migration works - I've tried to test both and seem fine.

## Additional Information

Some corners were cut here, mostly due to the poor implementation quality of the HRID logic and general form record handling. 

i.e. 

- in many places, the fact that hrid was the prefix for the name of the field was heavily depended on in that context to determine which field to use (so I've had to reverse engineer from the ui_spec where available based on the uniqueness of field names between forms)
- the HRID is depended on/used too much in general across the app - there is no need for the HRID to be embedded into the data and depended on, it should be rendered when needed - it is just a transient 'view' over form data which should only be necessary for visualising the data or maybe during export

## Checklist

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -98,6 +98,7 @@
         "@worker-tools/json-stream": "^0.1.0-pre.12",
         "animate.css": "^4.1.1",
         "assert": "^2.0.0",
+        "base64-arraybuffer": "^1.0.2",
         "buffer": "^6.0.3",
         "clsx": "1.2.1",
         "fast-json-stable-stringify": "2.1.0",
--- a/app/package.json
+++ b/app/package.json
@@ -98,7 +98,6 @@
         "@worker-tools/json-stream": "^0.1.0-pre.12",
         "animate.css": "^4.1.1",
         "assert": "^2.0.0",
-        "base64-arraybuffer": "^1.0.2",
         "buffer": "^6.0.3",
         "clsx": "1.2.1",
         "fast-json-stable-stringify": "2.1.0",
--- a/app/src/gui/components/record/branchingLogic.tsx
+++ b/app/src/gui/components/record/branchingLogic.tsx
@@ -44,7 +44,13 @@ export function getFieldsMatchingCondition(
   if (modified.length > 0 || fieldNames.length === 0) {
     // filter the whole set of views
     const result = allFields.filter(field => {
-      return ui_specification.fields[field].conditionFn(values);
+      const fieldDetails = ui_specification.fields[field];
+      return (
+        // Visibility condition function
+        fieldDetails.conditionFn(values) &&
+        // Hidden explicitly in element props - e.g. templated field
+        !fieldDetails['component-parameters']?.ElementProps?.hidden
+      );
     });
     return result;
   } else {
--- a/app/src/gui/components/record/branchingLogic.tsx
+++ b/app/src/gui/components/record/branchingLogic.tsx
@@ -49,7 +49,7 @@ export function getFieldsMatchingCondition(
         // Visibility condition function
         fieldDetails.conditionFn(values) &&
         // Hidden explicitly in element props - e.g. templated field
-        !fieldDetails['component-parameters']?.ElementProps?.hidden
+        !fieldDetails['component-parameters']?.hidden
       );
     });
     return result;
--- a/app/src/gui/components/record/relationships/RelatedInformation.tsx
+++ b/app/src/gui/components/record/relationships/RelatedInformation.tsx
@@ -18,10 +18,8 @@
  *   This is the file is to set the values for persistent state
  */
 import {
-  getFieldToIdsMap,
   getFirstRecordHead,
   getFullRecordData,
-  getHridFieldMap,
   getMetadataForSomeRecords,
   LinkedRelation,
   LocationState,
@@ -36,14 +34,10 @@ import {
 } from '@faims3/data-model';
 import * as ROUTES from '../../../../constants/routes';
 import {logError} from '../../../../logging';
-import {
-  getHridFromValuesAndSpec,
-  ValuesObject,
-} from '../../../../utils/formUtilities';
+import {getUiSpecForProject} from '../../../../uiSpecification';
+import {getHridFromValuesAndSpec} from '../../../../utils/formUtilities';
 import getLocalDate from '../../../fields/LocalDate';
 import {ParentLinkProps, RecordLinkProps} from './types';
-import {getUiSpecForProject} from '../../../../uiSpecification';
-import {projectIsActivated} from '../../../../sync/projects';
 
 /**
  * Generate an object containing information to be stored in
@@ -750,37 +744,6 @@ export async function getParentPersistenceData({
     }
   }
   return parentRecords;
-}
-
-/**
- * Returns the value to be used as the HRID - either the value of a specified
- * field, OR the record_id.
- * @param record_id The record ID which is a fallback HRID
- * @param values The values in the current response
- * @param hridFieldName The specific HRID field name if available
- * @returns The value to be used as the HRID
- */
-function getHRIDValue(
-  record_id: string,
-  values: {[field_name: string]: any} | undefined,
-  // if we know a specific hridFieldName - then use this
-  hridFieldName: string | undefined
-) {
-  if (values === undefined) return record_id;
-
-  // See if we can use the specified hrid field name
-  if (hridFieldName !== undefined && hridFieldName !== '') {
-    if (values[hridFieldName] !== undefined) {
-      return values[hridFieldName];
-    }
-  }
-
-  const possibleHRIDFields = Object.getOwnPropertyNames(values).filter(
-    (f: string) => f.startsWith('hrid')
-  );
-
-  if (possibleHRIDFields.length === 1) return values[possibleHRIDFields[0]];
-  else return record_id;
 }
 
 export async function getDetailRelatedInformation(
--- a/app/src/gui/components/record/relationships/RelatedInformation.tsx
+++ b/app/src/gui/components/record/relationships/RelatedInformation.tsx
@@ -18,28 +18,27 @@
  *   This is the file is to set the values for persistent state
  */
 import {
-  upsertFAIMSData,
-  getFullRecordData,
+  getFieldToIdsMap,
   getFirstRecordHead,
-  ProjectUIModel,
-  RecordReference,
+  getFullRecordData,
+  getHridFieldMap,
+  getIdsByFieldName,
+  getMetadataForSomeRecords,
   LinkedRelation,
   LocationState,
-  Relationship,
-  RecordID,
-  getMetadataForSomeRecords,
   ProjectID,
+  ProjectUIModel,
+  RecordID,
   RecordMetadata,
+  RecordReference,
+  Relationship,
   TokenContents,
-  getHridFieldNameForViewset,
-  getIdsByFieldName,
-  getHridFieldMap,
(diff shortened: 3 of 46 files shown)


PeterBaker0 (commented):
Accidentally reviewed

app/package.json
```diff
@@ -98,6 +98,7 @@
         "@worker-tools/json-stream": "^0.1.0-pre.12",
         "animate.css": "^4.1.1",
         "assert": "^2.0.0",
+        "base64-arraybuffer": "^1.0.2",
```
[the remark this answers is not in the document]
stevecassidy: Consider using Buffer instead (https://www.webdevtutor.net/blog/typescript-atob-deprecated).

designer/src/components/TemplateBuilder.tsx
```diff
@@ -0,0 +1,763 @@
...
+ * Converts a Mustache AST token into our TemplateBlock format
+ */
+const convertMustacheTokenToBlock = (token: any): TemplateBlock | null => {
+  const [tokenType, content, ...rest] = token;
+
+  try {
+    switch (tokenType) {
+      case 'text':
+        return {
+          id: Math.random().toString(36).substring(2, 9),
```
stevecassidy, line 116: This is not guaranteed to be unique, what happens when you get a collision? Since you use this code a lot it should at least be a fn. Suggest just using an incrementing variable to get a unique id.

designer/src/components/TemplateBuilder.tsx
```diff
@@ -0,0 +1,763 @@
...
+          <Grid item xs={12} sm>
+            <FormControl fullWidth>
+              <InputLabel>Variable</InputLabel>
+              <Select
+                value={block.content}
+                label="Variable"
+                onChange={e => updateBlockContent(e.target.value)}
+              >
+                <MenuItem value="">Choose variable...</MenuItem>
+                {allVariables.map(v => (
```
stevecassidy, line 235: You should only be able to add fields from this form to the expression. Doesn't make sense to add a field from another form since it can't have a value here.

designer/src/components/TemplateBuilder.tsx
```diff
@@ -0,0 +1,763 @@
...
+            <Select
+              value={block.type}
+              label="Block Type"
+              onChange={e =>
+                onUpdate({...block, type: e.target.value as BlockType})
+              }
+            >
+              <MenuItem value="text">Text</MenuItem>
+              <MenuItem value="variable">Variable</MenuItem>
+              <MenuItem value="if">Conditional</MenuItem>
```
stevecassidy, line 219: Conditional is not clear. When I make a conditional block I choose a variable but then I'm done - i then need to add another block that gets indented... doesn't seem very clear to me.  Maybe just some helper text will do?

stevecassidy: Also 'Unless' is also a conditional, so maybe 'Conditional' should be 'If'?

designer/src/components/Fields/HiddenToggle.tsx
```diff
@@ -0,0 +1,99 @@
...
+  FormControlLabel,
+  Paper,
+  Stack,
+  Tooltip,
+  Grid,
+} from '@mui/material';
+import {useAppDispatch, useAppSelector} from '../../state/hooks';
+import {FieldType} from '../../state/initial';
+
+/**
```
stevecassidy, line 15: Ryan's PR also adds a hidden toggle, so duplicating work here.

designer/src/fields.tsx
```diff
@@ -402,16 +402,15 @@ const fields: {[key: string]: FieldType} = {
...
-      label: 'Human Readable ID',
+      label: 'Templated String Field',
       fullWidth: true,
-      name: 'hrid-field',
-      helperText: 'Human Readable ID',
+      name: 'templated-field',
+      helperText: 'Templated String Field',
       required: true,
       template: ' {{}}',
       InputProps: {
```
stevecassidy, line 411: InputProps can probably go

app/src/gui/components/record/branchingLogic.tsx
```diff
@@ -44,7 +44,13 @@ export function getFieldsMatchingCondition(
   if (modified.length > 0 || fieldNames.length === 0) {
     // filter the whole set of views
     const result = allFields.filter(field => {
-      return ui_specification.fields[field].conditionFn(values);
+      const fieldDetails = ui_specification.fields[field];
+      return (
+        // Visibility condition function
+        fieldDetails.conditionFn(values) &&
+        // Hidden explicitly in element props - e.g. templated field
+        !fieldDetails['component-parameters']?.ElementProps?.hidden
```
stevecassidy, line 52: Do we really want this to be in ElementProps? I'd say put it in component-parameters directly. I want to get rid of ElementProps and it's friends.

stevecassidy: Ryan's PR puts this in component-parameters

app/src/gui/fields/TemplatedStringField.tsx
```diff
@@ -53,52 +44,14 @@ export class TemplatedStringField extends React.Component<
...
-      if (value !== '')
-        if (this.props.form.errors[this.props.field.name] !== undefined)
-          this.props.form.setFieldError(this.props.field.name, undefined);
-    }
-  }
-
   render() {
     return (
       <>
         <MuiTextField
```
stevecassidy, line 50: Do you want to render this as something other than a text field? We have a task to make it not look like it can be edited. Maybe just a plain text render would be better?

app/src/gui/components/record/relationships/RelatedInformation.tsx
```diff
@@ -674,7 +715,25 @@ export async function getParentPersistenceData(
         )
       );
 
-      const parent_hrid = getHRIDValue(record_id, latest_record?.data);
+      // Here we reverse engineer the correct hrid field to look for since we
```
stevecassidy, line 718: Repeated code from line 576 - factor out?

app/src/gui/components/record/relationships/RelatedInformation.tsx
```diff
@@ -739,7 +819,23 @@ export async function getDetailRelatedInformation(
     values
   );
 
-  const hrid = getHRIDValue(record_id, values);
+  // Here we reverse engineer the correct hrid field to look for since we
```
stevecassidy, line 822: And again, definitely factor out.

app/src/gui/components/record/relationships/RelatedInformation.tsx
```diff
@@ -897,14 +990,17 @@ export async function addRecordLink(
...
+    // here we are trusting that Record has enough of the fields of
+    // RecordMetadata for the purposes of the caller until such time as we
+    // rationalise the Record types
     child_record_meta = latest_record as unknown as RecordMetadata;
     // it's missing the HRID so grab it here
-    child_record_meta.hrid = getHRIDValue(
-      child_record_meta.record_id,
-      child_record_meta.data
-    );
+    // Just use record id here and see what happens ¯\_(ツ)_/¯
```
stevecassidy, line 998: What happens!? I'm dying to know!

stevecassidy: Child links seem to be fine.

stevecassidy: Ah no, so this populates the list of records if you select 'add link to an existing XXX', we now see a list of record IDs rather than the HRIDs. So we really want to have the HRID here.

app/src/gui/fields/TakePhoto.tsx
```diff
@@ -28,6 +28,7 @@ import {Alert, Box, Link, Paper, Typography, useTheme} from '@mui/material';
 import IconButton from '@mui/material/IconButton';
```
stevecassidy, line 28: Why are you fixing TakePhoto in this PR?

app/src/gui/fields/TakePhoto.tsx
```diff
@@ -28,6 +28,7 @@ import {Alert, Box, Link, Paper, Typography, useTheme} from '@mui/material';
...
 import ImageListItem from '@mui/material/ImageListItem';
 import ImageListItemBar from '@mui/material/ImageListItemBar';
+import {decode} from 'base64-arraybuffer';
```
stevecassidy, line 31: See comment about using Buffer.

app/src/logging.tsx
```diff
@@ -125,6 +125,7 @@ export const ErrorPage = () => {
 
 export const logError = (error: any) => {
   if (BUGSNAG_KEY) {
+    console.error(error);
```
stevecassidy, line 128: we don't want error logging in production please!

app/src/utils/formUtilities.ts
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+ * Formats a timestamp into a date-time string in the format "DD/MM/YY H:MMam/pm"
+ *
+ * @param timestamp - Unix timestamp in milliseconds (e.g., from Date.now())
+ * @returns Formatted date-time string or empty string if input is invalid
+ *
+ * @throws Never - Returns empty string for all error cases
+ *
+ * Handles the following edge cases:
```
stevecassidy, line 34: Tests?

app/src/utils/formUtilities.ts
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * Converts the RecordContext into an object mapping from key -> value for use
+ * in template replacement
+ * @param context
+ * @returns The context values to inject
+ */
+function contextToTemplate(context: RecordContext): ValuesObject {
+  const vals: ValuesObject = {};
+  if (context.createdBy) {
+    vals[CREATOR_NAME_ID] = context.createdBy;
```
stevecassidy, line 101: Should there be a default? 'unknown user'?

stevecassidy (commented):
Generally good with some suggestions for improvement and a few things to fix.  Ok if the suggestions become future tasks.

stevecassidy:
I'm happy with this now. Very thorough work as usual!

PR #1290, opened 2025-02-05
fix: various misc build fixes CDK

PeterBaker0:
Missing double quotes around export causing truncated environment variables. 

Missing environment variables. 

Small fry... 

giphy


stevecassidy (approved):
LGTM

PR #1291, opened 2025-02-05
Fix/designer helper text overflow

ryankontos:
# Fix/designer helper text overflow

## Description

Wasn't assigned this, so feel free not to approve. But:  I wanted to quickly fix the field header UI issues when the window is too narrow or the helper text is too long. This change does increase field header height which may not be ideal over a more adaptive  approach, but for now I think it is a good fix.

## Proposed Changes

UI tweaks in the field header:
- The chips and helper text now appear below the field title, I don't know if this looks _as good_ but I think it looks fine, and gives us a lot more width.
- This resolves overlapping between the action buttons and helper text.

## How to Test

- Load a template in the designer, open design tab.
- Resize the window, ensuring the elements in the field headers do not overlap or generally look bad.
- Modify helper text, making sure that the helper text is not allowed to take uo more than 3 lines.

## Checklist


PeterBaker0 (commented):
I made some minor tweaks but this looks nice to me - happy to approve and I'll merge since I'm cleaning up a bunch of designer PRs

PR #1293, opened 2025-02-06
fix: my vs other record listing, pouch, sync, drafts

PeterBaker0:
# fix: listing bug, my vs other, pouch synchronisation and management issues

## JIRA Ticket

[BSS-708](https://jira.csiro.au/browse/BSS-708)
[BSS-712](https://jira.csiro.au/browse/BSS-712)

## Description

Fixes logical issue with record filtering for my vs other responses. 

Moved record listing logic into reusable custom hook which is smarter about caching, memoisation etc. 

Makes draft and normal record management more sensible with custom hooks for both.

Moves the record fetching to the top level so it can be done once and memoised.

Fixes the issue which was that at no point was the record table actually being configured to properly filter based on mine vs others. 

Added auto refresh every 10 seconds for the list to ensure it stays fresh.

Fixes or at least mitigates this bug https://jira.csiro.au/browse/BSS-712 

This was principally two main bugs

1. the options {push:{}, pull:{}} were used as a trigger to indicate two way sync - this was missing in token refresh, meaning new sync connections were one way
2. the pouch sync and db objects were not being cleaned up well (causing listener leaks) and also unstable behaviour with multiple syncs on the same local DB

And also propose this improvement for future work. 

https://jira.csiro.au/browse/BSS-713

Also fixes a bug with drafts - I'm not even sure what the bug was, but the implementation was so unnecessarily complicated I unified with record approach, removing some subscription logic while there.

## How to Test

User 1: create records
User 2: make sure you can see them 

Then for the above, try this under many conditions i.e. let tokens refresh, let the tab go to sleep, try refreshing the app, or leaving it un refreshed for a long time, you should continue to observe reliable syncing for both users, two way.

## Checklist

### DIFF
--- a/app/src/gui/components/notebook/index.tsx
+++ b/app/src/gui/components/notebook/index.tsx
@@ -40,7 +40,7 @@ import MetadataRenderer from '../metadataRenderer';
 import CircularLoading from '../ui/circular_loading';
 import AddRecordButtons from './add_record_by_type';
 import DraftTabBadge from './draft_tab_badge';
-import DraftsTable from './draft_table';
+import {DraftsTable} from './draft_table';
 import {OverviewMap} from './overview_map';
 import RangeHeader from './range_header';
 import {RecordsTable} from './record_table';
@@ -152,6 +152,7 @@ export default function NotebookComponent({project}: NotebookComponentProps) {
     projectId: project.project_id,
     filter: 'all',
   });
+  const forceDraftRefresh = drafts.refetch;
 
   /**
    * Handles the change event when the user switches between the Records and Drafts tabs.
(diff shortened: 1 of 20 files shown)


app/src/gui/components/notebook/index.tsx
```diff
@@ -359,41 +359,35 @@ export default function NotebookComponent({project}: NotebookComponentProps) {
...
+                  key={'allrecordstable'}
                   project_id={project.project_id}
                   maxRows={25}
                   viewsets={viewsets}
                   filter_deleted={true}
                   handleRefresh={handleRefresh}
                   onRecordsCountChange={handleCountChange}
                   recordLabel={recordLabel}
+                  // Everyone's records
+                  myRecordsOnly={false}
```
stevecassidy, line 387: Should this show all records or all records that aren't mine?

stevecassidy (approved):
Apart from that one question this looks good.

app/src/sync/databases.ts
```diff
@@ -157,177 +157,233 @@ export const metadata_dbs: LocalDBList<ProjectMetaObject> = {};
...
+  return (
+    JSON.stringify(existingRemote.info) === JSON.stringify(newConnectionInfo)
+  );
+}
+
+/**
+ * Creates a new synchronisation (PouchDB.sync) between the specified local and
+ * remote DB. This uses the preferred sync options to avoid misconfiguration in
+ * caller functions. Two way sync is always enabled, with live and retry. If
+ * attachment filter is supplied, the pull sync options are overrided with a
```
stevecassidy, line 314: overridden

stevecassidy (approved):
Much enhanced goodness!

PR #1297, opened 2025-02-10
fix: templated string field character escaping

PeterBaker0:
# fix: templated string field character escaping

## JIRA Ticket

[BSS-714](https://jira.csiro.au/browse/BSS-714)

## Description

For our use case, just disabling the escape function is sufficient. 

See code comment explaining risks assoc.

## Checklist

### DIFF
--- a/app/src/utils/formUtilities.ts
+++ b/app/src/utils/formUtilities.ts
@@ -7,6 +7,26 @@ import {
 import Mustache from 'mustache';
 import {RecordContext} from '../gui/components/record/form';
 
+/*
+Patch mustache to not escape values.
+
+This addresses JIRA BSS-714 where Observation/Point of Interest was being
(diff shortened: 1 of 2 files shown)


app/src/utils/formUtilities.ts
```diff
@@ -7,6 +7,26 @@ import {
...
+/*
+Patch mustache to not escape values.
+
+This addresses JIRA BSS-714 where Observation/Point of Interest was being
+rendered as "Observation&#x2F;Point of Interest"
+
+This is generally a risky approach but safe enough in our use case provided the
+output is never:
+
+- Inserted into the DOM using .innerHTML
```
stevecassidy, line 19: Should be ok since we only use the value through regular react value insertion.

stevecassidy (approved):

PR #1303, opened 2025-02-10
Fix-csv-json-export

stevecassidy:
# Fix CSV export format, missing columns

## JIRA Ticket

[BSS-441](https://jira.csiro.com/browse/BSS-441)
[BSS-632](https://jira.csiro.com/browse/BSS-632)

## Description

Exporting CSV files where there is a map form field that is empty didn't have enough commas to line up the columns. 

Also fixed up finding the HRID of a record when exporting (used to name photo files) according to the new HRID code.  

Add photo filenames to the JSON record export.

## How to Test

Create a record in BSS without entering a map location and another with a location.  From Conductor, export CSV and note that there are three empty columns for the missing location and that all subsequent fields line up properly.

Add photos to your records and note the photo filenames in the CSV and corresponding ZIP download contain the proper HRID of the record. 

Get the JSON export at /api/notebooks/<project-id>/records  and note the proper filenames in the Photo field.

## Checklist

### DIFF
--- a/api/src/api/notebooks.ts
+++ b/api/src/api/notebooks.ts
@@ -67,7 +67,8 @@ import * as Exceptions from '../exceptions';
 import {requireAuthenticationAPI} from '../middleware';
 
 import patch from '../utils/patchExpressAsync';
-import {slugify} from '../utils';
+import {generateTokenContentsForUser, slugify} from '../utils';
+import {generateUserToken} from '../authkeys/create';
 
 // This must occur before express api is used
 patch();
@@ -245,14 +246,13 @@ api.get(
     if (!req.user || !userHasPermission(req.user, req.params.id, 'read')) {
       throw new Exceptions.UnauthorizedException();
     }
-    const token = {
-      roles: req.user.roles,
-      server: slugify(CONDUCTOR_INSTANCE_NAME),
-      username: req.user.user_id,
-      // Five minutes from now
-      exp: Date.now() + 1000 * 60 * 5,
-    };
-    const records = await getRecordsWithRegex(token, req.params.id, '.*', true);
+    const tokenContent = generateTokenContentsForUser(req.user);
+    const records = await getRecordsWithRegex(
+      tokenContent,
+      req.params.id,
+      '.*',
+      true
+    );
     if (records) {
       const filenames: string[] = [];
       // Process any file fields to give the file name in the zip download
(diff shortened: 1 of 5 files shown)


api/src/api/notebooks.ts
```diff
@@ -244,14 +245,42 @@ api.get(
...
-    const records = await getNotebookRecords(req.params.id, {
+    const token = {
       roles: req.user.roles,
       server: slugify(CONDUCTOR_INSTANCE_NAME),
       username: req.user.user_id,
       // Five minutes from now
       exp: Date.now() + 1000 * 60 * 5,
-    });
+    };
+    const records = await getRecordsWithRegex(token, req.params.id, '.*', true);
```
PeterBaker0, line 255: get records with regex with .* is probably the same as get all records? Isn't there already that function?

api/src/couchdb/notebooks.ts
```diff
@@ -566,39 +562,6 @@ export const validateNotebookID = async (
...
-};
-
-const getRecordHRID = (record: any) => {
-  for (const possible_name of Object.keys(record.data)) {
-    if (possible_name.startsWith(HRID_STRING)) {
-      return record.data[possible_name];
-    }
-  }
-  return record.record_id;
-};
```
PeterBaker0, line 600: I see you removed - okay - reasoning?

api/test/backup.test.ts
```diff
@@ -61,16 +65,19 @@ describe('Backup and restore', () => {
...
-      // throw in a test of getNotebookRecords while we're here
-      const records = await getNotebookRecords(
+      // throw in a test of getRecordsWithRegex while we're here
+      const token = {
+        roles: user.roles,
+        server: slugify(CONDUCTOR_INSTANCE_NAME),
+        username: user.user_id,
+        // Five minutes from now
+        exp: Date.now() + 1000 * 60 * 5,
+      };
```
PeterBaker0, line 75: maybe we could make this a util function for testing/ i.e. buildUserToken or similar

PeterBaker0 (commented):
Looks improved, couple of issues

The Location_longitude column has a single quote at the start `'149.14524984130858` on excel - not sure what's going on there. 

Also I can't extract the ZIP for attachments, seeing this error 

image

Can you replicate that?

api/src/api/notebooks.ts
stevecassidy: Not one that returns the kind of record that we want...

api/src/couchdb/notebooks.ts
stevecassidy: It was only used in this module, getRecordsWithRegex now returns full records, when I wrote this it didn't so I had to go fetch every record separately.

PR #1304, opened 2025-02-10
fix: refresh projects updates ui specification

PeterBaker0:
# fix: refresh projects updates ui specification

Fixes https://jira.csiro.au/browse/BSS-719 

Issue was that the sync projects never considers the metadata DBs for activated surveys. 

Adds a specific method which does the minimal work to 'refresh' the metadata DB. Also does some code tidying and hides some buttons which by definition do not work offline.

## Checklist


stevecassidy (approved):
Looks good.

PR #1316, opened 2025-02-13
feat: configurable rate limiting with reasonable defaults

PeterBaker0:
# feat: configurable rate limiting with reasonable defaults

## JIRA Ticket

https://jira.csiro.au/browse/BSS-744

## Description

Adds configuration in the typical style for rate limit window, events and enable/disable. 

Sets much more generous defaults which evaluate over a longer period. Rate limiting 60 req/minute was excessive - this meant that a request rate of 1/second was too much. This sets it to 1000 requests per 10 minutes which allows for bursty loads while still protecting the API. We would need a more advanced configuration or possibly limiter to catch very high burst loads over shorter durations without punishing normal usage.

## Checklist

### DIFF
--- a/api/.env.dist
+++ b/api/.env.dist
@@ -52,3 +52,12 @@ REFRESH_TOKEN_EXPIRY_MINUTES=2880
 # Access expires in 5 minutes
 ACCESS_TOKEN_EXPIRY_MINUTES=5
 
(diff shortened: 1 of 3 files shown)


stevecassidy (approved):
LGTM

PR #1317, opened 2025-02-13
fix: theme error on build for bssTheme

PeterBaker0:
# fix: theme error on build for bssTheme

Stepper color theming approach was causing a 'cannot read property of undefined' error in the production BSS build. 

The implementation approach appears unnecessarily complex, so I've moved the properties directly into the theme and updating the theme type. This works and is validated to be a fix @ faims.bss.nbic.cloud.


app/vite.config.ts
```diff
@@ -31,6 +31,7 @@ export default defineConfig({
   base: '/',
   build: {
     outDir: 'build',
+    sourcemap: 'inline'
```
stevecassidy, line 34: Do you want to put that in here -  would prefer it in a separate PR.

stevecassidy (approved):
Good apart from the sneaky inline sourcemap config.

app/vite.config.ts
PeterBaker0: Reverted- will put in separate PR

PR #1321, opened 2025-02-14
build: debugging config CDK

PeterBaker0:
See change - default debugging settings.


stevecassidy:
Can you check the size of the APK build with this enabled? If it doesn't add too much weight then it's ok otherwise we should look at alternate ways of making it available.  You can upload the sourcemap to Bugsnag or just host it somewhere on the web for example.

PeterBaker0:
@stevecassidy  this is not the sourcemap - this is just debugging env variables

stevecassidy (approved):

PR #1324, opened 2025-02-17
Offline maps position fallback

stevecassidy:
# Offline maps position fallback

## JIRA Ticket

[BSS-747](https://jira.csiro.com/browse/BSS-747)

## Description

When offline, the map field doesn't work. As an interim measure until we have offline maps, show a button to take the current position instead. 

## Proposed Changes

Adds a button to record the current position instead of the usual map control if we're offline.

## How to Test

Load a notebook with a map control, go offline (via the dev tools network tab) and observe the map button changing.   Click to record current location.  Go online again and note that the point recorded is correct.

## Checklist


PeterBaker0 (approved):
Works well - just styling tweaks as shown

PR #1314, opened 2025-02-13
feat: survey invites 📩

luke-mcfarlane-rocketlab:
## JIRA Ticket

[BSS-665](https://jira.csiro.au/browse/BSS-665)

## Description

Added the ability to view and create new invitations to projects.

## Proposed Changes

- updated form component to allow select fields
- added API endpoints to get and create invites for a given notebook
- created hook to get invites
- created dialog with form for creating a new invite
- created invites table

## How to Test

1. Navigate to a project
2. Click the "Invites" tab
3. Click "Create Invite" button
4. Verify that invites table has updated
5. Test code or link on conductor platform

## Additional Information

Still to complete:
- QR code
- Delete/archive/restore logic

## Checklist

- [x] I have confirmed all commits have been signed.
- [x] I have added JSDoc style comments to any new functions or classes.
- [x] Relevant documentation such as READMEs, guides, and class comments are updated.

PR #1326, opened 2025-02-18
fix: activation of new notebook after refresh

PeterBaker0:
# fix: activation of new notebook after refresh

## JIRA Ticket

[BSS-715](https://jira.csiro.au/browse/BSS-715)

## Description

Fixes bug in which surveys could not be activated when they had arrived from a refresh survey list update. 

This was due to the `projects_dbs[listing_id]` not being updated during this action. 

This change is stupid and should not be necessary - the issue is that there are now two places in which this data is tracked!! The projects_db in the context management, and also in projects_dbs for the listing. They also have different data structures and data models. 

This whole thing needs a rewrite to unify the project/listing management to a single store. And there is literally no reason for it to be a pouch DB. 

## Proposed Changes

Through code surgery/shenanigans pulls out the logic which syncs the projects_dbs code from the process-initialisation file and reuses it in the context management code. 

## How to Test

Have no surveys. Open app, login. Add a survey on conductor. Refresh survey list. Activate the new survey that appears, ensure it works. 

## Checklist

### DIFF
--- a/app/src/sync/process-initialization.ts
+++ b/app/src/sync/process-initialization.ts
@@ -168,6 +168,55 @@ function delete_listing_by_id(listing_id: ListingID) {
 }
 
 /**
+ * This updates/creates records in the projects_dbs database for the given
+ * listing - used during initialisation and refresh
+ */
+export async function syncProjectDb({
+  projects,
+  listingId,
+  conductorUrl,
+}: {
+  projects: ProjectObject[];
+  listingId: string;
+  conductorUrl: string;
+}) {
+  // get the project DB
+  const projectDb = projects_dbs[listingId]?.local;
+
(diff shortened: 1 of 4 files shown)


PeterBaker0:
This includes #1314  woops!! Please merge that first

app/src/sync/process-initialization.ts
```diff
@@ -344,6 +366,7 @@ async function get_project_from_directory(
 
   if (listing_id in projects_dbs) {
     const project_db = projects_dbs[listing_id];
+    console.log(await project_db.local.allDocs());
```
stevecassidy, line 369: Console.log

stevecassidy (approved):
Looks like a reasonable fix. one debug log to remove

PR #1327, opened 2025-02-18
feat: back button display and intercept to safely alert draft creation

ranisa-gupta16:
# feat: back button display and intercept to safely alert draft creation

## JIRA Ticket

BSS-718

-As discussed with Elana:
- Added a back button that will take user back to the survey (where records are dispalyed that user has created)

New updates: 
- Used BackButton component to display a back to records button from the record creation phases.
- Improved styling for back button to be displayed  like "back to surveys" button to maintain consistency
      - Back button will be displayed like before in record draft view - in wrapped state as previosuly.
- Added android back button effect to detect when user click back button from device, if yes then popup is shown to alert user that there's a draft being saved.
- Back button from record creation also introduces a popup for user to know that there's draft being saved for that record.
image
image
image

## Checklist

- [y ] I have confirmed all commits have been signed.
- [ y] I have added JSDoc style comments to any new functions or classes.
- [y ] Relevant documentation such as READMEs, guides, and class comments are updated.

### DIFF
--- a/app/src/gui/pages/record-create.tsx
+++ b/app/src/gui/pages/record-create.tsx
@@ -155,6 +155,7 @@ interface DraftRecordEditProps {
   record_id: RecordID;
   state?: any;
   location?: Location;
+  onBack: () => void;
 }
 
 function DraftRecordEdit(props: DraftRecordEditProps) {
@@ -226,6 +227,9 @@ function DraftRecordEdit(props: DraftRecordEditProps) {
   return (
     <React.Fragment>
       <Grid container justifyContent={'space-between'} spacing={2}>
+        <Grid item>
+          <BackButton label="Back" onClick={props.onBack} />
+        </Grid>
         <Grid item xs>
           <ProgressBar percentage={progress} />
         </Grid>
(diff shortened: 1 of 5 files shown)


app/src/gui/pages/record-create.tsx
```diff
@@ -310,9 +345,44 @@ export default function RecordCreate() {
...
+      event.preventDefault();
+      setAndroidBackPressed(true);
+      setOpenDialog(true);
+    };
+    window.history.pushState(null, '', window.location.href);
+    const handlePopState = () => {
+      setOpenDialog(true);
+    };
+    window.addEventListener('popstate', handlePopState);
+    window.addEventListener('beforeunload', handleAndroidBackPress);
```
PeterBaker0: this is not just android back button, right? Just to be clear with naming maybe 'handleBackEvent' rather than 'handleAndroidBackButtonPress' ?

PeterBaker0: Was here so just updated

PeterBaker0 (approved):
Made some tweaks to styling as discussed, but works nicely :) If you're happy with my changes please merge once linting etc passes

app/src/gui/pages/record-create.tsx
ranisa-gupta16: Yes agree!
Initially added it only for android but it seems to work with all.

PR #1329, opened 2025-02-18
feat: data export 💿

luke-mcfarlane-rocketlab:
## JIRA Ticket

[BSS-582](https://jira.csiro.au/browse/BSS-582)

## Description

Add the ability to export survey data and images.

## Proposed Changes

- Created function to download files from blob
- Added export form/dialog for data and images
- Updated export tab to use new export dialogs

## How to Test

1. Navigate to a survey (project)
2. Click the "Export" tab
3. Click "Data/Photo Export"
4. Select a view and click "Submit"

## Checklist


PeterBaker0 (approved):
Code looks good

Can you change "View" to "Form" 

image

Also noting that export is very slow (!) even for only like 50 records (as is the case with the Rapid Impact v1.1). It's taking 10 seconds or so. No way it would scale to thousands of records. 

@stevecassidy  we might need to look into optimising the export code - or even longer term simplifying the DB structure to prioritise performance for bulk operations (e.g. read lots, write lots) over richness of provenance (with AVPs etc). I'm not convinced it will scale.

Otherwise all good 👍

PR #1334, opened 2025-02-20
feat: admin forgot password reset workflow

PeterBaker0:
# feat: admin forgot password reset workflow

## JIRA Ticket

[BSS-758](https://jira.csiro.au/browse/BSS-758)

## Description

Implements a new workflow for resetting your password. 

The workflow is as follows

1. Cluster-admin initiates a password reset by specifying an email (really a userId)
2. This generates a DB persisted one-time use expiring code which is stored hashed in the DB (so access to DB cannot grant ability to use the codes) 
3. This code is returned both directly to the user once, and also as a URL
4. The user navigates to this link which is in the new conductor, opening a public route which allows the user to specify a new password
5. This sends the code + new password to the API, which updates the local strategy password for that user

The user experience makes it easy to share this link by presenting a copyable link display + QR code which can be enlarged by clicking it.

Also adds an admin only 'forceInit' endpoint which forcefully updates design docs etc (disabling the 'if the doc already exists, don't write it again' check) - this is useful since this change involves updates to the design doc which are otherwise not able to be programmatically migrated. 

## Migrations 

**Required changes!**

### CouchDB migration

This functionality will not work until you apply the following migration

1. login as cluster admin to old conductor
2. copy the bearer token
3. update `/api/.env` with `USER_TOKEN=<>`
4. setup npm env `cd /api && npm i` 
5. run the force init action `npm run forceinitdb`

Try generating a code and make sure the workflow is functional.

### Deployment configuration

Add the new `NEW_CONDUCTOR_URL` environment variable to the conductor/api build. This is already done in the `aws-cdk` IaC - however in existing deployments you may need to update your build configuration.

**Note** ensure this includes the https:// prefix!

**This is a mandatory property and not including it will break the deployment**. 

## How to Test

Follow the workflow above through the new conductor interface. check that the code cannot be used twice and that invalid codes are rejected (noting you can change the query string arg in the URL). Check the updated password works, and that the old password doesn't.

## Additional Information

- right now, external IdP users who reset their password will actually have the local strategy added - there is a ticket to track this behaviour and make it more intentional https://jira.csiro.au/browse/BSS-772
- couldn't get the data-model to work in /web - tried adding to package.json, turbo building, etc etc but it just wouldn't pick them up (in the build, the VS code interpreter liked them) https://jira.csiro.au/browse/BSS-774

### Generating a code

navigate to users tab of new conductor and select the reset password button

image

Then share the URL or the QR code

image

### Resetting password

Navigate to the URL via QR code or direct link.

Enter your new password and submit

image

## Checklist

### DIFF
--- a/api/.env.dist
+++ b/api/.env.dist
@@ -63,4 +63,4 @@ RATE_LIMITER_PER_WINDOW=1000
 RATE_LIMITER_ENABLED=true
 
 # What is the URL of the new conductor?
-NEW_CONDUCTOR_URL=http://localhost:5173
+NEW_CONDUCTOR_URL=http://localhost:3001
--- a/api/.env.dist
+++ b/api/.env.dist
@@ -61,3 +61,6 @@ RATE_LIMITER_PER_WINDOW=1000
 
 # Enable/disable rate limiting (default: true)
 RATE_LIMITER_ENABLED=true
+
+# What is the URL of the new conductor?
+NEW_CONDUCTOR_URL=http://localhost:5173
--- a/api/api_testing/passwordreset.http
+++ b/api/api_testing/passwordreset.http
@@ -1,0 +1,76 @@
+@resetBaseUrl = {{$dotenv BASE_URL}}/api/reset
+@token = {{$dotenv API_TOKEN}}
+
+### Initiate password reset (unauthorized - no token)
+POST {{resetBaseUrl}}
+Content-Type: application/json
+
+{
+    "email": "test@example.com"
+}
+
+### Initiate password reset
+# @name initiateReset
+POST {{resetBaseUrl}}
+Authorization: Bearer {{token}}
+Content-Type: application/json
+
+{
+    "email": "test@gmail.com"
+}
+
+### Complete password reset (invalid code)
+PUT {{resetBaseUrl}}
+Content-Type: application/json
+
+{
+    "code": "INVALID_CODE",
+    "newPassword": "newSecurePassword123!"
+}
+
+### Complete password reset (success)
+# Uses the code from the previous initiate request
+PUT {{resetBaseUrl}}
+Content-Type: application/json
+
+{
+    "code": "{{initiateReset.response.body.code}}",
+    "newPassword": "newSecurePassword123!"
+}
+
+### Complete password reset (code already used)
+
+# Attempts to use the same code again
+PUT {{resetBaseUrl}}
+Content-Type: application/json
+
+{
+    "code": "{{initiateReset.response.body.code}}",
+    "newPassword": "anotherNewPassword456!"
+}
+
+### Complete password reset (password too short)
+PUT {{resetBaseUrl}}
+Content-Type: application/json
+
+{
+    "code": "{{initiateReset.response.body.code}}",
+    "newPassword": "short"
+}
+
+### Initiate password reset (invalid email format)
+POST {{resetBaseUrl}}
+Authorization: Bearer {{token}}
+Content-Type: application/json
+
+{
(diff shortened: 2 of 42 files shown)


jyucsiro (commented):
Both reset via QR code and via link works.

Tested Generate new link button in screen with QR code. I expected the new link to be generated on the spot rather than switching to a previous dialog and clicking "Generate Reset Link" again. Is this the expected behaviour?

Minor Suggestions:
	- Users list page: Rows per page could be 10 or 20 instead of the default 5

PeterBaker0:
> Both reset via QR code and via link works.
> 
> Tested Generate new link button in screen with QR code. I expected the new link to be generated on the spot rather than switching to a previous dialog and clicking "Generate Reset Link" again. Is this the expected behaviour?
> 
> Minor Suggestions: - Users list page: Rows per page could be 10 or 20 instead of the default 5

Okay I'll action those changes

jyucsiro (approved):
lgtm

ranisa-gupta16 (approved):
I have tested the password reset, via link and via QR code both, it worked really well.
I also tried couple of scenarios like setting the same password again and it gave me the error: "Error Resetting password" which is great. 
In terms of the time - it took me hardly 1.5 minutes to reset my password both ways.
Image (1)

READY TO MERGE. YAYYY..

api/.env.dist
```diff
@@ -60,4 +60,7 @@ RATE_LIMITER_WINDOW_MS=600000
 RATE_LIMITER_PER_WINDOW=1000
 
 # Enable/disable rate limiting (default: true)
-RATE_LIMITER_ENABLED=true
\ No newline at end of file
+RATE_LIMITER_ENABLED=true
+
+# What is the URL of the new conductor?
+NEW_CONDUCTOR_URL=http://localhost:5173
```
stevecassidy, line 66: Should add this to web/vite.config.js if you want this to work consistently, vite will use 5172 if it's the first to start. Actually suggest something else like 6001 to avoid clashing with other vite projects.

api/api_testing/passwordreset.http
```diff
@@ -0,0 +1,76 @@
+@resetBaseUrl = {{$dotenv BASE_URL}}/api/reset
```
stevecassidy, line 1: These could be actual tests!

api/test/emailReset.test.ts
```diff
@@ -0,0 +1,162 @@
+import {expect} from 'chai';
```
stevecassidy, line 1: Ah, excellent!

app/src/gui/components/record/view.tsx
```diff
@@ -356,28 +356,31 @@ export function ViewComponent(props: ViewProps) {
                 Other sections have errors. Click to navigate:
               </Typography>
 
-              {otherSectionsWithErrors.map((section, _index) => (
```
stevecassidy, line 359: Does this belong in here?

stevecassidy (approved):
Code wise seems ok apart from some changes that seem to be from some other PR. Haven't tested.

api/test/emailReset.test.ts
PeterBaker0: 💪

api/.env.dist
PeterBaker0: Good call

app/src/gui/components/record/view.tsx
PeterBaker0: GitHub is being silly - I was fixing all those lint warnings! It' just I added the comment and it thinks I deleted everything :)

PR #1336, opened 2025-02-21
Fix overview map re-render

stevecassidy:
# Fix overview map re-render

## JIRA Ticket

[BSS-749](https://jira.csiro.com/browse/BSS-749)

## Description

The overview map would re-render every 10s and reset the  map to the initial position.

## Proposed Changes

- Added 'memo' to the overview map component to prevent re-renders if the props don't change. 
- Added a custom equality comparison to the useRecords query to prevent re-renders if the list of records is the same as last time
- Pass records from the parent into overview map so it doesn't have to get them itself.

## How to Test

View the overview map and zoom/pan the map, note that it doesn't reset after 10s.

## Checklist

### DIFF
--- a/app/src/gui/components/notebook/index.tsx
+++ b/app/src/gui/components/notebook/index.tsx
@@ -147,8 +147,6 @@ export default function NotebookComponent({project}: NotebookComponentProps) {
   });
   const forceRecordRefresh = records.query.refetch;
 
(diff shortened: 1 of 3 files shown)


PeterBaker0 (approved):
Works well - only thing is when you click on an item it re-centers - but consider that a separate issue.

PR #1337, opened 2025-02-21
feat: conditional field deletion warning

ryankontos:
# feat: conditional field deletion warning

## JIRA Ticket

[<JIRA_TICKET>](https://jira.csiro.au/browse/BSS-695)

## Description

This PR introduces a warning dialogue when attempting to delete a field that is referenced in:  
- A **field condition**
- A **section condition**
- A **templated string field**

This ensures that deleting a field **does not break existing conditions or fields**.

## Proposed Changes

- **Added a validation check** before deleting a field
- **Implemented **`findFieldUsage()`** in `field-editor.tsx` to detect references to a field in:
  - **Field conditions**
  - **Section conditions**
  - **Templated string fields**
- **Implemented **`isFieldUsedInCondition`** in `condition.tsx` to recursively check if a given field is used within a given condition.
- **Modified field deletion logic** to prevent deleting fields if they are referenced

## How to Test

1. **Go to the Designer UI**  
2. **Create a new field** (e.g., `animal_field`)
3. **Use it in one or more of the following**:
   - A **field condition** (e.g., `Show if animal_field == "Elephant"`)
   - A **section condition**
   - A **TemplatedStringField** (e.g., `"The animal is {{animal_field}}"`)
4. **Try deleting the field**  
   - Expected result: A **warning** should appear, listing the places where the field is used.  
   - If the field is **not used anywhere**, it should be deleted normally.

## Additional Information

- This change **prevents unintended data loss or errors** by ensuring users are warned before deleting a referenced field.
- The warning lists **all references** to the field so users can update conditions or templates accordingly.

## Checklist


designer/src/components/field-editor.tsx
```diff
@@ -89,6 +90,41 @@ export const FieldEditor = ({
 
   const fieldComponent = field['component-name'];
 
+  const allFields = useAppSelector(
+    state => state.notebook['ui-specification'].fields
+  );
+  const allFviews = useAppSelector(
```
stevecassidy, line 96: You already have this above in `views`

designer/src/components/field-editor.tsx
```diff
@@ -149,6 +178,48 @@ export const FieldEditor = ({
     }
   };
 
+  /**
+   * Finds where a field is used in conditions or templated string fields
+   */
+  const findFieldUsage = (
+    fieldName: string,
+    allFields: Record<string, any>,
```
stevecassidy, line 186: Cool, I've not come across this before, very useful. Shame about the name though since we have many Record* types!

designer/src/components/field-editor.tsx
```diff
@@ -149,6 +178,48 @@ export const FieldEditor = ({
     }
   };
 
+  /**
+   * Finds where a field is used in conditions or templated string fields
+   */
+  const findFieldUsage = (
```
stevecassidy, line 184: This is good but probably belongs in condition.tsx, might be useful in other places.

designer/src/components/field-editor.tsx
```diff
@@ -331,6 +399,37 @@ export const FieldEditor = ({
...
+            <Button
+              onClick={e => {
+                e.stopPropagation();
+                setDeleteWarningOpen(false);
+              }}
+            >
+              Cancel
+            </Button>
+            <Button onClick={confirmDelete} color="error">
+              Delete Anyway
```
stevecassidy, line 429: I think I would prefer to forbid deleting the field if it is used in conditions - ask them to go delete the conditions first. Then we don't end up with invalid notebooks.

stevecassidy (changes requested):
Generally good, just a few comments and I think a change in behaviour - don't let them delete a field that is used.

designer/src/components/field-editor.tsx
ryankontos: Funny you say that, I originally had it in condition.tsx and decided to move it when I added template support

stevecassidy (approved):
Looks good

PR #1338, opened 2025-02-24
fix: improper couch DB url due to sync logic

PeterBaker0:
# fix: improper couch DB url due to sync logic

Fixes critical bug where activated notebooks had `localhost:5984` as their URL due to the `ProjectsObject` being misconfigured. This was introduced in the project activation fix (after refresh) - the issue is that `update_directory` and `get_projects_from_conductor` was doing more work which was not covered by the new logic. 

This adds a `update_directory` call to the syncProjects hook, and removes the previous logic which was overwriting the DB entries with improperly formed data on first initialisation. 

This was masked due to the actual couch DB of a local testing deployment matching the fallback. 

Added error/warning to log outputs to catch this in the future.

### DIFF
--- a/app/src/context/projects-context.tsx
+++ b/app/src/context/projects-context.tsx
@@ -130,7 +130,6 @@ export function ProjectsProvider({children}: {children: ReactNode}) {
     // Build project list (proposed update) from this merge
     const newProjects = [...newProjectsMap.values()];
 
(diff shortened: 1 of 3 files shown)


stevecassidy (approved):
'seems ok' - tested against remote server and things seem to work.

PR #1343, opened 2025-02-26
feat: breadcrumb names 🍞

luke-mcfarlane-rocketlab:
## JIRA Ticket

[BSS-780](https://jira.csiro.au/browse/BSS-780)

## Description

Added mapping for template/project names in breadcrumbs.

## Proposed Changes

- added mapping for template/project names in breadcrumbs
- added loading skeleton while query is running

## How to Test

1. Navigate to a project/template and verify the breadcrumbs are displaying the name

## Additional Information

Name will default to id if the query fails.

## Checklist


web/src/components/breadcrumbs.tsx
```diff
@@ -7,8 +7,41 @@ import {
...
+ *
+ * @param {string} name - The name of the breadcrumb.
+ * @param {any} data - The data for the breadcrumb.
+ * @param {boolean} isPending - The loading state of the breadcrumb.
+ * @returns {JSX.Element} The rendered breadcrumb with hook.
+ */
+const breadcrumbWithHook = (name: string, data: any, isPending: boolean) => {
+  if (isPending) return <Skeleton className="w-full h-4 rounded-full" />;
+
+  return <div>{data?.metadata?.name || data?.template_name || name}</div>;
```
ranisa-gupta16, line 43: Should we display something more user-friendly if both" metadata.name" & "template_name"  are missing? Or are we okay just displaying the URL segment?”

web/src/components/breadcrumbs.tsx
```diff
@@ -7,8 +7,41 @@ import {
...
+ * @returns {Promise<Data>} The data for the path.
+ */
+const useGetData = (user: User | null, pathname: string[]) => {
+  if (pathname.length < 2) return {data: null, isPending: false};
+
+  if (pathname[0] === 'projects') return useGetProjects(user, pathname[1]);
+  if (pathname[0] === 'templates') return useGetTemplates(user, pathname[1]);
+
+  return {data: null, isPending: false};
+};
```
ranisa-gupta16, line 29: if the api call fails for some reason or returns an error like network erorr, unauthorized... we should display a dedicated error message  like "error fetching project name or something " 
we can expose an "error" or "isError" from useGetProjects/useGetTemplates - and then in breadcrumbHook return that error div simply.

ranisa-gupta16 (changes requested):
Breadcrumbs looks good to me , just left few comments for fallback error state when the api call fails + few others to help catch issues at compile time. 
Thanks :-)

web/src/components/breadcrumbs.tsx
luke-mcfarlane-rocketlab: in this case it is the project/template id. This is the only thing available if the query fails as it is extracted from the url.

web/src/components/breadcrumbs.tsx
luke-mcfarlane-rocketlab: I'm not sure the user needs to know if there is an error fetching the project name in the breadcrumb - what would they do with that information? The app will work fine just displaying the project id

luke-mcfarlane-rocketlab:
@ranisa-gupta16 have added the type definition you suggested above. Please see replies for other requests.

web/src/components/breadcrumbs.tsx
ranisa-gupta16: That sounds good. Let's stick to the url as fallback.

ranisa-gupta16 (approved):
Thanks for those changes Luke! This lgtm!

ranisa-gupta16 (changes requested):
Would need to address few errors I am seeing now after pulling latest.

image

image
Switch between content and management sections of side nav, click on user section and you should be able to replicate these errors.

Might be because of a early return statement , would be related to the hook. Please have a look. Thanks !

luke-mcfarlane-rocketlab:
@ranisa-gupta16 resolved errors and fixed conflicts with main

ranisa-gupta16 (approved):

PR #1345, opened 2025-02-27
feat: archive template 📁

luke-mcfarlane-rocketlab:
## JIRA Ticket

[BSS-664](https://jira.csiro.au/browse/BSS-664)

## Description

Add the ability to archive templates.

## Proposed Changes

- added API endpoint for archiving a template
- added couchDB function to archive a template
- updated couchDB create new template function to have initial status of 'active'
- added dialog/action for archiving template
- added template status to template table

## How to Test

1. Navigate to template table
2. Open a template that is not 'archived'
3. Navigate to actions
4. Archive the template
5. Navigate back to template table and verify it is archived

## Additional Information

I have not implemented any changes to how the front-end/back-end changes based on whether templates are archived or not, this is just the ability to update the status.

## Checklist


ranisa-gupta16 (commented):
I didn't quite get the intention of what we're trying to do here. It doesn't work for me as per the PR description.

On click of "Archive" , I see a popup which shows this, 
image
If I click "Archive template" , it should set that template to read-only as per the PR description but I am still able to see it, edit it?  Also the flow  isn't user-friendly. We'd need t show user something visually that the template has been archived. 
Might need to discuss this further with team.

luke-mcfarlane-rocketlab:
> I didn't quite get the intention of what we're trying to do here. It doesn't work for me as per the PR description.
> 
> On click of "Archive" , I see a popup which shows this, image If I click "Archive template" , it should set that template to read-only as per the PR description but I am still able to see it, edit it? Also the flow isn't user-friendly. We'd need t show user something visually that the template has been archived. Might need to discuss this further with team.

Have added functionality so that archived templates are unable to be edited or used to create projects. Note that this is only blocked from a UI perspective - I have not updated the API.

ranisa-gupta16:
> > I didn't quite get the intention of what we're trying to do here. It doesn't work for me as per the PR description.
> > On click of "Archive" , I see a popup which shows this, image If I click "Archive template" , it should set that template to read-only as per the PR description but I am still able to see it, edit it? Also the flow isn't user-friendly. We'd need t show user something visually that the template has been archived. Might need to discuss this further with team.
> 
> Have added functionality so that archived templates are unable to be edited or used to create projects. Note that this is only blocked from a UI perspective - I have not updated the API.

Thanks for that Luke.

web/src/components/dialogs/archive-template-dialog.tsx
```diff
@@ -0,0 +1,87 @@
...
+ * @returns {JSX.Element} The rendered ArchiveTemplateDialog component.
+ */
+export const ArchiveTemplateDialog = ({archived}: {archived: boolean}) => {
+  const {user} = useAuth();
+  const {templateId} = Route.useParams();
+  const queryClient = useQueryClient();
+  const [open, setOpen] = useState(false);
+
+  const onClick = async () => {
+    const response = await fetch(
```
ranisa-gupta16, line 29: Can you please add try-catch block here to handle any errors

api/src/api/templates.ts
```diff
@@ -211,3 +212,29 @@ api.post(
     res.sendStatus(200);
   }
 );
+
+// Archive a template
+api.put(
+  '/:id/archive',
```
ranisa-gupta16, line 218: instead of api.put('id/archive',... , shoudn't we do 
api.put('/templates/:id/archive', ...) ? that way it'll align with Restful standards or patterns we use

web/src/components/dialogs/edit-template.tsx
```diff
@@ -23,15 +29,30 @@ import {useState} from 'react';
 export const EditTemplateDialog = () => {
   const {user} = useAuth();
   const {templateId} = Route.useParams();
-
   const {data} = useGetTemplates(user, templateId);
-
   const [open, setOpen] = useState(false);
+  const archived = data?.metadata.project_status === 'archived';
```
ranisa-gupta16, line 34: archived state is derived from meta data but we're using multiple times in the comp, can we use  "useMemo" here so it nly l re-run when "data"changes?

ranisa-gupta16: like 
const archived = useMemo(() => data?.metadata.project_status .......

web/src/components/dialogs/project-from-template.tsx
```diff
@@ -10,18 +10,52 @@ import {Button} from '../ui/button';
...
+  return archived ? (
+    <TooltipProvider>
+      <Tooltip>
+        <TooltipTrigger>
+          <Button
+            variant="outline"
+            disabled={data?.metadata.project_status === 'archived'}
+          >
+            Create {NOTEBOOK_NAME_CAPITALIZED}
+          </Button>
```
ranisa-gupta16, line 43: I think the logic for disabling the btn based on archived state is repeated in multiple places, can we etract this into a helper function please. like isTemplateArchived ..... and that can be used where needed. Thanks!

ranisa-gupta16 (changes requested):
Looking much better now Luke. Now as a user I know that template section has a status and it's easy to understand that it's in read-only mode.
Thanks for that
Just couple of more quick comments to address and we should be good to go. 
Many thanks!

api/src/api/templates.ts
luke-mcfarlane-rocketlab: this is defined on api/templates/:id/archive - you can check how this file is used in the main api files.

luke-mcfarlane-rocketlab:
@ranisa-gupta16 resolved all comments and merged with main

ranisa-gupta16 (approved):
lgtm thanks for all updates! :-)

PR #1347, opened 2025-02-27
feat: servers, projects, dbs, activation and deactivation, metadata overhaul

PeterBaker0:
# feat: servers, projects, dbs, activation, metadata overhaul

## Description

Completely overhauls the app's handling of 

- listings (now called servers)
- projects (i.e. surveys, notebooks etc)
- activation
- local DBs, syncs and remote DBs
- sync attachments setting
- notebook metadata
- notebook ui spec

Does this by building from the ground up a redux-store slice named 'projects' which manages all of this. 

It does this through a combination of 

- persisted serialisable state (for servers, projects, config etc)
- services which index in-memory instantiated objects which don't persist through app refresh and are not serialisable (e.g. pouch DB, pouch sync, compiled UI specs (since these have JS in them))
- reducers which act on the state, handling side effects where appropriate (e.g. activateProject is a single well defined action which does everything in one place)
- asynchronous thunks which, in one place, handle interacting with the APIs to merge details into the existing store, overwriting and even updating connection details on live DBs where suitable
- selectors which help with finding the right part of the state

Uses redux-persist to store data in local storage (works on all platforms). Expands the initialisation gate to safely dispatch all restore ops (e.g. generating DBs based on config, compiling UI specs) before the app loads. 

Removes **all** DB sync/events logic - this could have implications for one spot I could see which is the auto head merging occurring to presumably minimise conflicts(?). I tested conflicts and it still seemed to work - so unsure what this actually does. 

Also did a bunch of misc tidy up while deep in the code, including stripping out the annoying 'split project' vs 'unique project' etc etc which is not necessary - the app now knows which server a project refers to in all places and where only a project ID was available, I supplemented the routes to include the server ID.

Removes the metadata-db (there was no need for this to be a database, its just a static set of key value pairs which are one way sync'd from the API).

## New features

- deactivation
- proper handling of stop/resume sync

## TODO 

- [x] fix unit tests (for API at least. probably data model too)
- [ ] (deferred) migration approach
- [ ] (ongoing) thorough testing
- [ ] (deferred) conflict investigation
- [ ] (deferred) replace instances of `return <></>` with a suitable real return value
- [x] linting
- [x] createdAt and lastUpdated properties of Project
- [x] isSyncing - what happens when you disable this?

## How to Test

Test the entire app! There **will** be bugs.

## Checklist


api/src/api/notebooks.ts
```diff
@@ -203,7 +205,11 @@ api.get(
     const metadata = await getNotebookMetadata(project_id);
     const uiSpec = await getNotebookUISpec(project_id);
     if (metadata && uiSpec) {
-      res.json({metadata, 'ui-specification': uiSpec});
+      res.json({
+        metadata,
+        // silly business
+        'ui-specification': uiSpec as unknown as Record<string, unknown>,
```
stevecassidy, line 211: Don't we have a type for uispec? ProjectUIModelDetails. I see your zod type is just a record...

app/src/context/slices/projectSlice.ts
```diff
@@ -0,0 +1,1986 @@
...
+        serverId: payload.serverId,
+      });
+      const localDb = createLocalPouchDatabase<ProjectDataObject>({
+        id: localDatabaseId,
+      });
+      databaseService.registerLocalDatabase(localDatabaseId, localDb);
+
+      // creates the remote database (pouch remote)
+      const {db: remoteDb, id: remoteDbId} =
+        createRemotePouchDbFromConnectionInfo<ProjectDataObject>(
```
stevecassidy, line 570: deactivate

stevecassidy: and ditto below

stevecassidy, line 570: Danger!!! What if this is the only copy, I'd be happier if we didn't delete it here or only deleted if we had assurance that the data is safe. Can we just mark this project as deactivated and keep the database until such time as we have a watertight check on the safety of the data?

app/src/context/slices/projectSlice.ts
```diff
@@ -0,0 +1,1986 @@
...
+        throw new Error(
+          `Cannot update connection when we don't know the couchDBUrl. Server ID: ${payload.serverId}. Project ID: ${payload.projectId}`
+        );
+      }
+
+      // check the project exists
+      const project = projectByIdentity(state, payload);
+      if (!project) {
+        // abort
+        throw new Error(
```
stevecassidy, line 750: After reading this for the nth time I wonder if you can't factor all of the checks out to a helper fn?

app/src/context/slices/projectSlice.ts
```diff
@@ -0,0 +1,1986 @@
...
+  // What is the URL for the server?
+  serverUrl: string;
+
+  // Server unique ID
+  serverId: string;
+
+  // Display title
+  serverTitle: string;
+
+  // What is the URL of the couch database for this server?
```
stevecassidy, line 189: This is lovely!

app/src/drafts.ts
```diff
@@ -17,34 +17,3 @@
  * Description:
  *   TODO
  */
```
stevecassidy, line 19: Delete this file?

app/package.json
```diff
@@ -113,6 +113,7 @@
     "markdown-it": "^14.1.0",
     "moment": "2.29.4",
     "mustache": "4.2.0",
+    "nano": "^10.1.4",
```
stevecassidy, line 116: Why the added dep on nano? Doesn't seem to be used (and we have pouchdb for that)

api/src/api/notebooks.ts
PeterBaker0: Defer - see https://jira.csiro.au/browse/BSS-340

library/data-model/src/utils.ts
```diff
@@ -256,33 +250,3 @@ export const getFieldToIdsMap = (
 
   return fieldMap;
 };
-
-/**
- * Gets the raw ui spec (no compiled conditionals) for the given project ID
- * @param projectId The project ID to fetch the ui spec for
- * @returns The ui specification (without compiled conditionals) or undefined if unavailable
- */
-export async function getUiSpecForProject({
```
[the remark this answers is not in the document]
PeterBaker0: Believe I've removed it - can't find any mention o fit

app/src/context/slices/projectSlice.ts
```diff
@@ -0,0 +1,1986 @@
...
+  buildCompiledSpecId,
+  buildPouchIdentifier,
+  buildSyncId,
+  createLocalPouchDatabase,
+  createPouchDbSync,
+  createRemotePouchDbFromConnectionInfo,
+  fetchProjectMetadataAndSpec,
+  getRemoteDatabaseNameFromId,
+  SyncEventHandlers,
+} from './helpers/databaseHelpers';
```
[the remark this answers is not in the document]
PeterBaker0: Documentation has been updated, and not sure which exports you are referring to here.

app/src/context/slices/helpers/databaseHelpers.ts
```diff
@@ -0,0 +1,373 @@
...
+  remoteId,
+}: {
+  localId: string;
+  remoteId: string;
+}): string => {
+  return `${localId}-${remoteId}`;
+};
+
+/**
+ * Builds an identifier for the compiled spec service
```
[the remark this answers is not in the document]
PeterBaker0: They are used in projectSlice.ts

app/src/context/slices/projectSlice.ts
```diff
@@ -0,0 +1,1986 @@
...
+        ProjectIdentity & {
+          syncState: Partial<SyncState>;
+        }
+      >
+    ) => {
+      const {projectId, serverId, syncState} = action.payload;
+
+      // Check the server exists
+      const server = serverById(state, serverId);
+      if (!server) {
```
stevecassidy, line 1344: I'm seeing a crash here where localDBId is not set, localDB has a value though and it looks like a db id.

app/src/context/slices/projectSlice.ts
PeterBaker0: I don't think you can because they also behave as type assertions to the type checker. If you put into a separate function typescript is not smart enough to see that.

app/package.json
PeterBaker0: Accident

app/src/context/slices/projectSlice.ts
PeterBaker0: Refactor causes this issue due to you having the old version of the store. As discussed.

app/src/context/slices/helpers/databaseHelpers.ts
```diff
@@ -0,0 +1,373 @@
...
+   */
+  paused?: (err?: Error) => void;
+
+  /**
+   * Fired when the replication starts actively processing changes;
+   * e.g. when it recovers from an error or new changes are available
+   */
+  active?: () => void;
+
+  /**
```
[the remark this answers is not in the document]
PeterBaker0: done!

stevecassidy:
Tests fine on IOS - no problem pointing it to existing server.   One final suggestion - a confirmation dialog on deactivation might be good to prevent accidental button pushes.

stevecassidy (approved):
Yay!

PR #1348, opened 2025-02-27
feat: Improve select field condition UI

ryankontos:
# feat: Improve select field condition UI

## JIRA Ticket

[<JIRA_TICKET>](https://jira.csiro.au/browse/BSS-697)

## Description

Previously, when creating a condition in Designer, the value field was always just a text entry. If the targeted field is a select/multiselect/radiogroup then we already know what the options can be. This PR modifies the condition editor modal to offer an option list to choose from instead. 

## Proposed Changes

- Modified the condition modal to show options in the value select field for fields where there are multiple options.
- Added a warning alert that shows within the AccordionDetail area of a field with predefined values. The alert shows when that field is used within a condition, but the condition expects a value that is not one of the possible options.

## How to Test

- In a new notebook, create some fields with predefined options (Select, RadioGroup, MultiSelect, Checkbox)
- On another field, try to create a condition. The options should be visible in the value text field as a dropdown.
- Try to change one of the options after doing this and see if the warning alert appears.
- Continue to experiment with various edge cases-- I probably missed something!

## Checklist


designer/src/components/condition.tsx
```diff
@@ -423,11 +628,174 @@ const FieldConditionControl = (props: ConditionProps) => {
...
+    if (!fieldDef) return false;
+    return ['Select', 'RadioGroup', 'MultiSelect', 'Checkbox'].includes(
+      fieldDef['component-name']
+    );
+  };
+
+  /* Filter the allowed operators based on field type */
+  const getAllowedOperators = (fieldDef: FieldType | null) => {
+    return isPredefinedOptions(fieldDef)
+      ? ['equal', 'not-equal']
```
stevecassidy, line 647: If the target field is a multi-select the value will be an array so we need 'contains'. I note that we don't have 'does not contain' as a comparator so we should add that too. 
so this should be ['equal', 'not-equal'] or ['contains', 'does not contain'].

stevecassidy: Ah, we do have does-not-contain so it should appear in your longer list as well.

designer/src/components/condition.tsx
```diff
@@ -423,11 +628,174 @@ const FieldConditionControl = (props: ConditionProps) => {
...
+            sx={{minWidth: 200}}
+            error={
+              !selectedValues.every(v =>
+                possibleOptions.some((opt: any) => opt.value === v)
+              )
+            }
+          >
+            <InputLabel>Value</InputLabel>
+            <Select
+              multiple
```
stevecassidy, line 702: You're allowing me to choose multiple values here but then the condition won't work since it will say 'x equals [y, z]' or 'x contains [y, z]', I need an explicit OR if I want to check for alternate values - unless we have a 'contains one of' operator, which we could have...

designer/src/components/condition.tsx
```diff
@@ -423,11 +628,174 @@ const FieldConditionControl = (props: ConditionProps) => {
...
+            <Select
+              label="Value"
+              value={
+                isValidOption
+                  ? String(condition.value)
+                  : (String(condition.value) ?? 'false')
+              }
+              onChange={e => updateValue(e.target.value === 'true')}
+            >
+              <MenuItem value="false">False</MenuItem>
```
stevecassidy, line 743: Make these 'Checked' and 'Not checked'.

stevecassidy (commented):
Overall looks pretty good, a few minor tweaks.

stevecassidy:
Just noticed that when I use a checkbox field in a condition it gives me an error 'invalid value 'false'' - I can't choose a valid value.

PR #1377, opened 2025-03-19
fix: template actions crash 💥

luke-mcfarlane-rocketlab:
## JIRA Ticket

[BSS-840](https://jira.csiro.au/browse/BSS-840)

## Description

Fixed a bug where the page would crash when switching between templates with the action tab open.

## Proposed Changes

- added optional chaining to the prop calls from the data query to avoid error throwing when null
- added disabled prop to button when data has not loaded

## How to Test

1. Navigate to a template
2. Open the actions tab
3. Switch between templates using the sidebar with the actions tab open

## Checklist


ranisa-gupta16 (approved):
Great job identifying and preventing the crash when switching between templates while the actions tab is open. Crashes no longer occur for me after testing :-)  Thanks for the fix :-)

PR #1378, opened 2025-03-19
fix: darkmode red 🚗

luke-mcfarlane-rocketlab:
## JIRA Ticket

[BSS-836](https://jira.csiro.au/browse/BSS-836)

## Description

Updated the destructive color in dark theme to be more visible (used in warnings and error messages)

## Proposed Changes

- updated the destructive color in dark theme

## How to Test

- yurn on dark mode in the upper right corner of the screen
- navigate to a destructive template/project action (such as update/archive)
- verify red is more visible

## Checklist

### DIFF
--- a/app/src/gui/components/record/form.tsx
+++ b/app/src/gui/components/record/form.tsx
@@ -79,7 +79,6 @@ import {
 } from './relationships/RelatedInformation';
 import UGCReport from './UGCReport';
 import {getUsefulFieldNameFromUiSpec, ViewComponent} from './view';
-import {c} from 'vitest/dist/reporters-5f784f42';
 
 type RecordFormProps = {
   navigate: NavigateFunction;
@@ -795,7 +794,7 @@ class RecordForm extends React.Component<any, RecordFormState> {
     const allSections =
       this.props.ui_specification.viewsets[this.getViewsetName()].views;
 
-    const allVisited = allSections.every((section: string) =>
+    allSections.every((section: string) =>
       this.state.visitedSteps.has(section)
     );
   }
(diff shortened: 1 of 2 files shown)


luke-mcfarlane-rocketlab:
> To be honest, the current red warning block feels very harsh on the eyes and creates visual stress, especially against a dark theme image
> 
> the bright red background (--destructive) makes the warning feel like a critical system failure. Suggestions: Let's use a softer warning style instead of full red. for example : Dark mode: bg-yellow-900/20 + text-yellow-300 Light mode: bg-yellow-50 + text-yellow-700 just change this as per your current coding setup in css file. and may be add an informative icon like AlertTriangle or similar for clarity Just wanna make it more appealing to the eyes of the user :-) Happy to discuss further if needed. :-)

I don't really agree with this.

"the current red warning block feels very harsh on the eyes and creates visual stress, especially against a dark theme" - it used to be a more muted red (which was my go to red for destructive actions) but feedback was that it was hard to see so I have updated it to be more visible (its now very close to the destructive actions colour used in git).

"the bright red background (--destructive) makes the warning feel like a critical system failure." - I think its very normal for platforms to use red to signify a destructive action. Users shouldn't really see (or be aware of) system failures.

"Suggestions: Let's use a softer warning style instead of full red. for example : Dark mode: bg-yellow-900/20 + text-yellow-300 Light mode: bg-yellow-50 + text-yellow-700 just change this as per your current coding setup in css file. " - as above, I don't think destructive should be anything but red. I could setup `--warning` and `--warning-foreground` variables but they would likely need to be orange as yellow would be hard to read.

"may be add an informative icon like AlertTriangle or similar for clarity Just wanna make it more appealing to the eyes of the user" - sure thing we can add an icon.

ranisa-gupta16:
We can add the color change later, I am approving this to go ahead :-) @luke-mcfarlane-rocketlab

PR #1385, opened 2025-03-24
change record list ordering to be last updated instead of created

ranisa-gupta16:
## JIRA Ticket

[BSS-810
](https://jira.csiro.au/browse/BSS-810)

## Description

Improve record list ordering to be "last updated" instead of created

## How to Test
Record list should be sorted as per last updated not created.

Navigate to the survey
Create a record or update existing ones, the ordering the records in the table should be based on which record was last updated not created.

The latest updated record should be at the top of the recordlist.

image

## Checklist


jyucsiro (commented):
Tested this and it works. Records are sorted as last updated only. User isn't able to change the ordering.

Just a question about whether this is the desired UI - what if users want to sort by ascending order of updated or by another field?

ranisa-gupta16:
> Tested this and it works. Records are sorted as last updated only. User isn't able to change the ordering.
> 
> Just a question about whether this is the desired UI - what if users want to sort by ascending order of updated or by another field?

Thanks for reviewing Jono :-)
And yes you're exactly right, we need sorting for the user for sure, they should be able to sort however they want. I have just fixed it, Can you please review it again. Many thanks :-)

jyucsiro (approved):
Sorting is working now. Initial sort on last updated appears when list is presented. User can click to sort on fields asc and desc.

PR #1387, opened 2025-03-24
feat: Designer Undo/Redo

ryankontos:
# feat: Designer Undo/Redo

## JIRA Ticket

[BSS-829](https://jira.csiro.au/browse/BSS-829)

## Description

Adds Undo/Redo functionality to Designer.

## Proposed Changes

Designer
- Wraps the uiSpecReducer in the "undoable" reducer from "redux-undo".
- Replaced uses of `state.notebook['ui-specification']` with `state.notebook['ui-specification'].present`.
- Adds undo/redo buttons to design tab of designer.
- Adds keyboard shortcuts (OS-specific) for undo/redo.

## How to Test

- Run designer.
- Ensure the undo/redo buttons in the design tab are greyed out as there should be nothing to undo or redo.
- Load a notebook. The buttons should still be greyed out.
- Attempt to undo and redo with the keyboard shortcuts. A `snackbar` component should appear informing you that there is nothing to undo or redo.
- Make changes to the notebook and ensure that they are undoable/redoable  using the buttons and keyboard shortcut.
- Test that the snackbar appears when undoing and redoing.

## Checklist


PeterBaker0 (commented):
Good starting point - I fixed some bugs in the implementation (i.e. the migration was causing a doubly nested ui spec - and also you don't need to expose migration logic to the fact that the reducer is history wrapped-  keep that in the store only - it should be a transparent abstraction for the client). 

The main issue here is performance - until we debounce the text input, this is very very bad for performance as you are saving the entire notebook to memory on every keystroke! And you need to store for all historical versions. Awaiting #1415 to be merged, then can reassess. It's very cool though!

PeterBaker0 (approved):
Seems to work reasonably well - little bit laggy locally but it always is laggy - much much better with debounce of 500ms. Worth merging and seeing how it holds up. Well done.

PR #1407, opened 2025-04-02
Improve database initialisation and migration

stevecassidy:
# Improve database initialisation and migration

## JIRA Ticket

None

## Description

Database initialisation and migration were a bit confusing after the recent PRs introducing migration.   This
PR simplifies the process to a single step 'migrate'.

## Proposed Changes

Removed the old initdb and forceinitdb targets which worked through the API.  

Updated the migrate scripts to  work for both new installs and databases from previous versions by
moving the creation of the admin user to after db migration.  

The only step that is needed now for a new install or an updated version is to run `npm run migrate`.

Also updated the load-notebooks (now load-projects) and load-templates targets to use typescript 
scripts rather than the API. They no longer require copying of a token to work. 

Removed a few old scripts that are no longer relevant.

Replaced the old sample notebooks with newer versions more useful for testing.

Updates documentation to refer to the above rather than initdb.

## How to Test

With a database initialised under an older version (ie. before the permissions PR) run:

```bash 
npm run migrate
```

You should see a clean migration of your databases including people and invitest.

With a brand new couchdb database (e.g. remove the docker volume so that it is recreated
on restart).  Run:

```bash 
npm run migrate
```
Your database should be initialised properly and the admin user added.  Then you can run:

```bash
npm run load-templates
npm run load-projects
```

to get two sample notebooks and two sample templates.
## Checklist


PeterBaker0 (approved):
Lgtm - just add a note regarding targetting localhost vs couchdb in the .env file

PR #1408, opened 2025-04-02
feat: teams (backend)

PeterBaker0:
# feat: teams (backend)

## JIRA Ticket

https://jira.csiro.au/browse/BSS-855

## Description

Builds a team permission system. 

Teams are actually generic - in the sense that there is nothing special about them in the permission model. We could easily add organisations later, if desired, and I believe that the permission model would nest these permissions (e.g. Organisation grants roles in teams, which grants roles in resources). 

Teams are implemented as follows

- teams are a resource (CRUD permissions as you would expect)
- templates, projects are resources (this PR makes templates follow project pattern)
- introduces a concept called "virtualRoles" and "associations". Virtual roles are a map of resource types to roles on a Role. They say something like "If the user has this role on this resource, they should also have these roles on these other types of resources IF this resource is associated with other resources of this type". A concrete example is Teams -> Projects. Being a team manager makes you a project manager **for projects owned by the team**.
- there is one aspect of this logic that can't be centralised - determining the 'relevantAssociations' array which is a set of resource associations - this is heavily coupled to the API since it requires knowing 'who owns what' - this logic is contained in the API, but defers to centralised auth methods as soon as possible
- the API token signing had to be overhauled a bit - and the way the people DB stores roles was changed . There is now a slight separation between a Couch User and an Express User - it is important to understand this difference. A Couch User is the 'source' authorisation information - and is only relevant in the context of the API - it includes projectRoles, teamRoles, templateRoles and globalRoles - these are now intentionally separated  to improve the clarity of managing these roles. The Express User is the Couch User **plus** a set of resourceRoles. The resource roles **include the virtual roles** and such can be used in combination with global roles to provide an authoritative authorisation check (with the isAuthorized) method. The API also signs tokens, and in doing so it **takes an express User, and encodes the global, resource and couchdb roles according to the existing logic** - this means that all clients except the API have no idea about teams, which is fantastic in terms of separation of concerns - all clients can just look at the resourceRoles in the token, and trust that these are already taking into account any additional abstraction layers such as teams/orgs etc. If clients need explicit information about teams, they should contact the API and ask for it explicitly (which is in itself protected information based on the resource roles!). 
- teams have their own DB, with basic name + description and CRUD methods in the usual pattern
- projects and templates were updated to have an optional ownedByGroupId field which is immutable - API methods updated to accept this - also **added an index to lookup by team**
- everything is unit tested as much as possible

## Note regarding ownership

Right now I don't explicitly look at 'ownership' of resources in terms of permissions - instead I rely on the roles for that resource. In the future we may want to apply special protections/checks for ownership vs roles. For example, I now add the TEMPLATE_ADMIN role for the template to the creator, but if somehow that role was removed, then the person could no longer view their own template. 

## How to Test

Review the code primarily - since nothing is exposed in the front end - this is a backend only set of changes. 

We can review functionality more comprehensively once we can tinker with the teams in a nice UI.

## Checklist


library/data-model/src/permission/model.ts
```diff
@@ -706,9 +889,64 @@ export const roleActions: Record<
...
+      Action.ADD_MANAGER_TO_TEAM,
+      Action.REMOVE_MANAGER_FROM_TEAM,
+    ],
+    inheritedRoles: [Role.TEAM_MANAGER],
+    virtualRoles: new Map([
+      // Projects owned by team -> manager
+      [Resource.PROJECT, [Role.PROJECT_ADMIN]],
+      // Template owned by team -> admin
+      [Resource.TEMPLATE, [Role.TEMPLATE_ADMIN]],
+    ]),
```
stevecassidy, line 949: This module is quite long now, I wonder if we might want to refactor it to more manageable modules at some point?

library/data-model/tests/migrationService.test.ts
```diff
@@ -634,7 +648,7 @@ describe('Migration System Tests', () => {
       );
 
       const migrationDoc = migrationDocs.rows[0].doc as MigrationsDBDocument;
-      expect(migrationDoc.migrationLog.length).toBe(2); // Should still have just the original log entries
+      expect(migrationDoc.migrationLog.length).toBe(3); // Should still have just the original log entries
```
stevecassidy, line 651: Just noting that in the test below this one ('should handle migration failures and update migration log') I get error outputs which is expected because of the test but worrying when watching the tests.  Can we supress output during migrations while testing?

api/test/teamIntegration.test.ts
```diff
@@ -0,0 +1,1494 @@
...
+        .post(`${TEMPLATE_API_BASE}`)
+        .send({
+          ...getSampleNotebook(),
+          teamId: team._id,
+        }),
+      adminToken
+    ).expect(200);
+  });
+
+  it('correctly fetches templates by team ID', async () => {
```
stevecassidy, line 467: I get a test failure when running this as part of the overall suite. On it's own it passes. "InternalSystemError: An unexpected error occurred while trying to PUT the new template document into the teams DB. Exception {"status":409,"name":"conflict","message":"Document update conflict"}"

stevecassidy (commented):
Overall looks good, can't see any problems. Just a few comments to think about.

stevecassidy (approved):

library/data-model/src/permission/model.ts
PeterBaker0: Yeah - I would agree - could split into actions, roles, model ? Let's do later

library/data-model/tests/migrationService.test.ts
PeterBaker0: Done

PeterBaker0:
Woo merging!

PR #1413, opened 2025-04-03
Remove designer oldtextfield

ranisa-gupta16:
## JIRA Ticket

[BSS-827
](https://jira.csiro.au/browse/BSS-827)

## Description

-  Removed the old TextField field type from the field type selection list.
- Ensured onew text field FAIMSTextField are shown in the UI for new field creation.
- Cleaned up conditional logic in view.tsx that handled special rendering for TextField.
- Confirmed that TextField is no longer included in getFieldNames() and the component registry 

## How to Test
- Navigate to designer locally
   - cd designer
   - npm run start
   - http://localhost:5010
- Upload any notebook/survey JSON 
- Navigate to "DESIGN" tab
- Click on "add field button" and from the "Create a new field.. " popup, verify that TextField doesn't exist anymore in that field list.
- New FAIMSTextField is displayed at the top of list (as this is frequently used field in designer)

Screenshots for Reference:
BEFORE :
image

AFTER:
image

## Checklist

### DIFF
--- a/app/src/gui/component_registry/bundle_components.ts
+++ b/app/src/gui/component_registry/bundle_components.ts
@@ -76,7 +76,6 @@ registerField(
   Checkbox
 );
 
-
 registerField(
   'formik-material-ui',
   'TextField',
@@ -239,6 +238,15 @@ registerField(
   'Multiple line Input Box',
   'Text',
   FormikTextField
+);
+
+registerField(
+  'core-material-ui',
+  'TextField',
+  'HTML text field',
+  'A simple text field',
+  'Simple',
+  TextField
 );
 
 registerField(
--- a/app/src/gui/component_registry/bundle_components.ts
+++ b/app/src/gui/component_registry/bundle_components.ts
@@ -76,14 +76,6 @@ registerField(
   Checkbox
 );
 
-registerField(
-  'core-material-ui',
-  'TextField',
-  'HTML text field',
-  'A simple text field',
-  'Simple',
-  TextField
-);
 
 registerField(
   'formik-material-ui',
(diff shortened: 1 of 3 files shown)


app/src/gui/component_registry/bundle_components.ts
```diff
@@ -76,14 +76,6 @@ registerField(
   Checkbox
 );
 
-registerField(
```
stevecassidy, line 79: We can't remove this as many older notebooks use this field.  Ok to hide it in designer but it must still work.

designer/src/components/field-list.tsx
```diff
@@ -258,13 +258,13 @@ export const FieldList = ({viewSetId, viewId, moveFieldCallback}: Props) => {
               });
             }}
           >
-            {allFieldNames.map((fieldName: string) => {
-              return (
+            {allFieldNames
+              .filter(fieldName => fieldName !== 'TextField')
```
stevecassidy, line 262: Shouldn't be needed since you removed it in fields.tsx.

stevecassidy (changes requested):
Please ensure that we can still edit notebooks with old textfields in them even if we can't add new ones.  

Eventually it would be good to have a migration to replace old text fields with new ones where we can. Difficulty is that textfield is used to implement other field types like date field, number field.

designer/src/components/field-list.tsx
ranisa-gupta16: okay I'll update this thing.

app/src/gui/component_registry/bundle_components.ts
ranisa-gupta16: Ah yeah good point. thanks for that :-)

ranisa-gupta16:
@stevecassidy , I have updated the PR as per your mentioned comments. :-)

stevecassidy (approved):
Looks good now, thanks.

PR #1415, opened 2025-04-03
feat: Debounce Designer text fields

ryankontos:
# feat: Debounce Designer text fields

## JIRA Ticket

[BSS-700](https://jira.csiro.au/browse/BSS-700>)

## Description

This PR replaces usage of MUI's TextField component in the Designer with a new custom DebouncedTextField component. It essentially functions as a drop in replacement everywhere TextField was used. It has a default debounce of 300ms.

## How to Test

You can test this anywhere a text field appears in designer, all usages have been replaced. I have found a good example of field that interactively updates the UI is the "edit section name" field. You can check the debug statements in the Console to observe how the debouncing is functioning.

## Checklist


PeterBaker0 (changes requested):
Idea looks sound - tried it out and replaced a bunch more usages. I can see in the console that the FIRING event occurs repetitively every 300ms once you're inside a form/section. Not sure why - but currently this causing excessive dispatches to the store.

ryankontos:
@PeterBaker0 I'll look into it! I fixed an issue like that with another field but didn't catch this one. I may be wrong but I think it may have been occurring previously, just without the console logs to catch it.

app/package.json
```diff
@@ -155,7 +155,6 @@
     "@vitest/coverage-v8": "^0.34.4",
     "eslint-plugin-react": "^7.32.2",
     "fast-check": "2.25.0",
-    "gts": "^5.3.1",
```
PeterBaker0: was this intentional?

PeterBaker0 (approved):
Lgtm - works now - changed to 500ms. Removed gts changes - please put this in a separate PR explaining the change. I have no issues with linting/gts fixing in current configuration.

PR #1420, opened 2025-04-04
feat: team invites and improved permission UI filtering

PeterBaker0:
# feat: team invites and improved permission UI filtering

## Description

Overhauls the invites in the app to be

- multi resource (teams and projects supported) 
- much more rich with details, logs etc 
- standardised in the API (uses couchdb service layer + api on it's own router)

Adds more extensive unit tests.

Migrates the invites DB non destructively.

Brings back expiry + uses to address penetration testing concerns. Expiry required. Uses optional. 

Updates forms etc with new fields. 

Updates tables etc with new fields.

Users can only see what they can do i.e. view invites, add invites of different levels. 

Exposes new columns in invites, including deletion, expiry, remaining etc. 

Users can now register in the app by joining a team!

## How to Test

Thoroughly try out registering in the app through both teams and survey invites. 

## Checklist


api/src/api/invites.ts
```diff
@@ -0,0 +1,395 @@
...
+// This must occur before express api is used
+patch();
+
+export const api = express.Router();
+
+/**
+ * GET all project invites
+ */
+api.get(
+  '/project/:projectId',
```
stevecassidy, line 57: I think the rest of the API uses /notebook or /survey. I don't object to project but we should probably be consistent.

api/src/api/invites.ts
```diff
@@ -0,0 +1,395 @@
...
+  isInviteValid,
+} from '../couchdb/invites';
+import * as Exceptions from '../exceptions';
+import {requireAuthenticationAPI} from '../middleware';
+import patch from '../utils/patchExpressAsync';
+
+// This must occur before express api is used
+patch();
+
+export const api = express.Router();
```
stevecassidy, line 51: Hmm, I don't suppose the swagger.json documentation has been updated to align with the newer API routes.  We should update it or find another way to generate similar documentation.

library/data-model/src/data_storage/migrations/migrations.ts
```diff
@@ -154,14 +158,53 @@ export const invitesV1toV2Migration: MigrationFunc = doc => {
...
+    _id: inputDoc._id,
+    _rev: inputDoc._rev,
+    // Create a descriptive name
+    name: `${inputDoc.role} invite for ${inputDoc.projectId}`,
+    // Make them expire in one day from when this migration is applied
+    expiry: Date.now() + 24 * 60 * 60 * 1000,
+    // Project ID matches
+    resourceId: inputDoc.projectId,
+    // Invite for project
+    resourceType: Resource.PROJECT,
```
stevecassidy, line 184: If we happened to be migrating a V3 invite this would rewrite all invites as project invites, is it worth checking for inputDoc.resourceType to make this idempotent?

web/src/hooks/get-hooks.ts
```diff
@@ -166,25 +168,59 @@ export const useGetUsers = (user: User | null) =>
...
+export const useGetTeamInvites = (user: User | null, teamId: string) =>
+  useQuery({
+    queryKey: ['teaminvites', teamId],
+    queryFn: async () => {
+      const invites = (await get(
+        `/api/invites/team/${teamId}`,
+        user
+      )) as GetTeamInvitesResponse;
+      const promises = invites.map(async invite => {
+        const url = `${import.meta.env.VITE_API_URL}/register/${invite._id}`;
```
stevecassidy, line 214: Just noting it here for want of a better spot.  This URL will register someone for a team but what happens afterwards? For project invites we redirect back to the app but here I'm left in old conductor on a page that will soon not exist.  Need to add a redirect back to here.

stevecassidy (changes requested):
Generally looks good, all works well apart from being stranded after I accept a team invite.  Also thinking what I would do with the short-code for a team invite?  Should there be a short code entry on the Teams page in /web?

api/src/api/invites.ts
PeterBaker0: resolved

api/src/api/invites.ts
PeterBaker0: defer - there is an existing ticket for this

library/data-model/src/data_storage/migrations/migrations.ts
PeterBaker0: defer this - version sniffer or other approach needed - consistent with other migrations

web/src/hooks/get-hooks.ts
PeterBaker0: Done - redirects back to the teams page configurably

PR #1436, opened 2025-04-14
feat: improved template API and misc improvements

PeterBaker0:
# feat: improved template API and misc improvements

## JIRA Ticket

https://jira.csiro.au/browse/BSS-876

## Description

This PR originally intended to just fix the /web to properly reference the template name. However this lead to a chain reaction of changes which are all heading in the right direction for the project. 

- name was never actually used - it was relying on the metadata name reference which is in contrast to how the project logic worked - it was effectively ignored
- the references in /web were wrong to the name -hence missing menu items
- templates are managed in a combination of types.ts + api.ts in the data model - this was completely external to the migration system, so to run the migration to add the name field I needed to move these types into the couchdb models
- this was a good exercise, and also showed a really nice pattern we can use going forward (see below)
- this major refactor resulted in some issues in the data model - namely that we had some very sneaky circular imports. See below.

Everything appears to be working now.

## Note on couch db and zod

See the templates as a new pattern which has a great DX/maintainability to it. 

1. use zod schemas to define the fields in the types.ts for a couch db
2. expose a common suite of inferred interfaces 
3. use .pick(), omit() and extend() zod methods to create DTOs for API validation **without redefining any fields!**

See 

templates/types.ts
```typescript
import {z} from 'zod';
import {EncodedUISpecificationSchema} from '../../types';
import {CouchDocumentSchema, CouchExistingDocumentSchema} from '../utils';

// V1
export const TemplateV1Schema = z.object({
  // Version (internally incremented upon update)
  version: z.number().min(1),
  // NOTE: For some reason importing this from ./types causes an undefined error
  // (maybe circular import issue?). Metadata for the notebook (not optional -
  // refine to enforce this)
  metadata: z.record(z.any()).refine(val => !!val),
  // UI Spec of the template
  'ui-specification': EncodedUISpecificationSchema,
  // Which team owns this (optional)
  ownedByTeamId: z.string().optional(),
});
export type TemplateV1Fields = z.infer<typeof TemplateV1Schema>;
export type TemplateV1Document = PouchDB.Core.Document<TemplateV1Fields>;

// V2
export const TemplateV2Schema = TemplateV1Schema.extend({
  // Title of the template
  name: z.string().trim().min(1),
});
export type TemplateV2Fields = z.infer<typeof TemplateV2Schema>;
export type TemplateV2Document = PouchDB.Core.Document<TemplateV2Fields>;

// Current (V2)
// Fields
export const TemplateDBFieldsSchema = TemplateV2Schema; // v2
export type TemplateDBFields = z.infer<typeof TemplateDBFieldsSchema>;

// Document
export const TemplateDocumentSchema = CouchDocumentSchema.extend(
  TemplateDBFieldsSchema.shape
);
export type TemplateDocument = z.infer<typeof TemplateDocumentSchema>;

// Existing document
export const ExistingTemplateDocumentSchema =
  CouchExistingDocumentSchema.extend(TemplateDBFieldsSchema.shape);
export type ExistingTemplateDocument = z.infer<
  typeof ExistingTemplateDocumentSchema
>;

// Database
export type TemplateDB = PouchDB.Database<TemplateDBFields>;
```

And creation of DTOs from this 

```typescript
// POST create new template
export const PostCreateTemplateInputSchema = TemplateDBFieldsSchema.pick({
  metadata: true,
  'ui-specification': true,
  name: true,
}).extend({
  // prefer to use a nicer team ID input field
  teamId: z.string().trim().min(1).optional(),
});
export type PostCreateTemplateInput = z.infer<
  typeof PostCreateTemplateInputSchema
>;
export const PostCreateTemplateResponseSchema = ExistingTemplateDocumentSchema;
export type PostCreateTemplateResponse = z.infer<
  typeof PostCreateTemplateResponseSchema
>;
```
## Note on circular dependencies

the data model was having some very sneaky issues where values are undefined unpredictably at runtime. I think this is due to circular dependencies in the imports. I have added a lint check which uses the 'madge' tool to assert that there are no circular dependencies. 
 
I've gone through and removed all detected circular dependencies. This involved a) moving constants and types into standalone files which minimise any imports of other modules b) changing imports such as import {a, b, c} from '../../' which rely on the barrel from the root to instead target more specific modules.
 
In general we should be aiming to a) isolate types to types.ts files in the relevant module b) isolate constants to files where there are no imports - there should be no need to import into constant files - and if you do import, try to import from other constants files with no imports c) structure the files logically in a hierarchy such that for most cases you only import stuff from sibling modules, not the parent. 
 
To run this check manually - npm run circular-dep-checkThe check will run automatically in the CI/CD pipeline as part of the lint check!
 
The output can be a bit obtuse.  See below.
 
 ```
Plain Text
✖ Found 2 circular dependencies!
1) api.ts > data_storage/index.ts > data_storage/templatesDB/index.ts > data_storage/templatesDB/init.ts > data_storage/templatesDB/security.ts > index.ts
2) data_storage/index.ts > data_storage/templatesDB/index.ts > data_storage/templatesDB/init.ts > data_storage/templatesDB/security.ts > index.ts
```
 
in the top example, read this as "Module a imports module b, module b imports module c .... module z imports module a". You'll note the last item imports the first. 
 
So typically the easiest intervention is to go to the last module in the list, and try to see if you can refine/remove the import of the first module. 
 
Doing this process (which sucked) appears to have fixed the runtime issues with undefined constants. 

## How to Test

Please thoroughly test CRUD methods for projects and templates. 

I'll try and merge #1432 first so we can check this works in /web at latest version.

## Checklist


stevecassidy (approved):
All looks good.

PR #1445, opened 2025-04-15
fix: Moving/Deleting a summary field causes crash

ryankontos:
# fix: Moving/Deleting a summary field causes crash

## JIRA Ticket

[BSS-866](https://jira.csiro.au/browse/BSS-866)

## Description

A referential integrity issue was causing a crash in the `FormSettingsPanel`, specifically inside the Autocomplete for selecting summary fields. This happened when a field previously selected as a summary field was deleted or moved to a different form. The component tried to render a now-missing field, resulting in a crash.

The deletion/movement of a field did not remove its reference in `viewSet.summary_fields`.

## Proposed Changes

- Updated the affected Autocomplete to filter out `null`/`undefined` values, printing a warning if this occurs.
- Modified the `fieldMovedToSection` and `fieldDeleted` actions in the `uiSpecificationReducer` so that they perform the necessary cleanup of summary fields.
- Created file `designer/src/state/helpers/uiSpec-helpers.ts` with relevant helper functions. 
- Also moved the `slugify` helper function that existed in `uiSpec-reducer.tsx` to the helpers file.

## How to Test

1. Add a few fields to a form and set some as summary fields.
2. Delete one of those fields, or move it to a different form.
3. The summary field Autocomplete should **not cause a crash**.

You can verify that the state is correct, and not that the crash is simply being mitigated by the new filtering being done in the affected Autocomplete, as `Found null/undefined field while attempting to render summary fields!` will be printed by FormSettingsPanel if a summary field is found to be incorrectly missing.

## Checklist


PeterBaker0 (changes requested):
Really don't like this here - please make the reducer action which generated this invalid state not generate the invalid state - much cleaner approach! (Your instincts were right ;) )

ryankontos:
@PeterBaker0 I should've known! Fell into the _good enough_ trap for sure. I think what I have now is much nicer. Thanks for all the reviewing you've been doing by the way I appreciate it!

PeterBaker0 (approved):
Much nicer 👍 in the future we could have a warning if you try to remove/move a field used as a summary field - however this is _good enough_ for now. 

Thanks for all your contributions!

ryankontos:
@PeterBaker0 Same here, sorry!

PeterBaker0 (approved):
This makes sense - approved. Will merge.

PR #1449, opened 2025-04-16
chore: Remove default helper text, make certainty and annotation false by default

ryankontos:
# chore: Remove default helper text, make certainty and annotation false by default

## JIRA Ticket

[BSS-885](https://jira.csiro.au/browse/BSS-885)

## Description

Some minor improvements to Designer.

## Proposed Changes

- Removed default field helper text, from both the code and the sample notebook files.
- Annotation and certainty checkboxes are now **not checked** by default.

## How to Test

- Create a new notebook, form, then section.
- Add a field. It should have no helper text. Annotation and Certainty checkboxes should be _not selected_.

## Checklist


designer/notebooks/sample_notebook.json
PeterBaker0, line 1: can you revert changes to sample notebooks - they can have helper text in them

library/data-model/tests/mocks.ts
PeterBaker0, line 1: these are also fine

designer/src/test-notebook.ts
PeterBaker0, line 1: so are these

PeterBaker0 (approved):
Looks fine - but you don't need to remove instances of helper text from the sample notebooks + mocks. Just the default field instantiations + the ui spec reducer which adds the annotations. If you're confident with that update based on this comment + you have tested it, feel free to merge

ryankontos:
@PeterBaker0 Thanks for the approval, I am just unable to merge with main as I do not have authorisation!

PR #1450, opened 2025-04-16
feat: token exchange workflow

PeterBaker0:
# feat: token exchange 

## JIRA Ticket

Closes https://jira.csiro.au/browse/BSS-897

## Description

For two reasons, directly passing tokens in query string redirects is bad

- we hit a wall if tokens become too large due to many hosting providers URL maximum length limits (in our case with cloudfront we actually hit this limit on the BSS deployment due to inefficient encoding of global roles - see https://jira.csiro.au/browse/BSS-896 a related issue )
- security wise, this is not great, you are storing in your browser history/logs tokens which can remain valid (as in the case of a refresh token) for 48 hours - ideal not to expose these

## Implemented Changes

- upgraded auth DB models to follow the new zod model style which allows for runtime parsing/validation and is consistent with new style (e.g. templates, people DB)
- added exchangeTokenHash and exchangeTokenUsed fields 
- updated refresh token generation logic to also generate a short token which is then hashed into the exchangeTokenHash field
- there is an index looking up against this hash
- updated the redirectWithToken method used exclusively by the passport methods to return an exchangeToken and serverId (this helps the /app identify the source of the access redirect)
- added the /api/auth/exchange POST endpoint which performs a one-time handshake to exchange the token for a refreshToken + accessToken (this refresh token is not new, it is the refresh token for which this exchangeToken was generated previously)
- updates clients to use this rapid handshake (by synchronously performing the handshake as soon as they are exposed to the exchange token) 

## How to Test

Try logging in, logging out, changing users, adding users etc. to /web and /app. Also note that the unit tests have been updated with thorough testing for the exchange tokens. 

## Checklist


stevecassidy (approved):
Checks out ok.

PR #1452, opened 2025-04-16
feat: minimise redundant token encoding for global roles

PeterBaker0:
## JIRA Ticket

https://jira.csiro.au/browse/BSS-896

## Description

Tokens previously encoded all specifically granted resource roles -> resourceRoles + couchdb roles, then also the global roles -> couchdb roles. This changes this behaviour by **removing any resource roles/couchdb roles which are already implied by global roles**.

This defers the issue with global admins having massive tokens - very related to #1450 . 

However, we will hit header limits eventually - see https://stackoverflow.com/questions/686217/maximum-on-http-header-values

## Proposed Changes

Filters resource/couch roles to only include roles not implied by drilling global roles. 

## How to Test

Create some resources, ensure that both a) global admin users and b) users with specifically granted resource roles and c) users with teams implied roles, can do what you expect.

## Additional Information

We only **restrict** permissions here - so if there is a bug in this logic, it can only result in less access. 

## Checklist


stevecassidy (approved):
Code looks good, haven't been able to test.

PR #1453, opened 2025-04-16
fix: Notebook null conditions

ryankontos:
# fix: Notebook null conditions

## JIRA Ticket

[BSS-886](https://jira.csiro.au/browse/BSS-886)

## Description

This PR resolves an issue where conditions were being set to `null` upon deletion, as a side effect of a previous bug fix. This would cause a validation error when trying to import an affected notebook. It also introduces other relevant changes, detailed below.

## Proposed Changes

- Modified the `sectionConditionChanged` action in the UI Specification reducer such that conditions with a null value cause the condition key in the notebook to be removed.
- Created a new `fieldConditionChanged` action in the reducer that mirrors the existing `sectionConditionChanged` action, and also performs null removal. Previously, BaseFieldEditor was receiving the condition, adding it to the field, and dispatching `fieldUpdated` with the entire field. I see why things were done that way but I believe this change creates more high level consistency in how we handle conditions in Redux.
- Added migration functions `removeNullFieldConditions` and `removeNullFviewConditions` in `migrateNotebook`, to prevent imports of affected notebooks from failing.
- Migration is now run on the locally stored notebook upon load. This takes no more than 15-30ms on an averaged sized notebook in my testing.
- Removed debouncing of condition editor text fields.

## How to Test

- Import a notebook with null conditions (by manually editing JSON) for fields and/or sections and ensure the process goes smoothly.
- Delete a field condition and ensure that the notebook can still be imported after export.
- Delete a section condition and ensure that the notebook can still be imported after export.

## Additional Information
- I am not sure that we need the null condition migration code (Slippery slope if we decide to correct every notebook bug that ever existed during development) but I added it in case. Feel free to remove it now or in the future.
- Open to feedback on the migrate-on-load feature, it seems to work fine (other than the fact we lose undo history) but there's a chance I am missing something.

## Checklist


designer/src/components/condition.tsx
```diff
@@ -623,7 +623,9 @@ const BooleanConditionControl = (props: ConditionProps) => {
   };
 
   const deleteCondition = () => {
-    updateCondition(EMPTY_FIELD_CONDITION);
+    if (props.onChange) {
+      props.onChange(EMPTY_FIELD_CONDITION);
+    }
   };
```
PeterBaker0, line 629: Will this lose sync between the condition useState and the parent props.onChange ? (Not that it should have both!!!)

designer/src/state/migrateNotebook.ts
```diff
@@ -65,9 +68,23 @@ export class ValidationError extends Error {
...
+function removeNullFields(notebook: Notebook) {
+  if (!notebook['ui-specification']?.fields) return;
+
+  for (const fieldName of Object.keys(notebook['ui-specification'].fields)) {
+    if (notebook['ui-specification'].fields[fieldName] === null) {
+      delete notebook['ui-specification'].fields[fieldName];
+    }
+  }
+}
+
```
PeterBaker0, line 84: could you also migrate the issue causing this bug which was the condition property being null, not the field itself?

PeterBaker0 (changes requested):
I think the reasoning behind the change in the condition component requires more justification - why is changing from the updateCondition func which manages the react state synced to parent state with the props.onChange, to directly calling props.onChange better (the value is the same?). And I think the proposed migration does not actually remove the condition: null property in some of the fields (unless I'm missing something).

ryankontos:
@PeterBaker0 I've fixed the migration code, I committed an older version that was incorrect originally. Now we remove the condition prop if it is equal to null. That works for now.

There are two `deleteCondition` funcs in condition.tsx, I modified the wrong one. When I just modified the correct one, it ended up not fixing the issue like I thought it would, so going to need to spend some more time working out how conditions are getting set to null in the first place. Converting to draft for now until I fix it. I'll be more careful next time!

ryankontos:
@PeterBaker0 Properly fixed now. The whole PR is different so re-read when you have the chance. Thanks for your patience!

PR #1462, opened 2025-04-22
Fix Data Export from Control Centre

stevecassidy:
# Fix Data Export from Control Centre

## JIRA Ticket

[BSS-925](https://jira.csiro.com/browse/BSS-925)

## Description

The introduction of tokens has broken downloads in control center which relied on cookies to redirect the 
user to the download link.

## Proposed Changes

Introduces a two-step route for downloads of zip/csv data.  First request to /api/notebooks/:id/view.csv returns a redirect response to a URL containing a short lived (5 minute) JWT.   Requesting that URL gives the requested download.

Imlemented for zip and csv downloads here but can be extended to other download types.

## How to Test

Try exporting data from a notebook in the Control Center.

## Checklist

### DIFF
--- a/api/src/api/notebooks.ts
+++ b/api/src/api/notebooks.ts
@@ -43,7 +43,7 @@ import {
   Role,
   userHasProjectRole,
 } from '@faims3/data-model';
-import express, {Response} from 'express';
+import express, {response, Response} from 'express';
 import {z} from 'zod';
 import {processRequest} from 'zod-express-middleware';
 import {DEVELOPER_MODE, KEY_SERVICE} from '../buildconfig';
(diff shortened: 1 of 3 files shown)


api/src/api/notebooks.ts
```diff
@@ -381,64 +383,139 @@ api.get(
...
+  token: string;
+}): Promise<DownloadTokenPayload | null> => {
+  const signingKey = await KEY_SERVICE.getSigningKey();
+  try {
+    const result = await jwtVerify(token, signingKey.publicKey, {
+      algorithms: [signingKey.alg],
+      // verify issuer
+      issuer: signingKey.instanceName,
+    });
+    return result.payload as DownloadTokenPayload;
```
PeterBaker0, line 433: You have the zod model here - you could also parse it to make sure it's valid since this is a security sensitive action.

PeterBaker0 (approved):
I like this approach - like the use of zod here. neat approach with JWT being stateless.

PR #1463, opened 2025-04-22
Add Devcontainer Config for VSCode

stevecassidy:
# Add Devcontainer Config for VSCode

## JIRA Ticket

None

## Description

Adds configuration for devcontainer for vscode.

## Proposed Changes

This devcontainer config will create a pair of docker containers to allow working with the project in a
standardised environment.    It creates a couchdb container and a workspace container that can 
be used to run the other services.   A minimal set of environment settings suitable for development
are included in the config.  

This can be ignored if you prefer another environment.  Having this in the repo enables opening the repo
or a pull request in an isolated volume (https://code.visualstudio.com/docs/devcontainers/containers#_quick-start-open-a-git-repository-or-github-pr-in-an-isolated-container-volume).   

On MacOS this helps because it puts the code inside the container in a volume which is much faster than bind mounting
a local directory. 

## How to Test

In VSCode try opening this PR in a container as per the above link. 

After opening the container, you will need to run `npm migrate` in the api folder and then run each of the services.   Some configuration in .env files may be needed if you don't want the default settings.

## Checklist

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,69 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/typescript-node
+{
+	"name": "Node.js & TypeScript",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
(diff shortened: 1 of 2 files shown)


PeterBaker0 (approved):
Haven't tested it out but it looks sensible and non-destructive.

PR #1466, opened 2025-04-23
feat: change password

PeterBaker0:
Change password feature. 

Adds auth route and auth page. 

Adds hooks to this page from /web and /app. 

Screeshots below:

image

image

image


jyucsiro (approved):
Tested user workflow in Control Center and in FAIMS app. I was able to successfully change my password. I tried mismatching passwords and it failed as expected.

PR #1468, opened 2025-04-23
feat: forgot password

PeterBaker0:
# feat: forgot password

## Description

Implements a 'forgot password?' button and associated workflow. 

Involved: 

- implementing additional auth views including the forgot password email entry + the landing page for the link in the email sent
- associated POST auth routes in authRoutes for the above 
- integration into email service to send suitable password reset email 
- upgrading/migrating the email code model to include a created timestamp so that we can apply similar rate limiting principles as the verification emails (following the principle that no email enabled routes should be unmetered/limited)
- upgrading login page to show this forgot password (few other minor CSS tweaks while there to make fields more prominent) 
- removing old reset page from new-conductor - this is no longer needed as auth pages are centralised so as to be reusable to all clients

## How to Test

From fresh start

- login to control center as admin user
- make a team
- create invite to team
- open invite in new window/incognito and register as an email you can see emails to
- after registering, logout, then try to log back in
- test normal login works
- logout
- pretend you've forgotten your password
- click forgot password, submit an incorrect email address (which is syntactically valid) - note that this appears as success (for security reasons so you can't brute force emails)
- then put in your correct email - verify email arrives (don't click it yet)
- try generating it many times, and note the timeout - this has a 2 hour cooldown to prevent spamming email
- now check your emails, use the reset code in the email to update your password - on this form try putting in a) too short password b) not matching passwords. Then put in a correct one. Now try login with the old password, confirm it fails, then login with new one, confirm it succeeds. 

## Checklist


ranisa-gupta16 (approved):
I've reviewed this end-to-end with Peter over a call, and everything is working smoothly. Great work on implementing these features—email verification and the forgot password flow have been tested across multiple scenarios and perform flawlessly. Thanks for all the effort and attention to detail Peter. :-)

PR #1469, opened 2025-04-23
fix: qr scan of code with existing redirect causes double redirect

PeterBaker0:
# fix: qr scan of code with existing redirect causes double redirect

## JIRA Ticket

https://jira.csiro.au/browse/BSS-919

## Description

Replaces existing OR appends instead of blindly appending. This will always force a prioritisation of the app's preferred redirect. 

There could also be scope to prevent putting the redirect in in the first place (when generated from the control center) - relying on the API to have reasonable defaults - but this is harder as it would require inspecting the invite more thoroughly and determining a URL intelligently from that. 

This should fix this bug at least. 

## How to Test

Generate a team invite code in control center. Scan it on a mobile device. Ensure it opens, registers, then redirects back. 

@stevecassidy  hoping you can test on mobile.

## Checklist


stevecassidy (approved):
Not tested but I can't see any issue

PR #1474, opened 2025-04-24
feat: logout token invalidation

PeterBaker0:
Invalidates token when logging out. Adds endpoint /auth/logout expecting refreshToken as the payload. This supports a multi session model rather than invalidating all tokens for the current user. Still requires current auth - if the user cannot prove current auth they shouldn't be able to invalidate tokens. Returns 200 even in fail cases to avoid exposing internal information. Only 500s where a true failure occurs.


api/src/auth/authRoutes.ts
```diff
@@ -465,6 +471,51 @@ export function addAuthRoutes(app: Router, socialProviders: AuthProvider[]) {
   });
 
   /**
+   * Logout - logging out is an optional client 'good will' process in which the
+   * client indicates they wish for the specific refresh token to be
+   * invalidated. This is just a normal POST operation.
```
stevecassidy, line 476: Did you mean POST or PUT? It might be PUT if you were updating the state of an auth URI but you're not really doing that, POST is probably appropriate.

stevecassidy (approved):
Just one comment on the choice of verb.

PR #1475, opened 2025-04-24
fix: web cold load auth

PeterBaker0:
If you load /web on a cold load with an expired refresh token, the app does not recover from this state until it is forcefully refreshed. This was due to the following sequence

1. _protected route before load sees isAuthenticated true (since it was the last state of the app before close)
2. this causes it to return the child route - the child route throws a bunch of errors as isAuthenticated is incorrect
3. meanwhile, the auth context provider asynchronously refreshes - this refresh fails, and isAuthenticated is set to false
4. **_protected before load does not rerun!** - so this state is never picked up

The solution here was to add a splashscreen over the actual component code when isAuthenticated, as part of the react lifecycle. This will be caught with a re-render appropriately. It waits a couple of seconds as a safety against any settling actions, then redirects to login. You can recreate it most easily by deploying an API with a very short refresh token (api/.env), and setting the refresh token interval to every 10 seconds in the constants file in /web.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18834,16 +18834,6 @@
         "kuler": "^2.0.0"
       }
     },
-    "node_modules/appium-uiautomator2-driver/node_modules/@emnapi/runtime": {
-      "version": "1.4.3",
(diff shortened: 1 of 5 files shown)


stevecassidy (approved):
Looks good.

PR #1477, opened 2025-04-27
handle display of inherited data section, covered different scenarios…

ranisa-gupta16:
## JIRA Ticket

[BSS-535
](https://jira.csiro.au/browse/BSS-535)

## Description

Updated the logic for showing the "Inherited Data" accordion during child record creation.
Improved the conditions to avoid showing the inherited section unnecessarily
Ensured inherited data only appears when correctly configured in the form.

Inherited Data" accordion will now only display if:
A parent record is linked AND
The parent has fields with actual non-empty data AND
Those fields are explicitly marked with Display in Parent Record (displayParent: true).

## How to Test
i herited 1
inherited 2

When Parent has fields with Display Parent checked
Go to Form Designer and check "Display in Parent Record" (displayParent) for one or more fields.
Create a parent record and fill in some data.
Use RelatedRecordSelector → Add New Response to create a child in UI 
As an outcom, the Inherited Data accordion appears showing the fields with values.

2. When Parent has fields but Display Parent NOT checked 
Go to Form Designer and uncheck "Display in Parent Record" for all fields.
Fill in a parent record with data anyway.
Create a child record using Add New Response. as an outcomme, the inherited data accordion does NOT appear, even if parent has data.

3. When Parent Record has no data at all
Create a parent record but leave it empty (no fields filled).
Create a child record, as a result, the inherited Data accordion does NOT appear.
let me know for any queries.

## Checklist

- [y] I have confirmed all commits have been signed.
- [y] I have added JSDoc style comments to any new functions or classes.
- [y] Relevant documentation such as READMEs, guides, and class comments are updated.


PeterBaker0 (changes requested):
Approve of the approach - but would it be possible to refactor these two changes which are basically the same logic but very slightly differently implemented into a form utils helper function such as checkIfParentHasInheritedData() or similar?

ranisa-gupta16:
> Approve of the approach - but would it be possible to refactor these two changes which are basically the same logic but very slightly differently implemented into a form utils helper function such as checkIfParentHasInheritedData() or similar?

Updated the PR with the changes. Let me know if its good to go. :-) @PeterBaker0

ranisa-gupta16:
formutilities.ts updatd with checkIfParentHasInheritedData :-) @PeterBaker0

PeterBaker0 (approved):
Was missing typing and missing a .data call in persistentData - also the call inside the component was redundant as was already conditionally rendered. Tested now and working well under the two conditions.

app/src/gui/pages/record-create.tsx
```diff
@@ -165,6 +166,8 @@ function DraftRecordEdit(props: DraftRecordEditProps) {
   const [is_link_ready, setIs_link_ready] = useState(false);
   const [progress, setProgress] = useState(0);
 
+  console.log(parentLinks);
```
stevecassidy, line 169: remove debugging output

app/src/gui/pages/record-create.tsx
```diff
@@ -230,13 +233,16 @@ function DraftRecordEdit(props: DraftRecordEditProps) {
             variant={is_mobile ? undefined : 'outlined'}
           >
             {is_link_ready ? (
-              <InheritedDataComponent
-                parentRecords={parentLinks}
-                ui_specification={uiSpec}
-              />
+              checkIfParentHasInheritedData({parentLinks, uiSpec}) ? (
```
stevecassidy, line 236: I'd suggest moving this test inside InheritedDataComponent rather than requiring this construct here.

stevecassidy (approved):
Ok but some minor comments on code structure.

PR #1484, opened 2025-04-29
fix: project names change after edit

PeterBaker0:
# fix: project names change after edit

## JIRA Ticket

https://jira.csiro.au/browse/BSS-921

## Description

Improving handling of names for projects by exposing and respecting project.name instead of project.metadata.name.

Updates the API to return the project.name which is already being handled correctly. 

Updates the clients (web and API) to use this name rather than metadata.name. Also updates the redux store to expose Name directly, and merge it from API.

## How to Test

Make a notebook, check it works. Update it using web (can be the same file, that's fine), check that the name stays as the explicitly set name not the notebook metadata name. 

## Additional Information

We should improve this in the long term by removing the metadata DB and just having one place where this is configured. The specification name and the project name do not need to be/nor should be related.

## Checklist

### DIFF
--- a/api/src/api/notebooks.ts
+++ b/api/src/api/notebooks.ts
@@ -292,7 +292,7 @@ api.put(
     const metadata = req.body.metadata;
     const projectID = req.params.id;
     await updateNotebook(projectID, uiSpec, metadata);
-    res.json({notebook: projectID}).end();
+    return res.json({notebook: projectID});
   }
 );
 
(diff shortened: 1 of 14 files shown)


stevecassidy (approved):
Looks ok but noticed an error in api/src/api/notebooks.ts in the put('/:id' handler it is calling .end() after res.json() which gives an error about modifying headers after they have been sent.  .end() is not needed.

PR #1485, opened 2025-04-29
fix: remove stepper edge case for existing records by forcing all sections to be viewed

PeterBaker0:
# fix: remove stepper edge case for existing records by forcing all sections to be viewed

## JIRA Ticket

https://jira.csiro.au/browse/BSS-963

## Description

Adds a flag 'isExistingRecord' as prop to form.tsx. On mount, if it is an existing record, tells the stepper that all sections have been visited and that the form should try to validate. Also removes all special conditional formatting in the stepper which prevented showing proper colors when it was an existing record. Also made the formatting more obvious for which section is active.

## How to Test

Publish a form with multiple sections and errors. Go back to the record, you should see all sections have appropriate errors visible. 

## Checklist

### DIFF
--- a/app/src/gui/components/record/RecordData.tsx
+++ b/app/src/gui/components/record/RecordData.tsx
@@ -38,6 +38,7 @@ import {SHOW_RECORD_LINKS} from '../../../buildconfig';
 interface RecordDataTypes {
   project_id: ProjectID;
   serverId: string;
+  isExistingRecord: boolean;
   record_id: RecordID;
   hrid?: string;
   record_type: string;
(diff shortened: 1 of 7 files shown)


stevecassidy (commented):
Able to reliably get to the end with no buttons.

(In Rapid Survey) open a new Response, don't enter anything on the first screen, skip to the second, third, etc to the end - no buttons.  In the stepper, the first section is not highlighted as visited (still grey).

PeterBaker0:
> Able to reliably get to the end with no buttons.
> 
> (In Rapid Survey) open a new Response, don't enter anything on the first screen, skip to the second, third, etc to the end - no buttons. In the stepper, the first section is not highlighted as visited (still grey).

This is not attempting to resolve that issue - instead its about seeing errors across all forms for existing responses. Please create a record which has errors, publish it, then open it from the record list, and note that all sections show errors.

stevecassidy (approved):
Ok, approved for what it does.

PR #1486, opened 2025-04-29
Fix maps

ranisa-gupta16:
## JIRA Ticket

[BSS-961
](https://jira.csiro.au/browse/BSS-961)

## Description
Fixed:
Console error related to getZIndex() in MapWrapper.tsx.
Earlier, when reopening the map, a console warning appeared because layer.getZIndex() was being accessed unsafely. This has now been properly handled inside a try-catch, and the error no longer appears.

**Note on Geolocation Warnings:**
⚠️ Geolocation timeout or expired warnings (GeolocationPositionError: Timeout expired) may still appear when testing in the browser.
This is expected behavior because browser-based GPS is inherently less reliable and depends on network and device permissions.
In real mobile apps (with device hardware GPS), live location will work smoothly and automatically without timeout errors.

## How to Test

Open the map in the app as usual
: No console error related to getZIndex() appears when opening or interacting with the map.

Please be mindful -> If testing in browser:
Some geolocation timeout warnings may still be logged (expected and harmless).
The blue live location cursor might not always appear immediately — this is normal in browsers.

## Checklist

### DIFF
--- a/app/src/gui/components/map/map-component.tsx
+++ b/app/src/gui/components/map/map-component.tsx
@@ -19,7 +19,6 @@
  * This component renders an interactive OpenLayers map with support for:
  * - Real-time GPS tracking (blue dot, accuracy circle, direction triangle)
  * - Offline map support using cached tiles
- * - Optional fake GPS simulation for browser testing
  * - Dynamic center control and zooming
  * - Integration with parent component through callback
  */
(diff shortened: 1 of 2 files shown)


stevecassidy (approved):
Looks good to me.

PeterBaker0 (approved):
Acknowledge that it fixes the zindex problem - but don't agree with using **Zindex** as an identifier for a layer! As discussed this doesn't resolve the repeated browser slow GPS issue - but agree this seems to be an issue with browser APIs so non-trivial to fix. Thanks.

PR #1488, opened 2025-04-29
feat: upgrade date fields

PeterBaker0:
Refactors faims text field to expose base input class, then passes in types for text, date time, date, month. Updates designer to follow suit of text field. Updates bundling to expose these components. 

No migration needed for existing notebooks as previous implementation will continue to work (as it uses text field with input props). notebooks wanting to use new version will need to update the field type to match the FAIMSTextField but use the new names i.e. DateTimePicker, DatePicker, MonthPicker

Example of a valid date field

```
            "Date-of-bushfire": {
                "component-namespace": "faims-custom",
                "component-name": "DatePicker",
                "type-returned": "faims-core::Date",
                "component-parameters": {
                    "label": "Date of bushfire",
                    "fullWidth": true,
                    "helperText": "TEXT",
                    "variant": "outlined",
                    "required": true,
                    "name": "Date-of-bushfire",
                    "advancedHelperText": "",
                    "protection": "none"
                },
                "validationSchema": [
                    [
                        "yup.string"
                    ]
                ],
                "initialValue": "",
                "meta": {
                    "annotation": {
                        "include": false,
                        "label": "annotation"
                    },
                    "uncertainty": {
                        "include": false,
                        "label": "uncertainty"
                    }
                },
                "persistent": false,
                "displayParent": false
            }
```

image


ranisa-gupta16 (approved):
Looks great to me,  I have tested all the three pickers, DateTimePicker, DatePicker, MonthPicker, they are now consistent and is nicely displayed in UI. 
image

designer/src/fields.tsx
```diff
@@ -30,6 +30,70 @@ const fields: {[key: string]: FieldType} = {
...
+  DatePicker: {
+    'component-namespace': 'faims-custom',
+    'component-name': 'DatePicker',
+    'type-returned': 'faims-core::Date',
+    'component-parameters': {
+      label: 'Date time picker',
+      fullWidth: true,
+      helperText: '',
+      advancedHelperText: '',
+      variant: 'outlined',
```
stevecassidy, line 60: I don't think this is needed and it would be good to get rid of all display config from the uiSpec.

stevecassidy (approved):
Just remove the variant property.

designer/src/fields.tsx
PeterBaker0: Follow up task as discussed.

PR #1492, opened 2025-04-30
fix: cannot add child records when there are form errors

PeterBaker0:
This is due to formik refusing to run onSubmit when there are form errors. The RelatedRecordSelector relied on this behaviour (hence probably why old FAIMS folk instead chose to **sneakily** disable adding a child record until all errors are gone - good one!!). Instead we pass through 'forceSave' which bypasses this check. Very gross approach but it's the smallest intervention to fix the problem.

### DIFF
--- a/app/src/gui/components/record/relationships/create_links/create_record_link.tsx
+++ b/app/src/gui/components/record/relationships/create_links/create_record_link.tsx
@@ -29,7 +29,7 @@ export function AddNewRecordButton(props: {
   pathname: string;
   state: LocationState;
   text: string;
-  handleSubmit: Function;
+  handleSubmit: () => Promise<RevisionID>;
   project_id: string;
   serverId: string;
   save_new_record: Function;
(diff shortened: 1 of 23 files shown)


app/src/gui/components/record/relationships/create_links/create_record_link.tsx
```diff
@@ -47,12 +47,12 @@ export function AddNewRecordButton(props: {
         .handleSubmit()
         .then((revisionID: RevisionID) => {
           const newState = props.state;
```
stevecassidy, line 49: So if handleSubmit fails we'll still spin, maybe we could pop up an error alert in this case just so the UI still responds.  It would be the un-useful 'Error saving record' alert that we use somewhere in the save method.

stevecassidy (approved):
Apart from that comment about showing an alert. Works well and well done for tidying as you go!

PR #1495, opened 2025-04-30
Replace map pin icon URL with local file

stevecassidy:
# Replace map pin icon URL with local file

## JIRA Ticket

None

## Description

The map pin icon was referenced as a URL so didn't work when offline.  Replace with a local file in the app.  Also use a suitably licensed icon rather than the previous one that required attribution.

## Checklist


PeterBaker0 (approved):
Good catch

PR #1513, opened 2025-05-30
 Fix Autoincrementer settings buttons

stevecassidy:
# Fix Autoincrementer settings buttons

## JIRA Ticket

None

## Description

When there is more than one auto incrementer field in a notebook, the settings button opened all
settings dialogs at once so only the last auto incrementer could be changed.

This PR fixes this by factoring out the button open/close status to a separate component.

## Checklist


ranisa-gupta16 (approved):
Looks good to me, thanks for this fix :-)

PR #1522, opened 2025-06-11
Feat/sync-audit

stevecassidy:
# Provide feedback to users that records have been synced

## JIRA Ticket

[BSS-998](https://jira.csiro.com/browse/BSS-998)

## Description

Users would like to know that their records are safe on the server. We need to provide feedback at the record level on which records are saved.   This may also facilitate us taking remedial action if we find that records are not being saved.

## Proposed Changes

Implement a record audit in the API that can verify that records on the client have been synced to the server. 

Provide an indication in the app that an individual record is synced.

## How to Test

TBD

## Additional Information

## Checklist

### DIFF
--- a/library/data-model/src/data_storage/internals.ts
+++ b/library/data-model/src/data_storage/internals.ts
@@ -49,6 +49,7 @@ import {
   RevisionID,
   UnhydratedRecord,
 } from '../types';
+import {createHash} from './utils';
 
 // INDEX NAMES
 
@@ -801,6 +802,73 @@ export async function getRecord({
     id: recordId,
     typeField: RECORD_TYPE_FIELD,
   });
+}
+
+/**
+ * Creates a canonical hash from the record audit data for efficient sync checking.
+ *
+ * @param audit - Array of {id, rev} objects from getRecordAudit
+ * @returns SHA-256 hash of the canonical representation
+ */
+async function createAuditHash(
+  audit: Array<{id: string; rev: string}>
+): Promise<string> {
+  // The audit is already sorted by ID in getRecordAudit, but ensure consistency
+  const sortedAudit = audit
+    .slice() // Don't mutate original
+    .sort((a, b) => a.id.localeCompare(b.id));
+
+  // Create canonical JSON representation
+  const canonicalJson = JSON.stringify(sortedAudit);
+
+  // Return SHA-256 hash
+  return await createHash(canonicalJson);
+}
+
+/**
+ * Generate an audit for a record.
+ * This returns a list of all document ids and revision ids for documents
+ * making up this record.  All frevs, avps and attachment documents.
+ *
+ * We can then use this in the client to verify that we have a complete and
+ * up to date copy of the record on the server.
+ *
+ */
+export async function getRecordAudit({
+  recordId,
+  dataDb,
+}: {
+  recordId: RecordID;
+  dataDb: DataDbType;
+}) {
+  // need to get the record itself and then query for all
+  // documents that reference this record ID
+  const record = await getCouchDocument<EncodedRecord>({
+    db: dataDb,
+    id: recordId,
+    typeField: RECORD_TYPE_FIELD,
+  });
+  if (record) {
+    const parts = await dataDb.find({
+      selector: {
+        record_id: recordId,
+      },
+      fields: ['_id', '_rev'],
+      limit: 1000,
+    });
+    const audit = [
+      {
+        id: recordId,
+        rev: record._rev,
+      },
+      ...parts.docs.map(doc => ({
+        id: doc._id,
+        rev: doc._rev,
+      })),
+    ];
+    return createAuditHash(audit);
+  }
+  return [];
 }
 
 /**
--- a/library/data-model/src/data_storage/internals.ts
+++ b/library/data-model/src/data_storage/internals.ts
@@ -867,8 +867,29 @@ export async function getRecordAudit({
       })),
     ];
     return createAuditHash(audit);
+  } else {
+    throw new Error(`Record ${recordId} not found`);
   }
-  return [];
+}
+
+export type RecordAuditMap = {
+  [recordId: string]: string;
+};
+
+export async function getRecordListAudit({
+  recordIds,
+  dataDb,
+}: {
+  recordIds: RecordID[];
+  dataDb: DataDbType;
+}) {
+  const result: RecordAuditMap = {};
+  for (const recordId of recordIds) {
+    result[recordId] = await getRecordAudit({
+      recordId,
+      dataDb,
+    });
+  }
 }
 
 /**
--- a/library/data-model/src/data_storage/utils.ts
+++ b/library/data-model/src/data_storage/utils.ts
@@ -182,3 +182,46 @@ export const couchInitialiser = async ({
     });
   }
 };
+
+/**
+ * Cross-platform hash function that works in both Node.js and browsers
+ */
+export async function createHash(data: string): Promise<string> {
+  // Browser environment
+  if (typeof window !== 'undefined' && window.crypto && window.crypto.subtle) {
+    const encoder = new TextEncoder();
+    const dataBuffer = encoder.encode(data);
+    const hashBuffer = await window.crypto.subtle.digest('SHA-256', dataBuffer);
+    const hashArray = Array.from(new Uint8Array(hashBuffer));
+    return hashArray.map(b => b.toString(16).padStart(2, '0')).join('');
+  }
+  // Node.js environment
+  if (typeof require !== 'undefined') {
+    try {
+      const crypto = require('crypto');
+      return crypto.createHash('sha256').update(data, 'utf8').digest('hex');
+    } catch (error) {
+      // Fallback if crypto is not available
+      console.warn('Node.js crypto not available, falling back to simple hash');
+    }
+  }
+  // Fallback: Simple hash function (not cryptographically secure)
+  return simpleHash(data);
+}
+
+/**
+ * Simple hash function fallback (not cryptographically secure, but deterministic)
+ */
+function simpleHash(str: string): string {
+  let hash = 0;
+  if (str.length === 0) return hash.toString(16);
+
+  for (let i = 0; i < str.length; i++) {
+    const char = str.charCodeAt(i);
+    hash = (hash << 5) - hash + char;
+    hash = hash & hash; // Convert to 32-bit integer
+  }
+
+  // Convert to positive hex string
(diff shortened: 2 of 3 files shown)


app/src/utils/apiHooks/notebooks.tsx
```diff
@@ -52,3 +61,60 @@ export const useCreateNotebookFromTemplate = (
...
+  listingId: string;
+  username: string;
+}) => {
+  const N = 2; // refetch time in minutes - maybe configure in env?
+  const queryKey = ['record-audit', projectId, listingId, username];
+  return useQuery({
+    queryKey,
+    queryFn: async (context: QueryFunctionContext) => {
+      const currentStatus: RecordStatus | undefined =
+        context.client.getQueryData(queryKey);
```
PeterBaker0, line 84: Might be unnecessarily exposing tanstack details here - maybe use a useState variable instead? Not the end of the world though. Also I think you might have a different queryFn available which knows of the previous data if any.

app/src/gui/components/notebook/record_table.tsx
```diff
@@ -175,6 +180,9 @@ function getDataForColumn({
       case 'KIND':
         return uiSpecification.viewsets[record.type]?.label ?? record.type;
 
+      case 'SYNC_STATUS':
+        return record.synced ? '✅' : '❌';
```
PeterBaker0, line 184: Should there be a status which is 'unchecked' - seeing a bright red X on all new records until the 2 minute timer fires is a questionable user experience. Red means error to many users. What about you show 'Syncing...' or a gray line or something until the check happens?

app/src/utils/apiOperations/notebooks.tsx
```diff
@@ -29,3 +36,102 @@ export const createNotebookFromTemplate = async (input: {
...
+  listingId,
+  currentStatus,
+}: {
+  projectId: ProjectID;
+  username: string;
+  listingId: string;
+  currentStatus: RecordStatus | undefined;
+}): Promise<RecordStatus> => {
+  // get the list of record ids from the project
+  const dataDb = await getDataDB(projectId);
```
PeterBaker0, line 61: getDataDB I am wanting to deprecate - it's the annoying registered client thing. Would prefer 'localGetDataDB' (don't love the name) - that uses the redux store (which is what is registered in this data model anyway).

app/src/utils/apiOperations/notebooks.tsx
```diff
@@ -29,3 +36,102 @@ export const createNotebookFromTemplate = async (input: {
...
+  // get a list of record ids from the project
+  const records = await queryCouch({
+    db: dataDb,
+    index: RECORDS_INDEX,
+  });
+  const recordIds = records.map(r => r._id);
+  const audit = await getRecordListAudit({recordIds, dataDb});
+  let filteredAudit: Record<string, string> = {};
+
+  // now filter any records that we know are good from the last
```
PeterBaker0, line 77: why not filter before calling 'getRecordListAudit' ?

PeterBaker0: oh you want to be able to pull changes I guess

PeterBaker0: ah I think just misunderstanding - all good

app/src/utils/apiOperations/notebooks.tsx
```diff
@@ -29,3 +36,102 @@ export const createNotebookFromTemplate = async (input: {
...
+
+  // if we're online, do the request
+  if (isOnline) {
+    if (Object.getOwnPropertyNames(filteredAudit).length > 0) {
+      const response = await FetchManager.post<PostRecordStatusResponse>(
+        listingId,
+        username,
+        `/api/notebooks/${projectId}/sync-status/`,
+        {record_map: filteredAudit} as PostRecordStatusInput
+      );
```
PeterBaker0, line 101: can you put this into queries.ts or whatever it's called? Don't expose the API details here - that can be it's own method

app/src/utils/apiOperations/notebooks.tsx
```diff
@@ -29,3 +36,102 @@ export const createNotebookFromTemplate = async (input: {
...
+      console.log('response', response);
```
PeterBaker0, line 102: console.debug?

app/src/utils/apiOperations/notebooks.tsx
```diff
@@ -29,3 +36,102 @@ export const createNotebookFromTemplate = async (input: {
...
+      // we need to merge the returned value with the
+      // current status
+      const status = {
+        ...response.status,
+        ...currentStatus?.status,
+      };
+      console.log('status', status);
```
PeterBaker0, line 109: debug?

app/src/utils/apiOperations/notebooks.tsx
```diff
@@ -29,3 +36,102 @@ export const createNotebookFromTemplate = async (input: {
...
+        recordHashes: audit,
+      };
+    } else if (currentStatus) {
+      return currentStatus;
+    } else {
+      return emptyStatus;
+    }
+  } else {
+    // not online
+    console.log('not online so...');
```
PeterBaker0, line 121: debug?

library/data-model/src/data_storage/internals.ts
```diff
@@ -803,6 +804,126 @@ export async function getRecord({
...
+  // Early return for empty input
+  if (!recordIds || recordIds.length === 0) {
+    return {};
+  }
+
+  // Batch fetch all records at once
+  const recordsResult = await dataDb.find({
+    selector: {
+      _id: {
+        $in: recordIds,
```
PeterBaker0, line 860: does this get problematic with too many records? I.e. this will make the query size potentially large for thousands of records? You may need to batch this over a certain amount?

library/data-model/src/data_storage/internals.ts
```diff
@@ -803,6 +804,126 @@ export async function getRecord({
...
+      },
+    },
+    fields: ['_id', '_rev'],
+    limit: recordIds.length + 10,
```
PeterBaker0, line 864: I'd definitely deal with batching this rather than force the limit

library/data-model/src/data_storage/internals.ts
```diff
@@ -803,6 +804,126 @@ export async function getRecord({
...
+    selector: {
+      record_id: {
+        $in: recordIds,
+      },
+      attach_format_version: {
+        $exists: false,
+      },
+    },
+    fields: ['_id', '_rev', 'record_id'],
+    limit: recordIds.length * 100, // Adjust based on expected documents per record
```
PeterBaker0, line 878: !! - dangerous - can we deal with paging/batching this instead?

library/data-model/src/data_storage/internals.ts
```diff
@@ -803,6 +804,126 @@ export async function getRecord({
...
+    string,
+    Array<{id: string; rev: string}>
+  >();
+
+  // Type assertion for the partial documents
+  type PartialDoc = {
+    _id: string;
+    _rev: string;
+    record_id: string;
+  };
```
PeterBaker0, line 892: no need to put this here - maybe outside function?

library/data-model/src/data_storage/internals.ts
```diff
@@ -803,6 +804,126 @@ export async function getRecord({
...
+  const relatedDocsResult = await dataDb.find({
+    selector: {
+      record_id: {
+        $in: recordIds,
+      },
+      attach_format_version: {
+        $exists: false,
+      },
+    },
+    fields: ['_id', '_rev', 'record_id'],
```
PeterBaker0, line 877: can you highlight this with a comment - important that we only look at id, rev, record_id as the fields being hashed

library/data-model/src/data_storage/internals.ts
```diff
@@ -803,6 +804,126 @@ export async function getRecord({
...
+  }
+
+  // Batch fetch all records at once
+  const recordsResult = await dataDb.find({
+    selector: {
+      _id: {
+        $in: recordIds,
+      },
+    },
+    fields: ['_id', '_rev'],
```
PeterBaker0, line 863: Please highlight this with comment - id and rev only

library/data-model/src/data_storage/internals.ts
```diff
@@ -803,6 +804,126 @@ export async function getRecord({
...
+ *
+ * We can then use this in the client to verify that we have a complete and
+ * up to date copy of the record on the server.
+ *
+ */
+export type RecordAuditMap = {
+  [recordId: string]: string;
+};
+
+export async function getRecordListAudit({
```
PeterBaker0, line 844: I think we should consider caching strategies for this - if we have many users hitting the DB this could result in many scans of full DB. |

Could we embed a hash function into validate_doc_update?

PeterBaker0 (changes requested):
Code looks reasonable with some minor comments. 

I wonder about this approach's scalability/efficiency - could we embed a hash function into the validate_doc_update so that the hash is generated once and only upon update. Then read that out rather than recomputing it a lot? Also think the DB query code needs to not use magic values for the limits (Given this is critical path code now) and instead deal with pagination.

library/data-model/src/data_storage/internals.ts
stevecassidy: As discussed, I don't see a good way to cache at the moment.

app/src/utils/apiHooks/notebooks.tsx
stevecassidy: This seems to be the way to use the current status in a queryFn from what I can see.  That's why the context is passed in to the fn.

app/src/gui/components/notebook/record_table.tsx
stevecassidy: Replaced with a nice green cloud icon and an amber 'pending' icon.

library/data-model/src/data_storage/internals.ts
```diff
@@ -853,74 +868,70 @@ export async function getRecordListAudit({
...
-        $exists: false,
-      },
-    },
-    fields: ['_id', '_rev', 'record_id'],
-    limit: recordIds.length * 100, // Adjust based on expected documents per record
-  });
+      // we'll only fetch _id and _rev to compute the audit hash
+      fields: ['_id', '_rev'],
+      limit: batch.length + 10,
+    });
```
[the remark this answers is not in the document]
stevecassidy: as discussed, I think this is fast enough for now and we can look at other optimisations later if necessary.

PR #1536, opened 2025-06-25
Improved Backup and Restore

stevecassidy:
# Improved Backup and Restore

## JIRA Ticket

None

## Description

Following a failure in trying to restore a large backup, this PR has some changes to improve both the backup and restore processes.

## Proposed Changes

Update the on-device backup process to use low-level methods to get the databases to dump containing records.  This prevents a problem we've seen recently where the backup failed because (we think) the internal data structures referencing databases were broken.  To avoid this in future, since this is an emergency procedure, we just look at the low level IDB database names to find dataDBs to back up.

For restoring backups, there was a memory leak that prevented large backups being restored.  I don't think it was a genuine leak but due to the rapid reading and writing of records, the GC did not catch  up and heap memory was exceeeded.  Here we add a batch write operation to write many documents at once to the database and use this to batch up writes from the restore process. 

Previous restore operation was via a special page on Conductor.  I've moved it to a package.json script which makes more sense for this operation.

I've also added the option to select which databases to restore from a backup file. 

Restoring a backup no longer overwrites existing documents by default.  This should mean that we can restore only the missing documents from a backup.

I added documentation about the backup format and process in the developer docs.

Successfully tested on a large (8G) backup from a device.  

## How to Test

Needs to be tested on device (can't save the backup on web).   Save a backup from a device having a bunch of records.   Try restoring this backup to a different server.

## Checklist

### DIFF
--- a/docs/developer/docs/source/markdown/Backup.md
+++ b/docs/developer/docs/source/markdown/Backup.md
@@ -1,0 +1,49 @@
+# On Device Backup of Data
+
+There is an option in the app to share a backup dump of the on-device databases.  This is
+intended as a safety measure. If a user is in the field and is concerned about data loss,
+they can create a local backup of data that could be recovered at a future time.
+We've also used this when devices have refused to sync data even though they have
+a network connection (due to an unknown bug).
+
+The backup option is available via the About Build screen in the app.  Clicking the
+backup button initiates a database dump. The dump file is then available to
+the device 'share' service which allows the user to save it as a local file or
+send it to a service (GDrive, email etc).
+
+## Dump Format
+
+The dump format is JSONL with one JSON line per record, interspersed with header
+lines.  At the start of each database dump the header line is as follows:
+
+```JSON
+{type: 'header', database: db.name, info: info}
+```
+
+Where `db.name` is the database name and `info` is the result of calling the `.info()`
+method on the database.
+
+Following lines are just raw documents from the database as single line JSON documents.
+No ordering is done on the documents, they are just the result of calling `allDocs()` on
+the database.
+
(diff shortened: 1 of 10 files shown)


docs/developer/docs/source/markdown/Backup.md
```diff
@@ -0,0 +1,49 @@
...
+
+This would restore records from the backup file to any databases matching the
+regular expression `pattern`.  
+
+By default, existing records will not be overwritten when the backup is restored.
+To overwrite existing records, add `true` to the end of the command line:
+
+```shell
+npm run restore-backup fieldmark-backup-2025-06-19-1750363850254.jsonl pattern true
+```
```
PeterBaker0, line 49: nitpick - but typically you'd use a flag e.g. --force rather than a positional argument. This is easier to maintain as the script expands in complexity

library/data-model/src/data_storage/dataDB/design.ts
```diff
@@ -237,20 +237,6 @@ export const recordAuditDocument = {
...
-          emit(
-            [doc.record_id, doc.attach_format_version ? 'attachment' : 'data'],
-            {
-              _id: doc._id,
-              _rev: doc._rev,
-            }
-          );
-        }
-      }),
-    },
```
PeterBaker0, line 253: Intentional removal?

PeterBaker0 (approved):
Looks good - couple of nitpicks

library/data-model/src/data_storage/dataDB/design.ts
stevecassidy: Yes, was not used.

PR #1539, opened 2025-06-27
Control Center: refresh auth token after template/project creation

stevecassidy:
# Control Center: refresh auth token after template/project creation

## JIRA Ticket

[BSS-1003](https://jira.csiro.com/browse/BSS-1003)

## Description

When I create a new template or project in Control Center while logged in as a non-admin user, the list of templates or projects is not updated until I force a page refresh.  This is due to the auth token not being updated with new permissions after creation.

## Proposed Changes

Force refresh of the auth token after making a template or project.  Also, add the user token to the useQuery dependencies so they are re-fetched when the token changes.

Also fixes a minor issue with template creation where the template id was derived from the metadata.name property rather than the name.   This then requires that we pass the name property into createTemplate. 

## How to Test

In CC, create a template or notebook, note that it appears in the list of templates/notebooks immediately you close the dialog.

## Checklist

### DIFF
--- a/web/src/components/tabs/teams/team-templates.tsx
+++ b/web/src/components/tabs/teams/team-templates.tsx
@@ -9,7 +9,6 @@ import {useNavigate} from '@tanstack/react-router';
 
 const TeamTemplates = ({teamId}: {teamId: string}) => {
   const {user} = useAuth();
-  console.log('teamTemplates user', user);
   const {isPending, data} = useGetTemplatesForTeam({user, teamId});
 
   // can the user see the add button?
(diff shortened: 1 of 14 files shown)


web/src/components/tabs/teams/team-templates.tsx
```diff
@@ -9,7 +9,7 @@ import {useNavigate} from '@tanstack/react-router';
 
 const TeamTemplates = ({teamId}: {teamId: string}) => {
   const {user} = useAuth();
-
+  console.log('teamTemplates user', user);
```
PeterBaker0, line 12: remove

web/src/hooks/queries.ts
```diff
@@ -173,7 +173,7 @@ export const useGetTemplatesForTeam = ({
   teamId: string;
 }) =>
   useQuery({
-    queryKey: ['templatesbyteam', teamId],
+    queryKey: ['templatesbyteam', teamId, user?.token],
```
PeterBaker0, line 176: same here

PR #1557, opened 2025-07-09
feat: duplicate condition on split/add in condition editor

ryankontos:
# feat: duplicate condition on split/add in condition editor

## Description

Improves the condition editor UX by duplicating the current condition when:
- a user splits a single condition into an `and/or` group (via the "Split" button),
- or adds a new condition to an existing `and/or` group.

This supports the common case where users want to define multiple values for the same (non multi select) field (e.g., `status = active OR status = pending`) without re-selecting the field and operator each time.

## Proposed Changes

- `BooleanConditionControl`:  
  `addCondition()` now duplicates the first existing condition (or uses `EMPTY_FIELD_CONDITION` if none).
- `FieldConditionControl`:  
  `handleSplitCondition()` now creates an AND group with two copies of the current condition.

## How to Test

1. In Designer, open the condition editor for a section or field.
2. Click the “Split” button:
   - Expected: the new group has two identical conditions.
3. Add a third condition using “Add another condition”:
   - Expected: new condition is a duplicate of the first in the group.
4. Modify one of the values:
   - Expected: only that condition updates; others remain unchanged.
5. Repeat with other target fields types.

## Checklist

### DIFF
--- a/web/src/designer/components/condition/ConditionControl.tsx
+++ b/web/src/designer/components/condition/ConditionControl.tsx
@@ -129,10 +129,15 @@ const BooleanConditionControl = (props: ConditionProps) => {
   const addCondition = () => {
     if (condition) {
       const existing = condition.conditions || [];
-      // construct a condition with a new empty field condition
+      // Duplicate the first existing condition (or use empty condition if none)
+      const template =
+        existing.length > 0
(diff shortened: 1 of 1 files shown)


stevecassidy (approved):
Excellent!

PR #1558, opened 2025-07-10
feat: Add Save/Cancel buttons to condition editor

ryankontos:
# feat: Add Save/Cancel buttons to condition editor

## Description

This PR adds **Save** and **Cancel** functionality to the condition editor modal.

Currently, any changes to conditions were applied immediately in redux. This could lead to accidental or partial edits being committed without confirmation. This change makes sure users can review changes and choose to either commit to them or discard them.

## Proposed Changes

- Refactored `ConditionModal` to:
  - Store a local `draft` of the condition
  - Only call `onChange` when the user clicks **Save**
  - Add **Cancel** button with confirmation if unsaved edits exist
- Updated modal layout to include `DialogActions` for buttons
- Minor tidy-up: removed now-unused state management logic

## How to Test

1. Open Designer
2. Click the button to create/edit a condition to open the modal.
3. Make some changes:
   - Click **Cancel** → verify you're prompted if edits were made.
   - Click **Save** → verify the changes are applied to the form.
4. Ensure that unmodified edits do not trigger unnecessary updates.

## Checklist

- [x] I have added JSDoc-style comments to any new functions or classes.

### DIFF
--- a/web/src/designer/components/condition/ConditionModal.tsx
+++ b/web/src/designer/components/condition/ConditionModal.tsx
@@ -12,33 +12,88 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {Button, Dialog} from '@mui/material';
-import {useState} from 'react';
+import {
+  Button,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  Stack,
+} from '@mui/material';
+import {useCallback, useEffect, useState} from 'react';
 import {ConditionControl} from './ConditionControl';
-import {ConditionProps} from './types';
+import {ConditionProps, ConditionType} from './types';
 import QuizIcon from '@mui/icons-material/Quiz';
 
 export const ConditionModal = (props: ConditionProps & {label: string}) => {
   const [open, setOpen] = useState(false);
+  // Local draft copy
+  const [draft, setDraft] = useState<ConditionType | null>(
+    props.initial ?? null
+  );
+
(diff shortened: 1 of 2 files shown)


web/src/designer/components/condition/ConditionModal.tsx
```diff
@@ -12,33 +12,88 @@
...
+  };
+
+  const close = () => setOpen(false);
+
+  // Warn when cancelling
+  const handleCancel = useCallback(() => {
+    const changed =
+      JSON.stringify(draft) !== JSON.stringify(props.initial || null);
+    if (changed) {
+      const confirm = window.confirm(
```
stevecassidy, line 53: Good idea but can you use a dialog component rather than the system confirm.  What's here is confusing since there is a 'cancel' button in the confirm dialog which cancels the cancellation...

web/src/designer/state/uiSpec-reducer.ts
```diff
@@ -23,6 +23,7 @@ import {
   removeFieldFromSummaryForViewset,
   replaceFieldInCondition,
 } from './helpers/uiSpec-helpers';
+import {v4 as uuidv4} from 'uuid';
```
stevecassidy, line 26: Just spotted this and fixed in another PR.

stevecassidy (changes requested):
Good but one little change requested.

PR #1559, opened 2025-07-11
Facilitate migration from v1.0

stevecassidy:
# Facilitate migration from v1.0

## JIRA Ticket

None

## Description

Users of the v1.0 app would need to reset devices to use the new app and would lose any data on device that was not synced.

## Proposed Changes

Adds a placeholder page at /auth on conductor with a message about updating the app, this should catch any existing users who try to login with the old app.

Adds a migration in the app to rename any existing dataDbs to use the new name format.   The old database is then deleted.  This means that when the user re-activates the notebook, they will see whatever data was present before the update without having to sync.  This should include any data on device that had not been synced before.

## How to Test

Complex to test.   Set up a v1.0.0 server and create a notebook, create some records, activate the notebook in the app and sync.   

Shut down the v1.0.0 server. Run migrate from current main, run the api server, run the app.  On refreshing the app page in the browser, the migration should happen.  Note the messages in the console.

Tested in browser but not on device as it would be too complex to establish the same sequence.  Hoping that it would work as expected!

## Checklist

### DIFF
--- a/api/src/auth/authRoutes.ts
+++ b/api/src/auth/authRoutes.ts
@@ -89,6 +89,15 @@ export const DEFAULT_REDIRECT_URL = WEBAPP_PUBLIC_URL + '/auth-return';
  * @param socialProviders an array of login provider identifiers
  */
 export function addAuthRoutes(app: Router, socialProviders: AuthProvider[]) {
+
+  // For legacy versions of the app, we provide a message on /auth to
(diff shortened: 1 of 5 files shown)


PeterBaker0 (approved):
Nothing glaringly wrong - do other aspects migrate okay i.e. all the changes to local storage etc?

PR #1564, opened 2025-07-16
Remove frequent JWT error logging

stevecassidy:
# feat/fix/chore: pr title

## JIRA Ticket

None

## Description

I was seeing very frequent errors logged by the api about expired JWTs. This is not useful.

## Proposed Changes

Remove the output.

## Checklist

### DIFF
--- a/api/src/auth/keySigning/read.ts
+++ b/api/src/auth/keySigning/read.ts
@@ -19,7 +19,7 @@
  */
 
 import {decodeAndValidateToken} from '@faims3/data-model';
-import {jwtVerify} from 'jose';
+import {jwtVerify, errors as joseErrors} from 'jose';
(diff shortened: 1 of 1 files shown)


PeterBaker0 (changes requested):
If you're going to look at this now, I think a better approach would be to have a set of custom/known exceptions for the JWT validation e.g. expired, invalid server claim, user not present etc, and throw this error, but catch it and handle it how we want in the client code. Swallowing the error for the whole function with no logging could make this a bit tricky to manage if we are seeing weird edge case errors that aren't expiry such as "No sub claim" or "User not present in database".

PeterBaker0 (approved):
assuming you've tested this works - looks good - thanks

PR #1571, opened 2025-07-21
Deal with loading states for teams

stevecassidy:
# Deal with loading states for teams

## JIRA Ticket

None

## Description

Previous PR didn't really handle the loading state from useGetTeams properly.

## Proposed Changes

Handle the case of teams still loading, no teams and no permission differently. 

TODO:  This could be finessed to also deal with the isError state and show a useful error message.  We could maybe have a component to use in these cases and deploy it through the app. 

## How to Test

Generally no visible change unless there is an error/no permission.

## Checklist


PeterBaker0 (approved):
Small change but let's close up this very stale PR. Still worth merging.

Issue #1581, opened 2025-07-30
BUG: Adding the first autoincrementer field to a notebook breaks it


stevecassidy:
If there is no existing auto incrementer in a notebook then adding one triggers a bug - in api/src/couchdb/notebooks.ts it tries to get the existing local-autoincrementer document but that doesn't exist, the crash results in a 404 error to the request. 

We don't use that document ever, just remove this bit of code.

PR #1588, opened 2025-07-31
BUG: Adding the first autoincrementer field to a notebook breaks it

stevecassidy:
# BUG: Adding the first autoincrementer field to a notebook breaks it

## JIRA Ticket

[#1581](https://github.com/FAIMS/FAIMS3/issues/1581)

## Description

If there is no existing auto incrementer in a notebook then adding one triggers a bug - in api/src/couchdb/notebooks.ts it tries to get the existing local-autoincrementer document but that doesn't exist, the crash results in a 404 error to the request.

We don't use that document ever, just remove this bit of code.

## Proposed Changes

Removes the code that checks for auto incrementers and creates the document in the metadata db.   This document was never read so this is no loss.

Also fix a bug in the redraw of the DesignerWidget that caused a crash when I saved my new notebook, check for 'empty' notebook was incorrect.

## How to Test

Create a notebook from scratch and then add an Auto Incrementer field.  Previously it would crash when you tried to save the notebook.  Now it won't.

## Checklist

### DIFF
--- a/api/src/couchdb/notebooks.ts
+++ b/api/src/couchdb/notebooks.ts
@@ -252,52 +252,6 @@ const generateProjectID = (projectName: string): ProjectID => {
   return `${Date.now().toFixed()}-${slugify(projectName)}`;
 };
 
-type AutoIncReference = {
-  form_id: string;
-  field_id: string;
-  label: string;
-};
-
-type AutoIncrementObject = {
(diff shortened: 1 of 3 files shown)


ranisa-gupta16 (approved):
Thanks for the fix. lg.:-)

Issue #1576, opened 2025-07-25
BUG: Markdown in select fields options doesn't display formatted


stevecassidy:
Add some markdown to a select option as it says you can but the text isn't displayed with formatting in the editor or in the app.

PR #1590, opened 2025-07-31
Render markdown in select fields and in the value of multiselect

stevecassidy:
# Render markdown in select fields and in the value of multiselect

## JIRA Ticket

[1576](https://github.com/FAIMS/FAIMS3/issues/1576)

## Description

We support markdown text in multiselect but not in select fields. Also when you select an item with
markdown formatting in a multi-select it is not rendered properly in the field.

## Proposed Changes

Adds formatting in the right places.

## How to Test

Create a select and multi-select with formatting and note the presentation of the labels.

## Checklist


PeterBaker0 (approved):
Looks fine to me

PR #1596, opened 2025-08-04
New E2E testing setup to generate screenshots

stevecassidy:
# New E2E testing setup to generate screenshots

## JIRA Ticket

[BSS-991](https://jira.csiro.au/browse/BSS-991)

## Description

The old e2e tests were very out of date. Need to rebuild them from scratch.  Do this with
a specific focus on generating screenshots that will be useful in writing user
documentation.  

## Proposed Changes

Establishes a new testing framework for both the web app and mobile that can generate
screenshots during tests.  

Tests can be run in browser for the most part but mobile based testing will be supported for
at least some of the tests (all if possible). 

Generates screenshots in a folder named for the configured theme.  We should be able to use
these in documentation via suitable configuration.  Suggest we keep screenshots in VC and 
update them as needed when the UI changes. 

## How to Test

Run the tests in the e2e folder. 

## Checklist


PeterBaker0 (approved):
I like the framework and approach. 

I had some issues on my platform as I'm headless. Added headless configurations - screenshots still work. 

Had some race condition/throttle issues with excessive parallel headless browsers so set that limit to 1 to avoid parallelisation issues. 

Also I think we missed some base paths as I was getting 'URL '/' is invalid errors from the setUrl('/') calls - just added the base path which sorted that out. 

Working now can run the full suite. 

My request would be to add documentation which describes the 'nothing to e2e tests running' use case. 

Something like a) ./dev.sh b) cd api c) seed data d) e2e tests ... etc.

PR #1606, opened 2025-08-12
New default batch size for PouchDB Sync

stevecassidy:
# New default batch size for PouchDB Sync

## JIRA Ticket

None

## Description

A problem with sync was traced to our use of a large batch size (1000) for database sync which was causing failure when there were many large images attached to records - the resulting request was too big for JSON.stringify and sync just broke.  A smaller batch size ensures that requests are smaller and while sync might be slower it should be more reliable.

## Proposed Changes

Change default batch size to 10.  Add a configuration option for the Github app build workflows to allow this to be adjusted at build time if needed.  

Updated documentation for Android and IOS builds to mention this new variable.

## How to Test

Hard to test but everything should work ok with this smaller batch size.

## Checklist


PeterBaker0 (approved):
👍

PR #1607, opened 2025-08-12
Improved PouchDB debug tool

stevecassidy:
# Improved PouchDB debug tool

## JIRA Ticket

None

## Description

We have a static page minifauxton.html  bundled with the app that can be used to browse the local db. It is a bit clunky and uses an old static version of PouchDB.

## Proposed Changes

Replace this with a React component with similar functionality. Add option to download database contents as JSONL.  Could be extended with other useful debugging tools in future.

## How to Test

Enable the tool with VITE_SHOW_POUCHDB_BROWSER=true and go to About Build, click the button to show the raw database interface. 

## Checklist


PeterBaker0 (approved):
Great- can you just update the table so that if you switch between databases it has a useEffect to reset the pagination to page 0 (if I go from one DB which has > pages than the new one I switch to, I have to page all the way back manually to the start).

Issue #1622, opened 2025-08-29
Need a role within teams that can make notebooks but not manage users


stevecassidy:
Currently we only have 'Member' and 'Manager' within a team, we need a role that can create notebooks in the team and administer their own notebooks but not manage users or other aspects of the team.  Ideally, they should only be able to see their own notebook and data (this is for students in a course who each make their own notebook).

As a first step, maybe give the Member role the ability to make notebooks.

PR #1624, opened 2025-09-01
Add new role 'TEAM_MEMBER_CREATOR'

stevecassidy:
# feat/fix/chore: pr title

## Ticket

#1622

## Description

Currently we only have 'Member' and 'Manager' within a team, we need a role that can create notebooks in the team and administer their own notebooks but not manage users or other aspects of the team. Ideally, they should only be able to see their own notebook and data (this is for students in a course who each make their own notebook).

## Proposed Changes

Add a new role TEAM_MEMBER_CREATOR that can make notebooks but not contribute to other notebooks.   

Permissions to create these roles are the same as for the other member role. 

## How to Test

Test in the control centre/

## Checklist

### DIFF
--- a/library/data-model/src/permission/functions.ts
+++ b/library/data-model/src/permission/functions.ts
@@ -287,12 +287,15 @@ export function teamInviteToAction({
  * @returns The corresponding permission action
  */
 export function getTeamMembershipAction(role: Role, add: boolean): Action {
+  console.log('getTeamMembershipAction', role, add);
   switch (role) {
     case Role.TEAM_ADMIN:
       return add ? Action.ADD_ADMIN_TO_TEAM : Action.REMOVE_ADMIN_FROM_TEAM;
     case Role.TEAM_MANAGER:
       return add ? Action.ADD_MANAGER_TO_TEAM : Action.REMOVE_MANAGER_FROM_TEAM;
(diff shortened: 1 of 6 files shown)


library/data-model/src/permission/functions.ts
```diff
@@ -287,13 +287,16 @@ export function teamInviteToAction({
  * @returns The corresponding permission action
  */
 export function getTeamMembershipAction(role: Role, add: boolean): Action {
+  console.log('getTeamMembershipAction', role, add);
```
PeterBaker0, line 290: can you remove this?

library/data-model/src/permission/functions.ts
```diff
@@ -265,7 +265,7 @@ export function teamInviteToAction({
     } else {
       actionNeeded = Action.DELETE_MANAGER_TEAM_INVITE;
     }
-  } else if (role === Role.TEAM_MEMBER) {
+  } else if (role === Role.TEAM_MEMBER || role === Role.TEAM_MEMBER_CREATOR) {
```
PeterBaker0, line 268: One of the gotchas... gotcha

stevecassidy: indeed

Issue #1635, opened 2025-09-08
Filtering doesn't work in the record table


stevecassidy:
Filter options in the table seem to be broken - any condition just gives me no results.

PR #1636, opened 2025-09-08
Fix record sort and filter

stevecassidy:
# Fix record sort and filter

## Ticket

#1635 

## Description

Filtering and sorting is broken for 'summary fields' in the records table and search only sometimes works.  

## Proposed Changes

The problem was with the way that summary field values in the table were generated, we used renderCell which meant that the value that was being sorted/filtered was not what was shown.   Replace this with valueGetter to return a string value.   This then re-enables sort and filter in the table. 

Search didn't work on values that were arrays. Modified the query to use a regex within arrays as well as for plain values.

## How to Test

In the record table for a notebook with configured summary fields, click on a table header to sort. Use filters via the table header. Search via the search box.

## Checklist

### DIFF
--- a/app/src/gui/components/notebook/record_table.tsx
+++ b/app/src/gui/components/notebook/record_table.tsx
@@ -144,8 +144,8 @@ export const RECORD_GRID_LABELS = {
  * @param record The record to get data from
  * @param column The column type
  * @param uiSpecification ui Spec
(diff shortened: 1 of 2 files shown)


PeterBaker0 (approved):
Looks fine to me - if you have an instance with big number of records might be worth checking performance of new query/searching in general

Issue #1632, opened 2025-09-08
Database Connection Failures


stevecassidy:
## Error in Fieldnote

**Error** in **/notebooks/fieldmark-server/1756644038349-biol3410-vegetation-monitoring/new/Vegetation-assessment/draft/drf-1d3f375b-05e2-4fe2-a2ae-e032b8236720/record/rec-8e7d2e34-9f94-4425-94a0-f91547f227d5**
Unsaved record error:InvalidStateError: Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.

[View on Bugsnag](https://app.bugsnag.com/electronic-field-notebooks/fieldnote/errors/68bb7aa5ede8f8c8cedeeeca?event_id=68bbb49101477693c2e70000&i=gh&m=ci)

## Stacktrace

    au.edu.faims.fieldmark://localhost/assets/index-CemQ_bAR.js:484 - null

[View full stacktrace](https://app.bugsnag.com/electronic-field-notebooks/fieldnote/errors/68bb7aa5ede8f8c8cedeeeca?event_id=68bbb49101477693c2e70000&i=gh&m=ci)

*Created by Steve Cassidy via Bugsnag*

PR #1643, opened 2025-09-15
Add wrapper around PouchDB in the app to intercept connection failures

stevecassidy:
# Add wrapper around PouchDB in the app to intercept connection failures

## Ticket

#1632

## Description

We are seeing intermittent failures of IndexedDB in IOS webview due perhaps to a bug.   It seems that one way around this is to re-open the database connection should it ever disappear.  

## Proposed Changes

Adds PouchDBWrapper in the app to act as a wrapper class for all database connections, mimicing the 
PouchDB.Database interface.   All methods are passed through a retry method that will re-open 
the database connetion if required, up to three times.   

In the normal course of things, this is a no-op wrapper.  Should the connection fail, it will re-open the
connection and try again.  We'll get a logError in this case to see that it is happening. 

## How to Test

Should have no effect on operation unless this bug hits.  All tests pass and it seems to work in casual testing. 

## Checklist

### DIFF
--- a/app/src/context/slices/helpers/pouchDBWrapper.ts
+++ b/app/src/context/slices/helpers/pouchDBWrapper.ts
@@ -18,6 +18,16 @@
  * and mitigate the indexedDB failure that occurs on IOS devices
  */
 
+/*
+ * This wrapper should have the same interface as PouchDB.Database
+ * so that it can be used as a drop-in replacement.  All operations
+ * then go through the withRetry() method which will try to re-connect
+ * to the database if one of the known errors occurs.
+ *
+ * This should have no performance impact apart from the additional
+ * function call overhead.
+ */
+
 import {DatabaseInterface} from '@faims3/data-model';
 import {RUNNING_UNDER_TEST} from '../../../buildconfig';
 import {logError} from '../../../logging';
@@ -139,7 +149,6 @@ export class PouchDBWrapper<T extends {}> implements DatabaseInterface<T> {
     doc: PouchDB.Core.PutDocument<T & Model>,
     options?: any
   ) {
-    console.log('db.pub', doc, options);
     if (options)
       return this.withRetry(() => this.db.put<Model>(doc, options), 'put');
     else return this.withRetry(() => this.db.put<Model>(doc), 'put');
--- a/app/src/context/slices/helpers/pouchDBWrapper.ts
+++ b/app/src/context/slices/helpers/pouchDBWrapper.ts
@@ -184,6 +184,8 @@ export class PouchDBWrapper<T extends {}> implements DatabaseInterface<T> {
     return this.withRetry(() => this.db.query(fun, opts), 'query');
   }
 
+  // security is not implemented for local browser based databases but we need
+  // to include the method to match the interface
   async security() {}
 
   async destroy() {
--- a/app/src/context/slices/helpers/pouchDBWrapper.ts
+++ b/app/src/context/slices/helpers/pouchDBWrapper.ts
@@ -1,0 +1,188 @@
+/*
+ * Copyright 2021, 2022 Macquarie University
+ *
+ * Licensed under the Apache License Version 2.0 (the, "License");
+ * you may not use, this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND either express or implied.
+ * See, the License, for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Description:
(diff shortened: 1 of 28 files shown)


app/src/context/slices/helpers/pouchDBWrapper.ts
```diff
@@ -0,0 +1,188 @@
...
+    this.db = new PouchDB(this.name, LOCAL_POUCH_OPTIONS);
+  }
+
+  // Execute an operation with retry on certain errors
+  private async withRetry<R>(
+    operation: () => Promise<R> | R,
+    operationName: string
+  ): Promise<R> {
+    try {
+      return await operation();
```
PeterBaker0, line 90: Are there any non-idempotent operations which can fail, but still side-effect the DB? 

If an error occurs you retry potentially multiple times - I see you only catch certain errors so hopefully this makes this impossible.

app/src/context/slices/helpers/pouchDBWrapper.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  async get<Model>(id: string, options?: any) {
+    return this.withRetry(() => this.db.get<Model>(id, options), 'get');
+  }
+
+  async put<Model extends {}>(
+    doc: PouchDB.Core.PutDocument<T & Model>,
+    options?: any
+  ) {
+    console.log('db.pub', doc, options);
```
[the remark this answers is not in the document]
PeterBaker0: Agree

app/src/context/slices/helpers/pouchDBWrapper.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  async query<Result extends {}, Model extends {} = T>(
+    fun: string | PouchDB.Map<Model, Result> | PouchDB.Filter<Model, Result>,
+    opts?: PouchDB.Query.Options<Model, Result>
+  ) {
+    return this.withRetry(() => this.db.query(fun, opts), 'query');
+  }
+
+  async security() {}
```
PeterBaker0, line 178: Can you add a comment justifying the empty security function?

app/src/sync/draft-storage.ts
```diff
@@ -38,8 +38,9 @@ import {databaseService} from '../context/slices/helpers/databaseService';
 import {selectProjectById} from '../context/slices/projectSlice';
 import {store} from '../context/store';
 import {logError} from '../logging';
+import {PouchDBWrapper} from '../context/slices/helpers/pouchDBWrapper';
 
-export type DraftDB = PouchDB.Database<EncodedDraft>;
+export type DraftDB = PouchDBWrapper<EncodedDraft>;
```
PeterBaker0, line 43: Why do we use DatabaseInterface in some places, and PouchDBWrapper in others?

library/data-model/src/data_storage/storageFunctions.ts
```diff
@@ -218,12 +220,12 @@ export async function listFAIMSRecordRevisions({
...
   dataDb,
 }: {
-  dataDb: DataDbType;
+  dataDb: DataDbType | RevisionDbType;
 }): Promise<ProjectRevisionListing> {
   try {
     // get all revision
     const result = await queryCouch<Revision>({
-      db: dataDb,
+      db: dataDb as RevisionDbType,
```
PeterBaker0, line 228: all this type casting is scaring me a bit - why do we need to do this now, and not previously?

library/data-model/src/types.ts
```diff
@@ -329,7 +329,91 @@ export type DraftMetadataList = {
...
+
+  // query<Result extends {}, Model extends {} = Content>(
+  //   fun: string | Map<Model, Result> | PouchDB.Filter<Model, Result>,
+  //   opts: PouchDB.Query.Options<Model, Result>,
+  //   callback: PouchDB.Core.Callback<PouchDB.Query.Response<Result>>
+  // ): void;
+  // query<Result extends {}, Model extends {} = Content>(
+  //   fun: string | Map<Model, Result> | PouchDB.Filter<Model, Result>,
+  //   callback: PouchDB.Core.Callback<PouchDB.Query.Response<Result>>
+  // ): void;
```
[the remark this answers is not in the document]
PeterBaker0: agree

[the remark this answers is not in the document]
PeterBaker0: wait - don't we use query in some places?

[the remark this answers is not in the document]
PeterBaker0: oh I see - its below

PeterBaker0 (approved):
Needs thorough testing - assuming you've played around with the app a bunch - activating/deactivating, syncing records etc. There is some changes such as additional type casts, and the change from removeDoc(id, rev) to removeDoc(doc) which warrants some suspicion. Otherwise it looks sensible given the need to mitigate this error.

library/data-model/src/data_storage/storageFunctions.ts
stevecassidy: I'm not sure why it worked with the native type declarations but not these, my types are compatible but I couldn't get a build to work without this.

PR #1662, opened 2025-10-02
feat: scalable zip exports

PeterBaker0:
# feat: scalable zip exports

## What the problem was

1. hydrate records loads the attachments using all_docs into a blob in the the pouchDB response
2. loading this into memory as File requires buffering 
3. even if you remove this, and then fetch directly with pouch, you still get a Buffer object
4. streaming a lot of buffers into a node archiver stream produces backpressure which GC cannot clean up quickly enough

## Solution

Use nano's couch DB client - this has a get attachment as stream method https://github.com/apache/couchdb-nano?tab=readme-ov-file#dbattachmentgetasstreamdocname-attname-params

This is a direct NodeJS.Streamable - this can be passed straight into archiver. 

This alleviates memory leak. 

## Other improvements

Added record iterator attachment flag. 

**Note** - the record hydration logic needs serious re-thinking or at least more fine-grained control. The behaviour is opaque and confusing. The _loader_ functions is a registered method (making debugging confusing) - and it's behaviour is opaque. It performs some 'loading' which puts the 'data' into the 'data' field of the AVP - which is then later put into the .data property of the notebook iteration object. The includeAttachments optional flag is passed all the way down to disable this behaviour. But we still want to expose the faims attachment info - so instead of loading the blob we put the attachment info object into the .data of avp so that we can access it all the way out at the top. This is a really confusing implementation. 

Also enables a single ZIP file export for all forms. 

Updates web frontend to support this workflow. 

## Later TODO

- overhaul record iterator function to make it clearer, and more configurable/clear and properly typed
- allow single ZIP which has the desired export format AND the attachments, all in one


api/src/couchdb/attachmentExport.ts
```diff
@@ -0,0 +1,503 @@
...
+ * - External: Memory used by external C++ objects
+ *
+ * Only active when DEVELOPER_MODE is enabled to avoid performance overhead
+ * in production.
+ *
+ * @param stage - Label for this logging checkpoint
+ * @param extraInfo - Additional context information
+ */
+function logMemory(stage: string, extraInfo = ''): void {
+  if (!DEVELOPER_MODE) return;
```
stevecassidy, line 415: In general we don't want this in developer mode all the time, note sure it's worth keeping beyond this PR but if you want to leave it in can you add a more specific setting.

api/src/couchdb/attachmentExport.ts
```diff
@@ -0,0 +1,503 @@
...
+  viewID: string;
+  includeViewIDPrefix: boolean;
+}): Promise<void>[] {
+  const promises: Promise<void>[] = [];
+
+  // Iterate through all fields in the record's data
+  for (const fieldId of Object.keys(record.data)) {
+    const fieldType = record.types[fieldId];
+
+    console.log(
```
stevecassidy, line 315: remove logging

api/src/couchdb/attachmentExport.ts
```diff
@@ -0,0 +1,503 @@
...
+      | {
+          attachment_id: string;
+          filename: string;
+          file_type: string;
+        }[]
+      | null;
+
+    // Null indicates an incomplete or pending attachment field
+    if (attachmentList === null) {
+      console.log(`[ZIP] Skipping null attachment list in field ${fieldId}`);
```
stevecassidy, line 340: and here

api/src/couchdb/attachmentExport.ts
```diff
@@ -0,0 +1,503 @@
...
+ * - Error handling: Warnings logged, errors thrown
+ *
+ * @param outputStream - The destination stream for the ZIP archive
+ * @returns Configured archiver instance ready to accept files
+ */
+function createConfiguredArchive(
+  outputStream: NodeJS.WritableStream
+): archiver.Archiver {
+  const archive = archiver('zip', {
+    zlib: {level: 9}, // Maximum compression for smaller file sizes
```
stevecassidy, line 242: Just did some tests, on a 6.5G collection of images, level 9 compression only shrinks it to 6.4G, level 1 does the same.  Since JPEG images are already compressed we might be better using a lower level here to make it faster/less resource hungry.

stevecassidy (changes requested):
Overall looks great, well done for finding a way through the memory jungle!  Just a few things to finalise.

stevecassidy (approved):
Looks good!

PR #1700, opened 2025-10-16
build: improved local dev

PeterBaker0:
Adds --db-only mode to the local dev script which only runs the couch service in docker. Improves default behaviour of the `env.dist` file for API to not use the docker only `couchdb` label for the couchdb URL. This is overridden in `docker-compose` instead. 

Adds `turbo dev` which live reloads all dev targets.

Adds a `dev.sh` script which combines this behaviour for a 'one click' run with live local reloading services, no manual intervention needed to `.env` files.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ COPY turbo.json .
 # monorepo package* which are relevant here
 COPY api/package.json api/package-lock.json ./api/
 COPY app/package.json ./app/
-COPY web/package.json web/package-lock.json ./web/
+COPY web/package.json ./web/
 COPY library/data-model/package.json library/data-model/package-lock.json ./library/data-model/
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -16,46 +16,89 @@ The repository contains the following:
...
 ```bash
-source ~/.bashrc
+./localdev.sh --all
 ```
 
-You can then setup Node v20 and activate it
+This spins up three services:
+
+- Conductor API (`/api`) live reloading on http://localhost:8080
+- FAIMS3 app (`/app`) live reloading on http://localhost:3000
```
stevecassidy: Should also run /web on 3001

stevecassidy (approved):
Seems to work as advertised.  I don't think I'll use it though...

PR #1702, opened 2025-10-17
fix: data loss when swapping from editing record to info tab

PeterBaker0:
fix: data loss when swapping from editing record to info tab

## Previous sequence of events

- create record normally
- click on existing record
- make edits
- draft is created in the background but the route is never updated to 'track' this state (using routes for state tracking is insane but sobeit)
- you swap to info
- when you come back, all components remount, it re-reads the draftId from the route params, which is undefined, so it thinks this is just an existing record

## Changes proposed

- add a parent state in the record.tsx form (where tabs are tracked) 
- initialise this draftId with the params from the route as the default
- replace all instances of previous draftId from the route, with this state variable
- prop drill the setDraftId deep into the draft storage system
- on the draft created listener, run the setDraftId to the new draft Id

This fixes the issue of data loss, but introduces a new, less serious (but still annoying issue) of having the form re-render (due to the state change), which then means you lose focus in the form (and therefore also position in the sections. 

@stevecassidy  what do you think about this compromise? The only other thing I can think is you could try and force a route swap, but this seems very dangerous and likely to force a worse re-render anyway, or some other nasty route side effects.

### DIFF
--- a/app/src/gui/fields/qrcode/QRCodeFormField.tsx
+++ b/app/src/gui/fields/qrcode/QRCodeFormField.tsx
@@ -92,6 +92,7 @@ const useStyles = createUseStyles({
 
 export interface QRCodeFieldProps extends FieldProps {
   label?: string;
+  disabled?: boolean;
 }
 
 export function QRCodeFormField({


stevecassidy (approved):
A joint effort!

PR #1707, opened 2025-10-21
Invalidate project query after uploading a notebook JSON

stevecassidy:
# Invalidate query after uploading a notebook or template JSON

## JIRA Ticket

None

## Description

When you upload a new JSON for a notebook or template, the notebook is replaced on the server but the app doesn't re-fetch it for a subsequent edit.

## Proposed Changes

Invalidate the project query when we upload JSON.

## How to Test

Download project JSON, make a change in the editor, re-upload the JSON, you should see the change you made reverted.

Ditto for templates.

## Checklist

### DIFF
--- a/web/src/components/dialogs/edit-project-dialog.tsx
+++ b/web/src/components/dialogs/edit-project-dialog.tsx
@@ -16,9 +16,11 @@ import {NOTEBOOK_NAME, NOTEBOOK_NAME_CAPITALIZED} from '@/constants';
  * EditProjectDialog component renders a dialog for editing a project.
  * It provides a button to open the dialog and a form to update the project.
  *
+ * The onSuccess callback is called after a successful update.
+ *
  * @returns {JSX.Element} The rendered EditProjectDialog component.
  */
-export const EditProjectDialog = ({callback}: {callback: () => void}) => {
+export const EditProjectDialog = ({onSuccess}: {onSuccess: () => void}) => {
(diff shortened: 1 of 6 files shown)


web/src/components/dialogs/edit-project-dialog.tsx
```diff
@@ -37,7 +37,7 @@ export const EditProjectDialog = () => {
             new file must be a valid JSON file.
           </DialogDescription>
         </DialogHeader>
-        <UpdateProjectForm setDialogOpen={setOpen} />
+        <UpdateProjectForm setDialogOpen={setOpen} callback={callback} />
```
PeterBaker0: Callback is a bit vague- maybe `onSuccess` ?

web/src/components/forms/update-project-form.tsx
```diff
@@ -47,6 +49,9 @@ export function UpdateProjectForm({
       }
     );
 
+    // call the callback to invalidate the project query
+    callback();
```
PeterBaker0: This component doesn't know what this callback does - we shouldn't put a comment here to this effect as may confuse future us if a diff callback is used later which no longer does this.

PeterBaker0 (approved):
Some minor comments - but looks good otherwise - would be nice if we only encoded the query keys in one place - bit fragile at the moment. You could have functions e.g. 'generateSomeOperationQueryKeys' which predictably builds them, for use in either place

PR #1730, opened 2025-10-31
feat: db interface

PeterBaker0:
# TODO

- [ ] refactor to pass in the equality fn stuff as part of main constructor rather than configure step
- [ ] test out CRUD logic with some tests, particularly the creating new things e.g. upsert
- [ ] (maybe) handle more fine-grained operations which the form needs such as linking documents?
- [ ] handle attachment loading/dumping (in a less convoluted way)

### DIFF
--- a/library/data-model/src/databaseEngine/engine.ts
+++ b/library/data-model/src/databaseEngine/engine.ts
@@ -11,6 +11,7 @@ import {
   ExistingRecordDBDocument,
   ExistingRevisionDBDocument,
   getDocumentType,
+  HydratedDataField,
   HydratedRecord,
   isExistingAttachmentDocument,
   isExistingAvpDocument,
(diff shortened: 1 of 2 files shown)


library/data-model/src/databaseEngine/engine.ts
```diff
@@ -0,0 +1,1209 @@
...
+   * NOTE we may wish to consider moving the ID generation functions to this
+   * module, and a 'seed' method rather than create which allows for
+   * incomplete/empty data.
+   *
+   * @param formRecord - The form data to create a record from
+   * @returns The ID of the newly created revision
+   * @throws Error if revision_id is not null (should use updateRecord instead)
+   * @throws Error if record creation fails
+   */
+  async createRecord(formRecord: FormRecord): Promise<string> {
```
stevecassidy: Assumes that formRecord already has a suitable ID - who creates this and how?

stevecassidy: Also since I'm creating a record from scratch here we won't have a revision_id.

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -0,0 +1,1209 @@
...
+    const validated = formRecordSchema.parse(formRecord);
+
+    if (validated.revision_id !== null) {
+      throw new Error(
+        'Cannot create a new record with a non-null revision_id. Use updateRecord instead.'
+      );
+    }
+
+    // Generate new revision ID
+    const revisionId = generateRevisionID();
```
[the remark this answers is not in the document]
stevecassidy: and this suggests you're ignoring the revision_id in the input record object anyway.

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -0,0 +1,1209 @@
...
+          : !(await equalityFn(parentAvp.data, fieldValue));
+
+      const hasAnnotationChanged =
+        parentAvp === undefined
+          ? true
+          : !(await equalityFn(parentAvp.annotations, fieldAnnotation));
+
+      if (hasDataChanged || hasAnnotationChanged) {
+        // Create new AVP
+        const avpId = generateAvpID();
```
stevecassidy: Here we're creating the id inside the creator function, which I think is the right thing to do

library/data-model/src/databaseEngine/check.ts
```diff
@@ -0,0 +1,826 @@
...
+  const testHydration = testHydrationIndex !== -1;
+
+  // Remove the flag from args if present
+  if (testHydrationIndex !== -1) {
+    args.splice(testHydrationIndex, 1);
+  }
+
+  if (args.length !== 3) {
+    console.error(
+      'Usage: tsx validate-database.ts <url> <username> <password> [--test-hydration]'
```
stevecassidy: That's not the name of the script, how are we going to use this if at all?  If it's going to be useful maybe make it a script in the api project.  Refactor so that the core functions could be called from web if this kind of audit is useful.

PeterBaker0: Let's leave here for now as it's helpful for testing out the engine - we can then decide to keep or integrate properly into the CLI/API.

library/data-model/src/databaseEngine/types.ts
```diff
@@ -0,0 +1,410 @@
...
+  // The project this belongs to
+  projectId: z.string().optional(),
+  // The ID of the form (also known as viewsetId)
+  formId: z.string(),
+  // The actual form data Map string -> any
+  data: z.record(z.string(), z.unknown()),
+  // Annotations
+  annotations: z.record(z.string(), annotationsSchema),
+  // Author
+  createdBy: z.string(),
```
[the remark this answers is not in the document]
stevecassidy: also since hydrated records have created_by, different bits of code will need a different convention.

stevecassidy (approved):

PR #1734, opened 2025-11-05
feat: attachment engine

PeterBaker0:
A separate service injection pattern attachment engine. 

Keeping outside of main DB interface, as attachments should be interchangeable in the future. 

Fields will need to be injected with context to allow attachment management.


stevecassidy (approved):
Nice

PR #1735, opened 2025-11-05
feat: form context

PeterBaker0:
Drills down a type discriminated form context.

Mode: 'full' | 'preview' 

Full = full power!
Preview = for form previews - can still edit values etc but special stuff like attachments etc may not work. The field can use this to intelligently render different versions.


PeterBaker0:
Discussed offline as suitable - will merge as low risk.

PR #1741, opened 2025-11-12
feat: revised form level engine controls

PeterBaker0:
Implements our discussed API surface for forms. 

- createRecord - now only creates an 'empty' record  + revision - no data
- createRevision - now creates only a new rev/fork of an existing revision, no data changes
- updateRevision - the only place to apply data changes, very carefully diffs and applies AVP updates, context aware to existence of parent revision (in 'parent' mode) - NOTE: doesn't handle the 'two parent' case - open to ideas on what to do here if we have a merge conflict parent? But unsure under what conditions this occurs.
- getExistingFormData - revised to return the exact data specification needed for updateRevision.

### DIFF
--- a/library/data-model/src/databaseEngine/engine.ts
+++ b/library/data-model/src/databaseEngine/engine.ts
@@ -1,7 +1,7 @@
-import {isEqual as deepEquals} from 'lodash';
 import {v4 as uuidv4} from 'uuid';
 import {isEqualFAIMS} from '../datamodel';
 import {DatabaseInterface, UISpecification} from '../types';
+import {differenceSets} from '../utils';
 import * as Exceptions from './exceptions';
 import {
   AvpDBDocument,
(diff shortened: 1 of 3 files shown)


library/data-model/src/databaseEngine/engine.ts
```diff
@@ -1106,99 +1141,386 @@ class FormOperations {
...
+        break;
+
+      case 'reuse_parent':
+        if (!parentRevision) {
+          throw new Error(
+            'Cannot apply the reuse_parent AVP strategy when parentRevision is not provided.'
+          );
+        }
+        // Reference the parent's AVP (user reverted to previous version)
+        outputAvpMap[fieldname] = parentRevision.avps[fieldname];
```
stevecassidy: if we're re-using the parent because an edit has changed a value back to a parent, where do we delete the previously created avp?

Issue #1744, opened 2025-11-14
Use PreviewFormManager to preview forms in Designer


stevecassidy:
We should be able to add PreviewFormManager to designer and get a live preview of the current form.

PR #1759, opened 2025-11-14
Preview forms in Designer

stevecassidy:
# Preview forms in Designer

## Ticket

#1744

## Description

Adds a preview of the current form to the designer using the new FormManager component.

## How to Test

Edit a notebook (the simple-notebook.json) in the dashboard and you will see a preview to the side of the editor (need a reasonably wide screen).
Note this will only work for fields that we have in new forms (Text and Select) for now.

Update the notebook (add a field, edit a field) and see the change reflected in the preview.

## Checklist


web/src/designer/components/form-editor.tsx
```diff
@@ -74,6 +78,12 @@ export const FormEditor = ({
   const searchParams = new URLSearchParams(location.search);
   const sectionParam = searchParams.get('section');
 
+  const uiSpec = useAppSelector(
+    state => state.notebook['ui-specification'].present
+  );
+  // we need this to be a ProjectUIModel type for the PreviewFormManager
+  const uiSpecInternal: ProjectUIModel = {views: uiSpec.fviews, ...uiSpec};
```
PeterBaker0: nit, but shouldn't you remove fviews from the resulting object? rather than spread it?

PR #1773, opened 2025-11-27
feat: form validation

PeterBaker0:
Whole heap of work on validation. Pretty heady stuff. Getting close - discussions with Elana around UX. Ideas around validation 'modes' which will effect how we filter schemas. The core is here. 

Note @stevecassidy  the main contentious decision I made here was in pushing out attachment data into the data field in the photo and attachments - this is the simplest way to avoid attachments being in scope for validation. This would have complicated the schema a lot. We can revise if needed but happy with this for now.


stevecassidy:
Not in this PR but I note that the Finish button doesn't do anything any more in the EditableFormManager - no navigation back to the record list.

PR #1775, opened 2025-11-28
Lock out after failed login attempts

stevecassidy:
# Lock out after failed login attempts

## Ticket

None

## Description

In response to the issue flagged by the penetration testers, implements an increasing lockout on failed login attempts to mitigate brute force attacks.

## Proposed Changes

Introduces a delay after a number of failed logins. The delay gets longer the more failed attempts there are.  

## How to Test

Try to login with bad credentials, after 5 attempts you should see a message about a delay.

## Additional Information

This is the most basic implementation and can be improved upon, eg. introducing required CAPTCHA after failed logins, notifying users of failed login attempts, introducing MFA.  Some of these should be flagged for future implementation.

## Checklist

### DIFF
--- a/api/src/auth/authRoutes.ts
+++ b/api/src/auth/authRoutes.ts
@@ -73,7 +73,6 @@ import {
 import {upgradeCouchUserToExpressUser} from './keySigning/create';
 import {verifyUserCredentials} from './strategies/localStrategy';
 import {AuthProviderConfigMap} from './strategies/strategyTypes';
-import {log, time} from 'console';
 
 patch();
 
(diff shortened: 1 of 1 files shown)


PeterBaker0 (approved):
Seems sensible. It is worth noting that this in-memory store could be a vulnerability as a nice simple DDOS target. Just keep spamming different usernames (with distributed cluster) and you could fill it up pretty quick. I think combination of overall rate limiting + a size limit on this in-memory object might be sufficient?

PeterBaker0 (approved):
As discussed, some concessions here, but better than it was- thanks

Issue #1762, opened 2025-11-21
[BUG]  Catch back/commit not saving due to debounce


PeterBaker0:
If you go back/finish too quickly before the changes debounce, no changes are saved

PR #1787, opened 2025-12-09
feat: flush debounced changes on all navigation events

PeterBaker0:
#1762 
Tanstack form debounce doesn't expose flush controls. 

Implements custom debounce wrapper and integrates into navigation button and browser navigation events. 

The browser navigation requires the client context (e.g. react-router-dom). This means we have dynamically built events which are produced by a component which need to be called by the parent. The only way to do this is to pass information 'back' from child component to parent. React can do this with forward Refs, but I prefer the callback style. This is not an ideal pattern but is relatively simple in this case. 

The useBlocked react-router-dom also requires the new react router data API - AI was competent at re-writing this App.tsx file with the new API, all features seem to be functional in testing, but if we see issues with routing or protected routes, we may wish to review this. We could fall back instead to browser events- but these are a bit more complex. We may also wish to utilise other modern react router features which this upgrade exposes us to.

### DIFF
--- a/library/forms/lib/formModule/formManagers/components/NavigationBreadcrumbs.tsx
+++ b/library/forms/lib/formModule/formManagers/components/NavigationBreadcrumbs.tsx
@@ -248,7 +248,7 @@ export const FormBreadcrumbs = ({
           },
         }}
       >
-        {navigateToRecordList.label}
+        Record list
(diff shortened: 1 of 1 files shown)


stevecassidy (approved):
Looks good.

PR #1796, opened 2025-12-10
fix: preview improvements

PeterBaker0:
1) live reloading fix
2) consistent use of mui typography instead of HTML elements
3) stylesheet reset and mui theme reset

**note** we could eventually inject the customised theme instead of a default one - but okay for now imo

### DIFF
--- a/web/src/designer/components/form-editor.tsx
+++ b/web/src/designer/components/form-editor.tsx
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {UISpecification} from '@faims3/data-model';
+import {PreviewFormManager} from '@faims3/forms';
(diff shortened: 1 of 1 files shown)


web/src/designer/components/form-editor.tsx
```diff
@@ -54,7 +56,9 @@ import {findFormExternalUsage} from './condition/utils';
 import {DeletionWarningDialog} from './deletion-warning-dialog';
 import DebouncedTextField from './debounced-text-field';
 import {PreviewFormManager} from '@faims3/forms';
-import {ProjectUIModel} from '@faims3/data-model';
+import {compileUiSpecConditionals, UISpecification} from '@faims3/data-model';
```
stevecassidy: unused imports

stevecassidy (approved):
looks good

PR #1800, opened 2025-12-11
feat: auto incrementer integration into new forms

PeterBaker0:
Injects an incrementer service from the app -> the EditableFormManager through config. 

The EditableFormManager runs an initialisation process on mount, once (tracked by ref), which attempts to set field values for all relevant fields. It does all it can. If it finds errors, it pushes these out through a callback. The callback fires a dialog in the parent. The parent marks back when it's 'resolved' - this triggers the ref to be updated to indicate that the initialisation should run again. This won't result in writing twice to the same field as it checks for an existing value first. This process repeats until the dialog resolves all missing incrementers. 

NOTE: currenlty the user could get 'stuck' if they somehow cannot resolve issue with incrementers.

### DIFF
--- a/app/src/local-data/autoincrement.ts
+++ b/app/src/local-data/autoincrement.ts
@@ -293,7 +293,14 @@ export async function getAutoincrementReferencesForProject(
       if (fieldDetails['component-parameters'].num_digits) {
         try {
           numDigits = Number(fieldDetails['component-parameters'].num_digits);
-        } catch {}
+        } catch (e) {
+          console.error(
+            'num_digits property could not be parsed as a number. Value',
+            fieldDetails['component-parameters'].num_digits,
+            'error',
+            e
+          );
+        }
       }
 
       references.push({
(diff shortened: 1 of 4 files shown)


app/src/local-data/autoincrement.ts
```diff
@@ -272,17 +274,33 @@ export async function getAutoincrementReferencesForProject(
...
-    // TODO are there other names?
-    if (fields[field]['component-name'] === 'BasicAutoIncrementer') {
+  for (const [fieldId, fieldDetails] of Object.entries(fields)) {
+    if (fieldDetails['component-name'] === 'BasicAutoIncrementer') {
+      // Default
+      let numDigits: number = DEFAULT_NUM_DIGITS;
+      if (fieldDetails['component-parameters'].num_digits) {
+        try {
+          numDigits = Number(fieldDetails['component-parameters'].num_digits);
+        } catch {}
```
stevecassidy: do something here

library/forms/lib/formModule/formManagers/utils/autoIncrementInitializer.ts
```diff
@@ -0,0 +1,90 @@
...
+ * from the incrementer service and setting them on the form.
+ *
+ * Only sets values for fields that don't already have a value in initialData.
+ */
+export async function initializeAutoIncrementFields({
+  form,
+  formId,
+  incrementerService,
+  initialData,
+  onMissingRanges,
```
stevecassidy: this is called onIssue in the useIncrementerService hook - would be good to be consistent

stevecassidy (approved):
Happy with this, just a few notes but looks good.

Issue #1767, opened 2025-11-25
Implement map field


PeterBaker0:
Mapping fields may require some additional thought, as they should integrate with offline maps - may need to pass through some context.

PeterBaker0:
We need to consider

a) take point
b) map form field
c) the view only MapRenderer

They should all utilise the offline maps infrastructure where possible? 

@stevecassidy

PR #1804, opened 2025-12-12
Map Field in new forms

stevecassidy:
# Map Field in new forms

## Ticket

#1767

## Description

Implements the MapField in the new form module.

## Proposed Changes

All map related components have been moved over to the @faims3/forms module.  The map field itself and the other map components used in the app are largely unchanged apart from adjustments of imports etc.  

To configure the map source there is a new call required from the app to `setMapConfig()` to pass the map source/key from the configured environment.

## How to Test

Maps should work in the test application in the forms module and in the form preview in designer as well as in the main app itself. 

Test a map field. Look at the overview map for a notebook.  Look at offline maps. All should work as before.

## Additional Information

Noted in passing that OSM maps no longer work due to the style files having maptiler links in them.  Can't see a simple fix so created a new ticket.

## Checklist

### DIFF
--- a/library/forms/lib/fieldRegistry/fields/MapField/index.tsx
+++ b/library/forms/lib/fieldRegistry/fields/MapField/index.tsx
@@ -97,9 +97,6 @@ export function MapFormField(props: FieldProps): JSX.Element {
     drawnFeatures.features &&
     drawnFeatures.features.length > 0;
 
-  // state for visual indicators
-  const [animateCheck, setAnimateCheck] = useState(false);
-
   /**
    * canShowMapNear - can we show a map near this location?
    *
@@ -146,11 +143,6 @@ export function MapFormField(props: FieldProps): JSX.Element {
   ) => {
     if (action === 'save') {
       props.setFieldData(theFeatures);
-      setAnimateCheck(true);
-      setTimeout(() => setAnimateCheck(false), 1000);
-    } else if (action === 'close') {
-      setAnimateCheck(true);
-      setTimeout(() => setAnimateCheck(false), 1000);
     }
   };
 
--- a/app/src/gui/component_registry/bundle_components.ts
+++ b/app/src/gui/component_registry/bundle_components.ts
@@ -218,7 +218,6 @@ registerField(
   AddressField
 );
 
-
 //TODO: DO we need this?
 registerField(
   'formik-material-ui',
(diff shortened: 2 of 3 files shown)


app/src/index.tsx
```diff
@@ -26,6 +26,8 @@ import './index.css';
 import * as serviceWorkerRegistration from './serviceWorkerRegistration';
 import {shouldDisplayRecord} from './users';
 import {localGetDataDb} from './utils/database';
+import {MAP_SOURCE, MAP_SOURCE_KEY, MAP_STYLE} from './buildconfig';
+import {setMapConfig} from '@faims3/forms';
```
PeterBaker0: Would prefer that we configure the components directly with props/config, rather than having a 'register' style workflow. These historically have made the FAIMS code confusing to maintain, as there is no clear link between configuration and the source of that configuration. It also makes testing more confusing as you need to reason about which register function 'runs last'. Also it makes this file 'side effecting' in that it changes behaviour of other modules.

library/forms/lib/fieldRegistry/fields/MapField/index.tsx
```diff
@@ -289,7 +315,51 @@ export function MapFormField({
...
+
+  if (props.required) {
+    return baseSchema.refine(val => val.features && val.features.length > 0, {
+      message: 'A location selection is required.',
+    });
+  }
+
+  // Optional - allow undefined/null or valid schema
+  return baseSchema.optional().nullable();
+};
```
PeterBaker0: Nice ⚡

library/forms/lib/components/maps/map-download.tsx
PeterBaker0: map downloading feels really weird here - so does the implementation details of how a client should implement a tile caching feature. It's imposing constraints on what kind of clients can consume/use the 'reusable' parts of the forms module. Now a client needs to have indexedDB available in their runtime environment. Also we can conceive of different clients having different ways of caching tiles, and also different workflows/experiences around presenting regions to download.

So far in the forms module, any time that the form is asking of the client non forms-specific or platform/environment specific behaviours (e.g. how do I save data, how do I get an attachment, how do I get an incrementer), this behaviour has been specified as a reusable interface (IAttachmentService, DataEngine, IncrementerService), configured/instantiated/setup in the client environment, then passed in as part of the form config. This makes testing, reuse, environment specific implementations etc, much easier.

I think having large chunk of this behaviour in forms is okay, i.e. the map component, map wrapper, map form field etc. But the map downloading, tile caching, etc feels wrong here.

library/forms/lib/components/maps/MapPreview.tsx
PeterBaker0: Error: Unsupported GeoJSON type: undefined

I get an error if I 

a) choose location
b) see map preview
c) edit location
d) clear
e) save

PeterBaker0 (approved):
Lgtm - thanks 🚀

PR #1810, opened 2025-12-15
feat: QR code form field

PeterBaker0:
@stevecassidy will require testing with native build to ensure scanning works - uses similar approach to previous. Re-exports the QR code button component, removes from /app, and then uses from forms module. Has a separate overlay component to help with the whole screen portal effect (again needs testing on app).


stevecassidy (approved):
Tested on IOS, all working as expected.

PR #1816, opened 2025-12-16
feat: DateTimeNow field

PeterBaker0:
Implements DateTimeNow field alongside others. Needs a bit more control over things so doesn't use DateTimeBase but is similar. Adds some helper functions for display/conversion of date times.


stevecassidy (approved):
All good apart from noting the extra time formatting functions.

PR #1817, opened 2025-12-17
feat: other less clear fields, number, text, designer, zod

PeterBaker0:
Implements the remaining missing fields. 

Fully deprecates old formik fields by not showing them in designer, and providing alternatives (where the old entry is not worth preserving).

Removes the validationSchema entirely from designer - json schema still accepts it.

Adds migration to remove this property. 

Tested all new fields with required and not required - will add notebook attachment. 

[1765937141151-new-fields.json](https://github.com/user-attachments/files/24203703/1765937141151-new-fields.json)

Also **note** - zod was at v4 in forms and v3 in other modules (my fault). I have downgraded it in forms. I attempted migrating all modules to v4, but it clashed with peer deps in api which broke the validation for the routes.

### DIFF
--- a/library/forms/lib/fieldRegistry/fields/DateFields/index.tsx
+++ b/library/forms/lib/fieldRegistry/fields/DateFields/index.tsx
@@ -6,18 +6,24 @@
  * - DateTimePicker (datetime-local)
  * - DatePicker (date)
  * - MonthPicker (month)
+ * - DateTimeNow (datetime-local with "Now" button and optional auto-pick)
  */
 
-import React, {HTMLInputTypeAttribute} from 'react';
+import {
+  Button,
+  TextField as MuiTextField,
+  Stack,
+  Typography,
+} from '@mui/material';
+import React, {HTMLInputTypeAttribute, useEffect} from 'react';
 import {z} from 'zod';
-import {TextField as MuiTextField, Typography} from '@mui/material';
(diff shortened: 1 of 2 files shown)


library/forms/lib/fieldRegistry/fields/TextFields/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Field specification for Email.
+ * Validates and captures an e-mail address.
+ *
+ * This replaces the legacy formik-material-ui::TextField with email type.
+ */
+export const emailFieldSpec: FieldInfo = {
+  namespace: 'faims-custom',
+  name: 'Email',
```
stevecassidy: Note that the current 'email' field is actually a TextField with type set to email - this is a new field (which is fine and we should change the prototype in designer/fields.tsx)

stevecassidy: Ah, I see you did!

stevecassidy (commented):
Some  notes on compatibility.  We could consider updating migrateNotebooks.ts to upgrade some fields (email, number*) and maybe even apply it as part of the server side migrations (currently it runs when the notebook loads in designer).

stevecassidy (approved):
Looks good, migrations work well.

PR #1825, opened 2025-12-19
Improved view record page which brings back targeted info function

PeterBaker0:
Brings back info tab with metadata and delete button - deleted all draft functions from the delete button logic.

### DIFF
--- a/app/src/gui/components/notebook/delete.tsx
+++ b/app/src/gui/components/notebook/delete.tsx
@@ -46,6 +46,7 @@ type RecordDeleteProps = {
   projectId: ProjectID;
   serverId: string;
   recordId: RecordID;
+  hrid?: string;
   revisionId: RevisionID | null;
   showLabel: boolean;
   handleRefresh: () => void;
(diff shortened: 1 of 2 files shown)


app/src/gui/components/notebook/delete.tsx
```diff
@@ -172,13 +161,9 @@ export default function RecordDelete(props: RecordDeleteProps) {
...
           <AlertTitle>
-            Are you sure you want to {is_draft ? 'discard draft' : 'record'}{' '}
-            {is_draft ? draft_id : record_id}?
+            Are you sure you want to delete record {recordId}?
           </AlertTitle>
-          You cannot reverse this action! Be sure you wish to proceed
-          {!is_draft
-            ? ' (all associated drafts of this record will also be removed).'
-            : '.'}
+          You cannot reverse this action! Be sure you wish to proceed.
```
stevecassidy: Ah, but it can be reversed. We should have some UI for viewing deleted records and recovering them.

app/src/gui/components/notebook/delete.tsx
```diff
@@ -172,13 +161,9 @@ export default function RecordDelete(props: RecordDeleteProps) {
       >
         <Alert severity="error">
           <AlertTitle>
-            Are you sure you want to {is_draft ? 'discard draft' : 'record'}{' '}
-            {is_draft ? draft_id : record_id}?
+            Are you sure you want to delete record {recordId}?
```
stevecassidy: Can we use the HRID here, recordId is not meaningful.

PR #1849, opened 2026-01-06
Updated the record list to show current record on top not created rec…

ranisa-gupta16:
## JIRA Ticket

https://github.com/orgs/FAIMS/projects/4/views/6?pane=issue&itemId=147881145&issue=FAIMS%7CFAIMS3%7C1848

## Description

Showing the latest update record on top - "last updated" instead of created one as discussed.

## How to Test

- Create a record and view it in record list.
- Check the timestamps for both and update the old record.
- Navigate back to record list, newly updated record should be listed on top than newly created one.
## Checklist

### DIFF
--- a/app/src/gui/components/notebook/record_table.tsx
+++ b/app/src/gui/components/notebook/record_table.tsx
@@ -250,6 +250,7 @@ export function buildColumnFromSystemField({
     case 'LAST_UPDATED':
       return {
         ...baseColumn,
+        field: 'last_updated',
         type: 'dateTime',
         sortable: true,
         valueGetter: params => {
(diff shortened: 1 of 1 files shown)


ranisa-gupta16:
@stevecassidy we will need to work on the mobile view - sorting doesn't make sense here as the data grid cannot sort/filter over the values. This just sets the default sort column to lastUpdated.

stevecassidy:
This doesn't seem to work. The behaviour seems to depend on what fields are displayed, if lastUpdated is in view (on desktop) then it puts the most recent first but if it isn't (narrower desktop) then the ordering is random. On the mobile view it seems to work as expected.  

In testing the desktop view I see that clicking on the 'sort' icons at the top of a column has a random effect - the order of rows changes but they are not sorted.  Something is very broken!

PR #1852, opened 2026-01-09
feat: improved photo performance

PeterBaker0:
Makes some changes to hopefully make photos more stable on lower end devices, improve performance and UX. 

- tidy up the data path by storing directly as blob on native without ever converting to base 64
- reduce the quality (higher JPEG compression) and cap resolution at full HD
- implement optimistic photo loading so that it immediately displays the thumbnail upon capacitor capture, then does the storage in background, then shows the final loaded image


PeterBaker0:
@stevecassidy could you test on android?

stevecassidy (approved):
Looks good, one comment re. attachment saving. Works fine on Android.

PR #1854, opened 2026-01-09
fix: don't show sections with no visible fields

PeterBaker0:
Fix conditions for showing fields to handle case where section is in visibility map but has no visible fields.


stevecassidy (approved):
Looks ok but a bunch of lint changes that our lint disagrees with.

PR #1855, opened 2026-01-09
feat: survey detail initialisation performance

PeterBaker0:
Does more work to async parallelise the fetching and processing of the project details. This is to avoid 

a) the 'one by one' trickle in rendering effect
b) the slow performance


stevecassidy (commented):
Previously the main page would render quickly after login and then populate the notebook list, now it just hangs on the "Logging in please wait..." screen until they are all loaded. I think it's faster but seems slower because of this.

stevecassidy (approved):

PR #1862, opened 2026-01-15
feat: adaptive poll rate sync status

PeterBaker0:
Previously, a stale error state on sync (possibly from old token on app reload - unsure) resulted in cloud having exclamation icon for sync status until clicked, upon which it would refresh and then resolve. 

This PR makes the poll rate adaptive. If open - once per second. If closed - once per 10 seconds. This should resolve issue with stale exclamation. 

Also takes opportunity to let AI do some tidy up and improve docs a bit.


stevecassidy (approved):
Good with one comment.

PR #1867, opened 2026-01-16
feat: online only satellite 

PeterBaker0:
Adds a new 

- VITE_SATELLITE_SOURCE = 'esri' | 'maptiler'
- VITE_SATELLITE_SOURCE_KEY = optional key (for maptiler - replaces ${key} in url)

If configured as above, adds a new XYZ image layer based on the tile store hardcoded url for each of the esri/maptiler satellite layers. Zoom levels are bounded in the code, but view is kept flexible, to allow 'over zooming' on the satellite image and to not disrupt notebook zoom preferences. 

Responds to online environment ,hiding and swapping back to offline vector layer when offline. 

Adds CDK code. 

@stevecassidy  do we need to expose in iOS and Android builds? See the keys above. For DASS we just need esri source - no key.

### DIFF
--- a/infrastructure/aws-cdk/lib/components/front-end.ts
+++ b/infrastructure/aws-cdk/lib/components/front-end.ts
@@ -357,6 +357,15 @@ export class FaimsFrontEnd extends Construct {
       VITE_NOTEBOOK_NAME: props.notebookName,
       VITE_THEME: props.uiTheme,
       VITE_WEBSITE_TITLE: 'Control Centre',
+      // Maps setup for web
+      VITE_MAP_SOURCE: props.offlineMaps.mapSource,
+      VITE_MAP_STYLE: props.offlineMaps.mapStyle,
+      ...(props.offlineMaps.mapSourceKey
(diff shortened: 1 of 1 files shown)


app/src/buildconfig.ts
```diff
@@ -426,13 +426,31 @@ function get_map_style(): MapStylesheetNameType {
   return map_style || 'basic';
 }
 
+function get_satellite_source(): 'esri' | 'maptiler' | undefined {
+  const satellite_source = import.meta.env.VITE_SATELLITE_SOURCE;
+  return satellite_source || undefined;
+}
+
+function get_satellite_key(): string {
+  const satellite_key = import.meta.env.VITE_SATELLITE_SOURCE_KEY;
```
stevecassidy: Use VITE_MAP_SOURCE_KEY since it is only for maptiler, no need for duplicate config.

stevecassidy (commented):
Code looks ok and everything tests out fine.  Be very careful about ESRI and whether you can even do what you're doing here.  Would prefer one config for map sources.  One comment re offline.

stevecassidy (approved):

PR #1873, opened 2026-01-19
feat: map form field point select - use current location

PeterBaker0:
Adds a pin icon in the top right - when the current location is available, it is enabled. When clicked, it places the Point pin on the center of the current location.


library/forms/lib/fieldRegistry/fields/MapField/MapWrapper.tsx
```diff
@@ -416,6 +417,27 @@ function MapWrapper(props: MapProps) {
               center={props.center}
               extent={featuresExtent}
               zoom={props.zoom}
+              additionalControls={{
+                // Add use current location control for points
+                useCurrentLocation:
```
stevecassidy: right, so this is the code that is being called in MapComponent to save the point - too hard to find that out!

PR #1879, opened 2026-01-20
fix: offline record fetching

PeterBaker0:
Addresses https://jira.csiro.au/browse/BSS-1079 

Simple fix - also fixed bug with the on click logic.


stevecassidy (approved):
Thanks for the catch.

PR #1881, opened 2026-01-20
feat: map help text, grayed out save

PeterBaker0:
Adds help text for each geometry type explaining the action needed. 

Greys out the save option when the user a) had existing features and b) hasn't finished drawing a point (or selected using the new select current location).


stevecassidy (approved):
Very neat.

PR #1893, opened 2026-01-28
feat: speech to text fields

PeterBaker0:
Utilises the cap-go speech recognition capacitor plugin [https://github.com/Cap-go/capacitor-speech-recognition](https://github.com/Cap-go/capacitor-speech-recognition) . This is licensed under MPL and includes punctuation, silence allowed parameters. Both are useful for us - allowing entry of addresses for example. 

@stevecassidy I don't think there are any licensing issues with using MPL licenses under apache - so long as we don't relicense the binary/files themselves. We could fall back to https://github.com/[capacitor-community/speech-recognition](https://github.com/capacitor-community/speech-recognition) but will lose some features. 

Adds a `useSpeechToText.tsx` hook - used in the BaseTextField component in `forms` - requires some additions to the iOS manifest, and some additional packages. 

Seems to work reasonably reliably - tested on real Android device. Works offline since I've enabled offline voice to text on my device. 

Can run in either append or replace mode. 

Updated designer to expose this new setting for text fields.

### DIFF
--- a/library/forms/lib/fieldRegistry/fields/wrappers/BaseMuiTextField.tsx
+++ b/library/forms/lib/fieldRegistry/fields/wrappers/BaseMuiTextField.tsx
@@ -84,7 +84,7 @@ export const BaseMuiTextField: React.FC<BaseMuiTextFieldProps> = props => {
           appendMode: effectiveAppendMode,
           partialResults: true,
           addPunctuation: true,
-          debugMode: true,
+          debugMode: false,
           onResult: text => {
             setFieldData(text);
           },


stevecassidy:
Testing on IOS I get an error where recognition doesn't stop when I finish speaking and I need to touch the microphone icon again to stop.  If I do that after even a short delay, the input so far is deleted. If I do it very quickly it is kept. 

Here is a trace of the case where the input is deleted:

image

stevecassidy (approved):
Happy with this, nice feature!

Issue #1865, opened 2026-01-15
Add a new Admin role for system managers


stevecassidy:
We need a high level role for managers of a deployment who can manage teams and users but don't have access to any notebooks and data.  

In general, system administrators should not be able to see user data which might be confidential.  We should minimise the use of the GENERAL_ADMIN role for emergency use only.  

Add OPERATIONS_ADMIN role for this and give them enough permissions to create/modify/remove teams and users.   We can fine tune the permissions as we need to.

PR #1894, opened 2026-01-28
Add OPERATIONS_ADMIN global role

stevecassidy:
# Add OPERATIONS_ADMIN global role

## Ticket

#1865

## Description

We need a high level role for managers of a deployment who can manage teams and users but don't have access to any notebooks and data.

In general, system administrators should not be able to see user data which might be confidential. We should minimise the use of the GENERAL_ADMIN role for emergency use only.

## Proposed Changes

- Adds OPERATIONS_ADMIN role for this and give them enough permissions to create/modify/remove teams and users. We can fine tune the permissions as we need to.
- Some changes to the way that permissions are checked in the `web` app to make it more fine grained - previous assumption was based on 'admin' being able to do everything. 
- Makes the Teams page the default page in `web` and implements this via a redirect rather than just including the component.   Makes more sense for this to be primary if we assume that all users will be team members. 
- Removes display of notebooks and templates for users who can't access them (OPERATIONS_ADMIN).
- Does not add the current user as admin for a created team, assuming that teams are generally being made for someone else to manage and that OPERATIONS_ADMIN can manage any team
- Uses pretty names for roles in `web`

## How to Test

Create a user and assign the new role to them. This user should be able to manage user accounts and teams but not see any notebooks or data.

## Checklist

### DIFF
--- a/web/src/components/tables/cells/template-cell.tsx
+++ b/web/src/components/tables/cells/template-cell.tsx
@@ -20,7 +20,11 @@ export const TemplateCellComponent = ({
     return <p>Unauthenticated</p>;
   }
 
-  const {data: template, isLoading, isError} = useGetTemplate({user, templateId});
+  const {
+    data: template,
+    isLoading,
+    isError,
+  } = useGetTemplate({user, templateId});
 
   return isLoading ? (
     <Skeleton className="h-5 w-24" />


web/src/components/side-bar/app-sidebar.tsx
```diff
@@ -39,14 +39,11 @@ export function AppSidebar({...props}: React.ComponentProps<typeof Sidebar>) {
...
 
-  const {data: projects} = useGetProjects(user);
-  const {data: teams} = useGetTeams(user);
-  const {data: templates} = useGetTemplates(user);
-
   const topSectionNavItems: NavItem[] = [];
   const bottomSectionNavItems: NavItem[] = [];
 
   if (canSeeProjects) {
+    const {data: projects} = useGetProjects(user);
```
PeterBaker0: Please pass through an enabled flag for the query instead- this is a violation of react rules - custom hooks can not be run conditionally (as can change hook order between re-renders).

stevecassidy: added enabled param to all relevant hooks

PR #1897, opened 2026-01-29
feat: SAML extension of existing auth provider workflow

PeterBaker0:
Implements a comprehensive SAML integration based on the pattern for OIDC/google. Makes some minor non-breaking adjustments along the way. Have tested all the way through to it appearing in the list and metadata endpoint - but haven't been able to check auth return yet.

### DIFF
--- a/api/test/authConfig.test.ts
+++ b/api/test/authConfig.test.ts
@@ -21,6 +21,10 @@
 import {expect} from 'chai';
 import sinon from 'sinon';
 import {readAuthProviderConfigFromEnv} from '../src/auth/strategies/applyStrategies';
+import {
+  GoogleAuthProviderConfig,
+  OIDCAuthProviderConfig,
+} from '../src/auth/strategies/strategyTypes';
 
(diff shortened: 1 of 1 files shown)


api/src/auth/authRoutes.ts
```diff
@@ -865,7 +866,26 @@ export function addAuthRoutes(
...
-    const handlerDetails = socialProviders[provider];
+    const handlerDetails = socialProviders[provider].config;
+    const handlerStrategy = socialProviders[provider].strategy;
+
+    if (handlerDetails.type === 'saml') {
+      // Get the strategy instance from passport
+      const strategy = handlerStrategy as SamlStrategy;
+
+      // Metadata endpoint
+      app.get(`/auth/${provider}/metadata`, (req, res) => {
```
stevecassidy: Why are you sending a request to ourselves here? This route doesn't seem to exist. Is it supposed to be a call to the IdP?  I would be happier if all SAML stuff was in the provider module rather than needing special casing here.

PR #1912, opened 2026-02-04
feat: integration for bugsnag into /web

PeterBaker0:
Adds bugsnag integration. Catches render level errors in the existing error component. Tried using the Bugsnag error component approach, with and without tanstack router, and it seems that tanstack was more reliable. Reports error on mount of this error component. Tested with a test bug report component and works fine.


stevecassidy (approved):
Thanks for this.

PR #1913, opened 2026-02-04
feat: logging solution for forms

PeterBaker0:
Implements singleton generic logging service for forms. 

info, warn, debug follow the console.* equivalent. Error is more explicit about providing the actual Error type object, with other context kvs provided through an object {}. 

The app then implements a bugsnag implementation, and registers it. 

I got a long way through doing a context provider approach, but so many errors were outside of the scope of react components, so needed prop drilling to send through the logger, which was way more painful than worth for the slight benefit of re-renders for logging services, or better component tree encapsulation (e.g. different loggers for different component trees). So I went with a singleton service pattern.

### DIFF
--- a/app/src/logging.tsx
+++ b/app/src/logging.tsx
@@ -191,7 +191,9 @@ const createBugsnagFormLogger = (
         });
       },
       warn: (message, ...args) => {
-        console.warn('[FormLogger] Warning:', message, ...args);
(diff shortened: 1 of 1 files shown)


stevecassidy (approved):
Nice solution. Could we also use this from data-model? Should be able to import the logger from there as well?

Issue #1925, opened 2026-02-09
Make local login configurable in deployment


stevecassidy:
We need to be able to deploy without the option of local login - ie. SSO only.

Issue #1868, opened 2026-01-16
Should not be able to generate password reset for SSO user


stevecassidy:
Currently we show the password reset link for all users in the dashboard user table. For an SSO user, it doesn't make sense to do a password reset since they don't have one. 

Similarly, for an SSO user, don't generate a password reset email.  Maybe send an email saying to use whatever system to change their password.

PR #1927, opened 2026-02-09
Allow disabling of local login, no password reset for SSO users

stevecassidy:
# Allow disabling of local login, no password reset for SSO users

## Ticket

#1868 #1925

## Description

We need the option of disabling local login for an SSO only deployment. 

Password reset should only be an option for a locally registered user, not an SSO user.

## Proposed Changes

New configuration setting DISABLE_LOCAL_LOGIN defaults to false.  Internally we use LOCAL_LOGIN_ENABLED since most of the tests include stuff if it is enabled.   

`disableLocalLogin` setting added to CDK configuration for conductor.

Removes option to generate password reset for SSO only users.

## How to Test

Configure without local login, view the login page.

## Checklist

### DIFF
--- a/api/views/register.handlebars
+++ b/api/views/register.handlebars
@@ -1,5 +1,5 @@
 <div class="auth-header">
-    <h2>Registration</h2>
+    <h2>Create Account</h2>
     <p>Already have an account? <a href="{{loginURL}}" class="auth-link">Sign in</a></p>
 </div>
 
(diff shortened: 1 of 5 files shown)


PeterBaker0 (changes requested):
Looking good. 

However I think the quite significant changes to the headings/layout of the login/registration pages warrant further discussion, and a separate PR.

api/views/register.handlebars
```diff
@@ -1,5 +1,5 @@
 <div class="auth-header">
-    <h2>Create account</h2>
+    <h2>Registration</h2>
```
PeterBaker0: Please move to other PR with auth page refactors

PeterBaker0 (approved):
After comments are addressed - approved - functionality looks good thanks for this option. I don't think we can enable this for our use case as we don't have full SSO coverage, but it's a nice feature for organisations to avoid having to deal with user/pass auth.

PR #1956, opened 2026-02-26
Feat: Use bugsnag error logging in api in place of console.error

stevecassidy:
# Use bugsnag error logging in api in place of console.error

## JIRA Ticket

None

## Description

Replaces most uses of console.error with a call to logError which uses Bugsnag if configured so that we get error reports via bugsnag in production.

## Checklist


stevecassidy:
Also note in passing that since we use thown errors to report things back to the user via API responses we are logging some errors that don't need logging like expired tokens - maybe selectively use explicit return values for these rather than errors so that we minimise logging to actual problems.

PR #1971, opened 2026-03-11
build: website static build for docs

PeterBaker0:
Adds to CDK config to build out a docs.* website for the user docs.

**Theming** needs work, other themes don't currently work as they don't have the screenshots. It'd be nicer if themes with missing images would use the default theme as fallback.


stevecassidy (approved):
Thanks for this.

PR #1975, opened 2026-03-12
feat: adds an additional button click step for email verification

PeterBaker0:
Auto email scanners were clicking the email verification link - this resulted in the user seeing the email, clicking it, then seeing error the code was already consumed. Adds a step to manually click the button, which should resolve auto scanners consuming the code, improving the user experience.


stevecassidy (approved):
Thanks.

PR #1978, opened 2026-03-12
feat: adds grace period for login banner

PeterBaker0:
### Problem
Logged-out warning banner appears immediately on app launch/return, even while a background token refresh is succeeding, confusing users who briefly see “You are currently logged out”.

### Cause
`TolerantPrivateRoute` shows the `OfflineLoginBanner` whenever `!isAuthenticated && isToken && isOnline`, with no grace period around app initialisation or token refresh.

### Solution
Introduce a configurable login-banner grace period (`VITE_LOGIN_BANNER_GRACE_MS`) and update `TolerantPrivateRoute` to delay showing the banner until this grace period has elapsed, preventing logged-out warnings while tokens are being refreshed successfully.


stevecassidy (approved):
Nice fix.

PR #1989, opened 2026-03-17
feat: map overview improvements

PeterBaker0:
### Summary

Improves the notebook overview map with better UX, touch support, and richer feature styling and popover content.

### Changes

**1. Click / tap behavior**
- **Touch support:** Replaced map `click` with `pointerdown`/`pointerup` so taps work on Android (avoids pan swallowing the gesture).
- **Popover placement:** Popover is anchored to the tap position (`anchorPosition`) instead of the map container so it opens in the right place on first open.
- **Backdrop close:** Ignores `backdropClick` for 400 ms after opening so the same tap doesn’t immediately close the popover on touch.
- **Button guard:** “View record” is disabled for 400 ms after the popover opens so the opening tap doesn’t trigger navigation.

**2. Popover content**
- New `SelectedRecordPopoverContent` shows hydrated record details (HRID, form label, created date/author) via React Query (`dataEngine.hydrated.getHydratedRecord`) and a “View record” button linking to the record view (not edit).
- Loading and error states with retry/“Open record” fallback.

**3. Map styling**
- **Form-type colors:** Features colored by `form_id` using a fixed, colorblind-friendly palette (`FORM_TYPE_COLORS`); polygons with semi-transparent fill, points with stroke.
- **Selection highlight:** Selected feature gets a distinct style (larger, white stroke, 50% opacity fill).

**4. North-orientation control**
- “Reset to north” compass control (OpenLayers `Control`) resets rotation to 0

### DIFF
--- a/app/src/gui/components/notebook/OverviewMap.tsx
+++ b/app/src/gui/components/notebook/OverviewMap.tsx
@@ -225,13 +225,7 @@ const SelectedRecordPopoverContent = ({
   });
 
   return (
-    <Card
-      variant="outlined"
-      sx={{
-        minWidth: 'min(260px, 100%)',
-        maxWidth: 'min(320px, 100%)',
-      }}
-    >
+    <Card variant="outlined" sx={{minWidth: 260, maxWidth: 320}}>
       <CardContent sx={{'&:last-child': {pb: 2}}}>
         <Typography
           variant="subtitle1"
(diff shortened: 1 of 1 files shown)


ranisa-gupta16 (changes requested):
tested this locally, definitely much better than before. I have listed my observations before, it's up to us based on priority whether and how to action them but would like to hear your thoughts.

When two records are very close to each other the dots overlap and one of them almost disappears behind the other one, also there's current location point which sits there. there's no visual indicator that there are multiple records in the same spot, a user might not realise there's a second record underneath. might be worth covering in future but came up to my observation


the "view record" button looks clean and doesn't feel cluttered, 

The form colours are assigned by alphabetical sort of the form ids ? , so adding a new form type can change the colour of an existing one?. A user who's finds out that "green = Structure/Site" might find it's changed after a sync. ? that was confusing to me. 

A stable colour mapping would be safer.
for example : according to my thouhgt process
Blue = Building
Red = Infrastructure
Green = Site
Purple = Vehicle
with the 4 form types: Building, Infrastructure, Site, Vehicle. Alphabetically sorted that's Building → Infrastructure → Site → Vehicle, getting colours blue, red, green, purple in that order.
Now ifv someone adds "Fence" as a new form type to this notebook, the sort becomes Building → Fence → Infrastructure → Site → Vehicle. Everything from Infrastructure onwards shifts — Infrastructure which was red is now green, Site which was green is now purple, Vehicle which was purple is now something else entirely.

A field worker who hass been using this survey for weeks and knows "red dot = Infrastructure damage, go check it" now sees a different colour on that same dot after the next sync, with zero indication anything changed. On a damage assessment survey where you're quickly scanning the map to prioritise which sites to visit first ? 


the 400 ms value appears in three different places in the touch-handling code — the constant POPOVER_BUTTON_GUARD_MS is already defined, the close handler should reuse it rather than hardcoding the literal.


The PR description mentions hiding it at 0 rotation (Google Maps style) but the implementation shows it permanently. Not a bug, just the stated intent isn't there yet — worth flagging so it doesn't get forgotten. :-)

minor - the popover card has a right-side margin issue on smaller screens. The card clips or sits too close to the edge, making it feel cramped and slightly off. Worth adding a max-width with some horizontal margin so it doesn't butt up against the screen edge on phones.
Image :-)

code wise it all looks good.
may be add how to test in future for similar 
pr's to test cases according to developer's point of view. :-) thanks for all the work

Issue #1997, opened 2026-03-18
Don't throw errors when tokens expire


stevecassidy:
Seeing a very large number of errors reported on bugsnag for expired access tokens because we throw an error in this case rather than just returning a response.   Catch this and return an explicit response so we don't get the Bugsnag report.

PR #1998, opened 2026-03-18
Don't throw errors when tokens expire

stevecassidy:
# Don't throw errors when tokens expire

## Ticket

#1997

## Description

Seeing a very large number of errors reported on bugsnag for expired access tokens because we throw an error in this case rather than just returning a response. 

## Proposed Changes

Rather than throwing an error we just return the error response directly.

## How to Test

Tested by the API tests, behaviour is as before but no exceptions are reported.

## Checklist


PeterBaker0 (approved):
Looks fine

PR #1999, opened 2026-03-19
feat: data access APIs

PeterBaker0:
### Summary

Adds a **stateless REST API** for notebook record data under `/api/notebooks/:id/records`, as an alternative to full Pouch/Couch sync. Requests use Bearer JWT auth; **project- and record-level permissions** match existing rules (“my” vs “all” records), implemented via new `recordAuth` helpers.

### What’s included

- **CRUD-style operations**: list, get, create, patch/update, delete, plus **fork revision** (`POST …/records/:recordId/revisions`) aligned with the data model’s revision/head behaviour.
- **`@faims3/data-model`**: shared Zod schemas/types, engine support, and API surface used by the routes.
- **Tests**: `records.test.ts`, `recordAuth.test.ts`, fixtures, and small updates to existing API tests.
- **Docs**: `RecordsCRUDApi.md` (linked from the developer index) describing endpoints, auth, model, and an explicit **disclaimer** that attachment upload/download is not fully supported on this REST surface (metadata only; use export/sync paths for binaries).
- **App**: minor related tweaks in a few GUI files (e.g. app bar / notebook / map components).

### Notes for reviewers

- Existing export/list routes under `/:id/records/...` remain; the new router is mounted at `/:id/records` alongside them—worth confirming route ordering and any overlap with legacy paths.


api/src/api/records.ts
```diff
@@ -0,0 +1,421 @@
...
+    } catch (err) {
+      mapDataModelError(err);
+    }
+  }
+);
+
+/**
+ * GET /api/notebooks/:id/records - List records (permission-filtered)
+ */
+recordsRouter.get(
```
stevecassidy: Is this backward compatible with the previous API - this was the only option before and we have clients depending on it.   It looks like it should be but worth verifying.

api/src/api/records.ts
```diff
@@ -0,0 +1,421 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Stateless CRUD API for record data under /api/notebooks/:id/records
+ * (including POST …/records/:recordId/revisions to fork a revision).
+ */
+
```
stevecassidy: These routes all need adding to the api documentation (swagger).

api/test/records.test.ts
```diff
@@ -0,0 +1,931 @@
...
+registerClient(callbackObject);
+
+describe('Records CRUD API', () => {
+  beforeEach(beforeApiTests);
+
+  describe('list records', () => {
+    it('returns permission-filtered list after restore', async () => {
+      await withRecordsBackup(async projectId => {
+        const res = await requestAuthAndType(
+          request(app).get(recordsBasePath(projectId))
```
stevecassidy: I find it harder to read the tests with this abstraction - if I'm looking for where we test a particular route, I'll look for the route in the tests. This hides that.

Issue #1996, opened 2026-03-18
[BUG] social sign-in provider login callback throws 500 server crash


PeterBaker0:
Social sign-in provider, specifically Google, throws a 500 error due to an uncaught exception at the auth return route post social sign-in. Bugsnag does catch this uncalled exception and reports it. Below is information on the error. It appears that the route is trying to validate an invite code; however, this was during a regular login, not registration, so invite codes should be irrelevant. 

in `auth/helpers.ts` 

```typescript
 if (!isValid) {
            throw new Error(`Invite was correct, but is longer valid. Reason: ${reason !== null && reason !== void 0 ? reason : 'not provided.'}.`);
  }
```

This errors and is not caught.

PR #2024, opened 2026-03-29
Fix: stale invites interfere with SSO login

stevecassidy:
# Fix: stale invites interfere with SSO login

## Ticket

#1996

## Description

Some logins via SSO were failing due to the presence of stale invites in the user session. 

## Proposed Changes

There are two parts to this problem:  

1.  We were not resetting the session when we redirect after a successful login so any inviteId in the session was preserved until next login (cookie also lasted for a year).  

2.  There was some inconsistency in when invites were checked

The fix includes a refactor of the auth flow to move the invite checking code into a new helper function `completePostAuth` which optionaly applies the invite and then redirects after cleaning up the session.

Make sure that this helper is called in all routes through auth - noting that we can have an invite in both the login and register routes (login for existing users who are using an invite to gain new permissions).  

## How to Test

Login and Register using an invite using SSO and local login.

## Additional Information

The refactor was performed by Claude which helped work through the somewhat messy code and identify the inconsistent application of invites.   Tests were also updated by Claude to match the newly refactored helper functions. 

## Checklist

### DIFF
--- a/docs/developer/docs/source/index.rst
+++ b/docs/developer/docs/source/index.rst
@@ -7,8 +7,8 @@ FAIMS3 Developer Documentation
 
    markdown/Drafts.md
    markdown/CouchMigrations.md
-   markdown/NotebookMigrations.md
    markdown/PermissionModel.md
+   markdown/NotebookMigrations.md
    markdown/Long-lived-tokens.md
    markdown/TokenManagement.md
    markdown/Configuration.md
--- a/docs/developer/docs/source/markdown/Authorisation/PassportAuth.md
+++ b/docs/developer/docs/source/markdown/Authorisation/PassportAuth.md
@@ -20,9 +20,11 @@ The system is configured in `expressSetup.ts` which initializes:
 2. Authentication handlebar **pages** via `addAuthPages`
 3. Authentication API **routes** via `addAuthRoutes`
 
-Local password authentication is always available and if configured, one or
-more identity providers can also be used. There is a custom implementation for
-Google authentication and a more general OIDC based provider.  
+Configuration settings allow enabling of either local email/password authentication or
+one or more federated identity providers. There is a custom implementation for
+Google authentication, a more general OIDC based provider and a custom SAML provider
+implementation.  Further SSO protocols could be supported via their passport
+implementations.
 
 The available
 providers are configured via environment variables using a pattern of variable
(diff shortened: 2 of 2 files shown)


api/src/auth/helpers.ts
```diff
@@ -523,139 +624,77 @@ export async function ssoVerify({
...
     if (!(strategyId in matchedSingleUser.profiles)) {
       matchedSingleUser.profiles[strategyId] = profile;
       await saveCouchUser(matchedSingleUser);
     }
 
-    // upgrade user and return login success - invite to be processed later if
-    // at all
-    return done(
-      null,
-      await upgradeCouchUserToExpressUser({dbUser: matchedSingleUser})
```
PeterBaker0: stripped this again, intentional?

api/src/auth/helpers.ts
```diff
@@ -523,139 +624,77 @@ export async function ssoVerify({
...
-    // NOTE: This is the situation where you are trying to 'register' a new
-    // account but one already exists with SSO with matching email - we
-    // decide here to instead log them in - upgrading the potentially
-    // unconnected account
-
-    // We have precisely one matching email address, let's ensure that this
-    // account has the linked SSO profile, then return it (We can safely assert
-    // non-null here due to our previous filtering)
-
-    // Firstly - ensure they have the SSO profile linked
```
PeterBaker0: lots of explanation here - this logic is tricky and worth keeping this explanatory (hand written) comments

api/src/auth/helpers.ts
```diff
@@ -523,139 +624,77 @@ export async function ssoVerify({
...
 
-    await saveCouchUser(matchedSingleUser);
+    // Persist profile/email changes (invite consumption is done by completePostAuth)
+    await saveCouchUser(targetUser);
 
-    // upgrade user and return login success - invite to be processed later if
-    // at all
-    return done(
-      null,
-      await upgradeCouchUserToExpressUser({dbUser: matchedSingleUser})
```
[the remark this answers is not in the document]
PeterBaker0: I see later on that this is because post auth hook is called later - okay

PeterBaker0 (approved):
Some comments to restore (which claude wouldn't delete them!), some stuff to double check ,some tests have been removed which would preferably be replaced with new behaviour assertions or other replacements/explanation. Otherwise looks sensible. 

Would it be possible while you're working with claude to have it trawl all the auth code and a) ensure we have a sufficient technical documentation for it b) build a diagram? Mermaid JS would be good for this. Can be another task if need be.

Issue #1940, opened 2026-02-17
Remove mention of publishButtonBehaviour setting in designer


stevecassidy:
publishButtonBehaviour is no longer used and could be removed from the codebase although we need to allow for this being in old notebooks.  Remove it from the Form Settings in designer.

PR #2030, opened 2026-03-31
Fix: remove publishButtonBehaviour setting

stevecassidy:
# Fix: remove publishButtonBehaviour setting

## Ticket

#1940

## Description

We no longer use the publishButtonBehaviour setting but it is included in the designer.

## Proposed Changes

Removes all mention of this setting and tests that include it.

Fixes a minor bug in the uiSpec reducer that was adding duplicates to visible_types.

## How to Test

Nothing should have changed except that this setting is no longer shown in the designer form settings tab.

## Checklist


PeterBaker0 (approved):
thanks 👍

PR #2041, opened 2026-04-08
feat: optional templateFunction on field registry for templating

PeterBaker0:
**Summary**  
Mustache expansion for templated string fields was passing raw values into the template context. For JSON-shaped fields (notably **Address**), that produced useless output (e.g. `[object Object]`) instead of a readable string.

**Changes**  
- Extended **`FieldInfo`** with an optional **`templateFunction?: (value: unknown) => string`**.  
- **`recomputeDerivedFields` / `renderTemplate`** now resolve each field via **`getFieldInfo`** and, when **`templateFunction`** is defined, use its return value in the Mustache context; otherwise behaviour is unchanged. Record context keys (`_CREATOR_NAME`, `_CREATED_TIME`) are not transformed. Failures in **`templateFunction`** are caught, log a warning, and substitute an empty string.  
- **`AddressField`** implements **`templateFunction`** using **`display_name`**, with **`manuallyEnteredAddress`** as a fallback (aligned with view rendering).  
- Added **Vitest** coverage in **`templatedFields.test.ts`**.

**Out of scope (follow-up)**  
Other structured field types (map, GPS point, related records, attachments) can adopt the same hook later; no changes in this PR.

**Testing**  
- `pnpm test` in `library/forms` (includes new `templatedFields` tests).
- create a templated string from an address field and ensure it renders properly

### DIFF
--- a/library/forms/lib/fieldRegistry/fields/AddressField/index.tsx
+++ b/library/forms/lib/fieldRegistry/fields/AddressField/index.tsx
@@ -586,6 +586,23 @@ const valueSchema = (props: AddressFieldProps) => {
   return AddressValueNullableSchema;
 };
 
+function addressValueForTemplate(value: unknown): string {
+  if (value == null || typeof value !== 'object') {
+    return '';
+  }
+  const v = value as {
+    display_name?: unknown;
+    manuallyEnteredAddress?: unknown;
+  };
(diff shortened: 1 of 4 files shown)


stevecassidy (approved):
Looks good.

PR #2044, opened 2026-04-09
feat: overhaul of resource deletion and archiving lifecycle including projects, records, templates and users

PeterBaker0:
### Summary

This branch delivers the **deletion and archiving** epic: **templates** and **surveys/notebooks** can be **archived** and **permanently removed** with clearer permissions and data cleanup, plus a **web Archive** area. It also adds **disabled user accounts** (soft-off) so blocked users cannot authenticate while keeping historical data and roles on disk until you change them.

### Surveys / notebooks (projects)

Implements a complete survey lifecycle (closed, archived, deleted), including forcing remote deletion when opt-in allowed.

### Templates

Pulls archived flag out of the notebook JSON into a top level field (requiring migration). Applies this consistently

### Data model, permissions, and migrations

Permission model updated accordingly with new actions. Migrations where needed.

### User accounts (disable / re-enable)

Disabling a user moves them to the archive and stops any login related activity - 401 everything.

### Web archive mode

Special view where you can see archived entries then restore/delete.

### How to verify

- run all tests
- run migrations
- try
  - archiving a template
  - archiving template with attached surveys
  - deleting a template
  - closing/opening a survey when activated/not activated
  - archiving a survey when activated/not activated with remote deletion allowed and disabled
  - deleting a survey when activated/not activated with remote deletion allowed and disabled
  - deleting records from info panel
  - deleting records from the related record selector
  - exporting records which have deleted attached records (should not be present in export)


api/src/api/notebooks.ts
```diff
@@ -662,10 +724,35 @@ api.get(
...
       const filenames: string[] = [];
       // Process any file fields to give the file name in the zip download
-      records.forEach(record => {
+      for (const record of records) {
+        if (record.data) {
+          try {
+            const fields = getNotebookFieldTypes({
+              uiSpecification,
+              viewID: record.type,
+            });
```
PeterBaker0: I'm thinking this is probably quite inefficient

api/src/api/notebooks.ts
```diff
@@ -662,10 +724,35 @@ api.get(
...
+            const dataCopy = {...record.data};
+            await stripDeletedRelatedRefsFromRecordData({
+              fields,
+              data: dataCopy,
+              dataDb,
+              uiSpecification,
+            });
+            record.data = dataCopy;
```
PeterBaker0: do we actually need to make a copy here

api/src/couchdb/export/csvExport.ts
```diff
@@ -289,9 +292,17 @@ export const appendAllCSVsToArchive = async ({
         const hrid = record.hrid || record.record_id;
         const row = [...generateRecordPrefixInformation(record)];
 
+        const rowData = {...record.data} as Record<string, unknown>;
+        await stripDeletedRelatedRefsFromRecordData({
+          fields: viewState.fields,
+          data: rowData,
+          dataDb: dataDb as DatabaseInterface<DataDocument>,
+          uiSpecification,
+        });
```
PeterBaker0, line 297: again is it actuall y necessary to make copy here

api/src/couchdb/export/stripDeletedRelatedRefs.ts
```diff
@@ -0,0 +1,126 @@
...
+/** Couch data DB as returned by {@link getDataDb} or test fakes typed as {@link DataDocument}. */
+export type StripRelatedRefsDataDb =
+  | DataDbType
+  | DatabaseInterface<DataDocument>;
+
+const RELATIONSHIP_COMPONENT = 'faims-custom::RelatedRecordSelector';
+
+function componentKey(namespace: string, name: string): string {
+  return namespace ? `${namespace}::${name}` : name;
+}
```
PeterBaker0: this helper function would exist somewhere

api/src/couchdb/export/stripDeletedRelatedRefs.ts
```diff
@@ -0,0 +1,126 @@
...
+    const entries = Array.isArray(normalized) ? normalized : [normalized];
+
+    const kept: typeof entries = [];
+    for (const item of entries) {
+      const rid = item.record_id;
+      try {
+        const h = await engine.hydrated.getHydratedRecord({
+          recordId: rid,
+          config: {conflictBehaviour: 'pickLast'},
+        });
```
PeterBaker0: this is potentially not efficient- as each record then re-hydrates all related records - I think we should just check the deleted property more directly - just fetching the record head revision should be sufficient since we use no other information.

api/src/couchdb/notebooks.ts
```diff
@@ -126,6 +126,73 @@ export const putProjectDoc = async (doc: ProjectDocument) => {
...
+
+/**
+ * Clears `templateId` on all project documents that reference the template.
+ * Used when permanently deleting a template so surveys do not keep stale
+ * references.
+ */
+/**
+ * Drops the per-field metadata doc for `template_id` if it exists (see
+ * `writeProjectMetadata`), so notebook details no longer cite the template.
+ */
```
PeterBaker0, line 156: double metadata

app/src/context/slices/projectSlice.ts
```diff
@@ -32,6 +32,89 @@ import {databaseService} from './helpers/databaseService';
...
+      resolveNext = r;
+    });
+    await prev.catch(() => {});
+    try {
+      return await fn();
+    } finally {
+      resolveNext();
+    }
+  };
+}
```
PeterBaker0, line 80: add some documentation/explanation around this logic

app/src/context/slices/projectSlice.ts
```diff
@@ -1767,9 +1894,87 @@ export const initialiseProjects = createAsyncThunk<void, {serverId: string}>(
...
+        appDispatch(detachProjectRetainLocalData({serverId, projectId}));
+      }
+    }
+
+    if (needsAbsentConfirmationRetry) {
+      scheduleAbsentDirectoryRetry(serverId, appDispatch);
+    } else {
+      clearAbsentDirectoryRetry(serverId);
+    }
+    });
```
PeterBaker0: add some comments

library/forms/lib/fieldRegistry/fields/RelatedRecord/index.tsx
```diff
@@ -597,12 +640,138 @@ const FullRelatedRecordField = (props: FullRelatedRecordFieldProps) => {
...
+        userId: props.config.user,
+      });
+      await queryClient.invalidateQueries({queryKey: ['related-hydration']});
+      setDeleteTarget(null);
+    } catch (e) {
+      setDeleteError(
+        e instanceof Error ? e.message : 'Could not delete this record.'
+      );
+    }
+  };
```
PeterBaker0, line 744: add comments and doc strings

library/forms/lib/fieldRegistry/fields/RelatedRecord/index.tsx
```diff
@@ -660,23 +829,123 @@ const FullRelatedRecordField = (props: FullRelatedRecordFieldProps) => {
...
+      >
+        <DialogTitle>Delete related record?</DialogTitle>
+        <DialogContent>
+          {deleteError && (
+            <Alert severity="error" sx={{mb: 1}}>
+              {deleteError}
+            </Alert>
+          )}
+          <Typography variant="body2">
+            This deletes the linked record (soft delete), the same as deleting it
```
PeterBaker0: wording updates

PeterBaker0 (commented):
Self-review - will address then re-request from Steve

api/src/api/notebooks.ts
PeterBaker0: Did some investigation and I think its performance implications are minimal, and is a reasonable defensive choice since we wouldn't really expect this function in its contract to mutate the incoming data

web/src/project-archive/project-lifecycle-copy.ts
```diff
@@ -0,0 +1,51 @@
...
+export function getArchiveProjectDialogBody(): string {
+  if (FORCE_REMOTE_DELETION === 'allow') {
+    return [
+      `Archiving hides this ${NOTEBOOK_NAME_CAPITALIZED} from normal lists.`,
+      `Be absolutely sure every field user has uploaded their records before you archive. After users open the app, it will wipe this ${NOTEBOOK_NAME} from their device. Anything that was not uploaded will be lost forever and cannot be recovered.`,
+    ].join(' ');
+  }
+  return [
+    `Archiving hides this ${NOTEBOOK_NAME_CAPITALIZED} from normal lists.`,
+    `People should sync first so the server has their latest work. After users open the app, it will deactivate this ${NOTEBOOK_NAME} from their device. While data may be recoverable on some devices, we do not recommend relying on this.`,
```
stevecassidy: 'Users should sync...' 'After they open the app,...'.   By 'deactivate' here I think you mean that it will delete all data - if so it should say that.

stevecassidy (commented):
Wording in ProjectStatusDialog (from an earlier PR): "Closed: Project is read-only. Users will not be able to activate this project for data collection."  Maybe also note that users who have already activated will still be able to browse data on their device/browser.

web/src/project-archive/project-lifecycle-copy.ts
```diff
@@ -0,0 +1,51 @@
...
+  ].join(' ');
+}
+
+/** Short line for the project actions card (archive list item). */
+export function getArchiveProjectActionsDescription(): string {
+  if (FORCE_REMOTE_DELETION === 'allow') {
+    return `Hide this ${NOTEBOOK_NAME} from normal lists. After users open the app, their devices may wipe this ${NOTEBOOK_NAME}. Records that were never uploaded can be lost forever.`;
```
stevecassidy: What is a 'normal list'?

web/src/project-archive/project-lifecycle-copy.ts
stevecassidy: Ah, I see the conditional now, so in this case it isn't forcing deletion.   Now there is a problem - the notebook is not visible on my device but the database is still present and there is no way that I can remove it.   Need a mechanism for a user to clear out their device.

api/public/swagger.json
```diff
@@ -93,17 +93,82 @@
         }
       }
     },
+    "/directory/": {
+      "get": {
+        "summary": "List surveys visible to the caller (directory)",
```
stevecassidy: Hmm, we have a survey/notebook issue in this document. Ideally we find a way to customise it but we should at least be consistent - most other endpoints use 'notebook'.

api/src/api/notebooks.ts
```diff
@@ -588,6 +613,43 @@ api.put(
   }
 );
 
+/** Archive or un-archive a survey (`archive: true` sets {@link ProjectStatus.ARCHIVED}). */
```
stevecassidy: isn't it `{"archive": true}` as in a JSON body, also note `false` will un-archive.

api/src/api/notebooks.ts
```diff
@@ -588,6 +613,43 @@ api.put(
...
+  }),
+  async ({params: {id}, body: {archive}}, res) => {
+    await setNotebookArchived({projectId: id, archive});
+    res.sendStatus(200);
+  }
+);
+
+/** Restore an archived survey to closed (not open). */
+api.post(
+  '/:id/restore',
```
stevecassidy: But doesn't PUT to /:id/archive with {"archive": false} do this? We shouldn't need another endpoint here.

api/src/api/notebooks.ts
stevecassidy: Ah, but we have PUT to /:id/status above which could also do archiving since it is just changing the value of the status flag for this project which can be 'open', 'closed' or 'archived'.  I could use that endpoint to set status to 'ARCHIVED' and it would be allowed but would not archive the project.  Suggest we retain the single /:id/status endpoint and implement the right actions there depending on the status requested.

api/src/api/notebooks.ts
```diff
@@ -661,11 +723,58 @@ api.get(
     });
     if (records) {
       const filenames: string[] = [];
+      const viewIdsNeedingFieldTypes = new Set<string>();
```
stevecassidy: What is this from? Feels like it is from another fix but it's not currently in main.

api/src/auth/strategies/localStrategy.ts
```diff
@@ -74,6 +74,13 @@ export const verifyUserCredentials = async ({
     };
   }
 
+  if (isPeopleUserAccountDisabled(dbUser)) {
+    return {
+      success: false,
+      error: ambiguousErrorMessage,
```
stevecassidy: I wonder if we should explicitly say this account is disabled - doesn't leak any info to attackers except that this account did exist before.  Note that the error that I see when I try to login to a disabled account is "Login failed"

web/src/components/tables/users.tsx
```diff
@@ -156,6 +157,17 @@ export const useUsersColumns = ({
         <div className="flex justify-center items-center">Reset Password</div>
       ),
     },
+    {
```
stevecassidy, line 159: Not sure if we should not show this for SSO only users (as we do for password reset).  For SSO the account removal will be done on the IDP but they may then want to come in and clean up by removing accounts here.   Since we don't prevent disabled SSO users from logging in it might be best to not offer this for them.

web/src/components/dialogs/archive-project-dialog.tsx
```diff
@@ -0,0 +1,141 @@
...
+                  </p>
+                  <p>Please ensure the following:</p>
+                  <ol>
+                    <li>
+                      All users have opened the app whilst online and confirm
+                      synchronization has completed, indicated by a green tick
+                      next to each record.
+                    </li>
+                    <li>
+                      You have reported the data from the system and stored it
```
stevecassidy: exported?

web/src/components/side-bar/app-sidebar.tsx
```diff
@@ -107,6 +121,31 @@ export function AppSidebar({...props}: React.ComponentProps<typeof Sidebar>) {
...
 
+  if (canSeeTemplates) {
+    const archiveTabTitles: Record<ArchiveTab, string> = {
+      [NOTEBOOK_NAME_PLURAL]: NOTEBOOK_NAME_PLURAL_CAPITALIZED,
+      templates: 'Templates',
+      users: 'Users',
+      teams: 'Teams',
+    };
+    const archiveSubItems: NonNullable<NavItem['items']> =
+      ARCHIVE_TAB_VALUES.map(tabKey => ({
```
stevecassidy: I think we should exclude Teams here until we can archive teams. Also should exclude Users for those without permission to see users (`canSeeUsers`) and I guess templates (`canSeeTemplates`).

app/src/context/slices/projectSlice.ts
```diff
@@ -32,6 +36,104 @@ import {databaseService} from './helpers/databaseService';
...
+  const t = pendingAbsentDirectoryRetryTimer.get(serverId);
+  if (t !== undefined) {
+    clearTimeout(t);
+    pendingAbsentDirectoryRetryTimer.delete(serverId);
+  }
+}
+
+function projectDocIsArchivedOnServer(doc: ProjectDocument): boolean {
+  const s = doc.status as ProjectStatus | string;
+  return s === ProjectStatus.ARCHIVED || s === 'ARCHIVED';
```
stevecassidy: Why this disjunction?

app/src/gui/components/notebook/settings/index.tsx
```diff
@@ -209,16 +217,32 @@ export default function NotebookSettings(props: {uiSpec: ProjectUIModel}) {
...
         <DialogContent>
-          <DialogContentText>
-            Are you sure you want to {DE_ACTIVATE_VERB.toLowerCase()} this{' '}
-            {NOTEBOOK_NAME_CAPITALIZED}? {DE_ACTIVATE_ACTIVE_VERB} may result in
-            unintended data loss if your records have not been uploaded to the
-            server.
-          </DialogContentText>
+          <Stack spacing={2}>
+            <Typography variant="body1">
+              Are you sure you want to {DE_ACTIVATE_VERB.toLowerCase()} the{' '}
```
stevecassidy, line 223: Should include a warning about data loss here if we're going to delete the database.

app/src/gui/components/notebook/settings/index.tsx
```diff
@@ -180,10 +189,9 @@ export default function NotebookSettings(props: {uiSpec: ProjectUIModel}) {
             </Typography>
             <Box>
               <Typography variant={'body2'} sx={{mb: 2}}>
-                {DE_ACTIVATE_ACTIVE_VERB} this {NOTEBOOK_NAME_CAPITALIZED} will
-                remove it from your active notebooks list. Make sure all your
-                data has been synced to the server before{' '}
-                {DE_ACTIVATE_ACTIVE_VERB.toLowerCase()}.
+                {DE_ACTIVATE_ACTIVE_VERB} this {NOTEBOOK_NAME} will remove it
```
stevecassidy, line 192: ...and delete all records on your device.

api/src/api/templates.ts
```diff
@@ -106,6 +126,27 @@ api.get(
   }
 );
 
+/**
+ * GET count of surveys (projects) that still reference this template id.
+ */
+api.get(
+  '/:id/survey-references',
```
stevecassidy: /:id/references perhaps?

api/src/api/templates.ts
```diff
@@ -106,6 +126,27 @@ api.get(
...
+    getResourceId(req) {
+      return req.params.id;
+    },
+  }),
+  processRequest({
+    params: z.object({id: z.string()}),
+  }),
+  async (req, res: Response<GetTemplateSurveyReferencesResponse>) => {
+    const ids = await getProjectIdsReferencingTemplate(req.params.id);
+    res.json({surveyCount: ids.length});
```
stevecassidy: count

web/src/components/tables/users.tsx
stevecassidy: Noted that we do prevent SSO users logging in.  Happy to leave this in place for SSO users.

library/data-model/src/data_storage/migrations/notebookMigrations/migrateV3.ts
```diff
@@ -0,0 +1,37 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @file Notebook migration to schema 3.0 — removes legacy `project_status` from metadata.
+ * Archive state for templates now lives on the template document, not notebook JSON.
+ */
+
+import type {NotebookV1} from './migrateV2';
```
stevecassidy: NotebookV1 was defined in migrateV2 to avoid using the global notebook type (which would be V2 at that point).   Makes sense here to have a locally defined type for this here rather than importing although it will be the same as NotebookV1.  That type was exported for the tests rather than for broader usage.

stevecassidy (changes requested):
No big problems with this, thanks for the detailed work.   Some changes requested with the API endpoints and a few other comments/suggestions.    We are happy with forcing deletion of data on devices when projects are deleted so you can perhaps remove the flag and just go with that behaviour.

web/src/project-archive/project-lifecycle-copy.ts
PeterBaker0: As discussed - this is not priority to consider as no major users will have this combination - we could have a mitigation pathway here such as having users manually clear the DB? Regardless not priority for now

api/src/api/notebooks.ts
PeterBaker0: I think I caught it while here - it was just an inefficiency I noticed while fixing up the export code for handling deletion of related records. It was recomputing the field types heaps.

app/src/gui/components/notebook/settings/index.tsx
PeterBaker0: I think its covered pretty well since its on the prior screen too - I have added clarification "...and remove it from your device."

stevecassidy (approved):
Thanks for the thorough work.

PR #2059, opened 2026-05-01
feat: add AudioRecorder field type (#2004)

### LINKED ISSUE #2004: Add an Audio field for sound recording
Users want to capture sound recordings. Add a new field to capture an audio recording as an attachment.

Caedous:
# feat: add AudioRecorder field type (#2004)

## JIRA Ticket

N/A (GitHub issue #2004)

## Description

Adds a new AudioRecorder field type that allows users to capture audio recordings as attachments. The field uses the Capgo capacitor-audio-recorder plugin and follows the same attachment pattern as the existing FileUploader field.

## Proposed Changes

- Created AudioRecorder edit component with record, pause, resume, stop, and cancel controls
- Created AudioRecorder view renderer with inline audio playback showing "Recording N" labels
- Added inline playback in edit mode using the useAttachments hook
- Handles platform differences between web (Blob/webm) and native (URI/m4a) recording output
- MIME type and file extension derived from the actual recording blob with fallbacks for untyped blobs
- Added field to the registry and designer under the Media category
- Added @capgo/capacitor-audio-recorder dependency to forms and app packages
- Recording controls extracted as a standalone RecordingControls component for future reuse
- Controls pin to the bottom of the screen during recording for visibility while scrolling
- Stop button uses red (error) colour to match standard conventions
- Section navigation blocked during recording via setAttachmentSaving (same pattern as TakePhoto)
- Best-effort finalise on unmount — stops and saves the recording if the component is torn down
- Multiple audio fields on the same form cannot record simultaneously via module-level lock
- Added audio MIME types and codec suffix stripping to the API export filename map to fix .dat export issue

## How to Test

1. Start the local environment (CouchDB, API, app)
2. Open the web manager and create a template
3. Add an AudioRecorder field from the Media category in the field chooser
4. Create a project from the template and open it in the app
5. Create a new record, click Record, grant microphone permission, record a clip, click Stop
6. The saved recording should appear with an inline audio player for playback
7. Test navigation blocking: start recording, try clicking Back or the record list — an alert should appear
8. Test recording lock: add two audio fields, start recording on one — the other should show disabled with a message
9. Test export: export the project as zip — audio files should have .webm extensions, not .dat

## Screenshots

**Field chooser — Audio Recorder under Media category:**
Audio Recorder Field
Audio Recorder Field 2

**Edit mode — recording controls and saved recording with playback:**
Audio Recorder Field 3
Audio Recorder Field 4
Audio Recorder Field 5
Audio Recorder Field 6

**Recording lock — second audio field disabled while first is recording:**
2audiofield

**View mode — recordings shown as "Recording N":**
recording file name

**Export — audio files exported as .webm:**
exportfile

## Additional Information

Closes #2004

Configuration options available in the designer:
- maximumNumberOfRecordings: limits the number of clips (0 = unlimited)
- bitRate: audio bit rate passed to the plugin
- sampleRate: audio sample rate passed to the plugin

## Checklist


PeterBaker0:
@Caedous nice! keen to take a look. Is it ready for review?

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
```diff
@@ -0,0 +1,470 @@
...
+import {AudioRecorderRender} from '../../../rendering/fields/view/specialised/AudioRecorder';
+import {FieldInfo} from '../../types';
+import FieldWrapper from '../wrappers/FieldWrapper';
+
+// ============================================================================
+// Types & Schema
+// ============================================================================
+
+const audioRecorderPropsSchema = BaseFieldPropsSchema.extend({
+  maximum_number_of_recordings: z.number().optional().default(0),
```
PeterBaker0: camelCase consistently would be nice

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
```diff
@@ -0,0 +1,470 @@
...
+
+type AudioRecorderProps = z.infer<typeof audioRecorderPropsSchema>;
+type AudioRecorderFieldProps = AudioRecorderProps & FormFieldContextProps;
+
+interface FullAudioRecorderFieldProps extends AudioRecorderFieldProps {
+  config: FullFormConfig;
+}
+
+const AUDIO_MIME = 'audio/mp4';
+const AUDIO_EXT = 'm4a';
```
PeterBaker0: inline comments explaining what these do would be good - and consequences of changing them (if a later developer was to look)

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
```diff
@@ -0,0 +1,470 @@
...
+  const [elapsed, setElapsed] = useState(0);
+  const [error, setError] = useState<string | null>(null);
+  const tickRef = useRef<ReturnType<typeof setInterval> | null>(null);
+  const startedAtRef = useRef(0);
+
+  const attachmentIds = (state.value?.data as string[]) ?? [];
+  const atLimit = maxRecordings > 0 && attachmentIds.length >= maxRecordings;
+
+  // Load attachments for playback
+  const attachmentService = props.config.attachmentEngine();
```
PeterBaker0: may want to useMemo() this object constructor - depending on rendering frequency - will leave to you.

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
```diff
@@ -0,0 +1,470 @@
...
+      250
+    );
+  }, []);
+
+  const stopTimer = useCallback(() => {
+    if (tickRef.current) {
+      clearInterval(tickRef.current);
+      tickRef.current = null;
+    }
+  }, []);
```
PeterBaker0, line 386: some inline comments would be good here - what's it doing?

PeterBaker0 (changes requested):
Some small code changes - nice! Is there a mechanism to complete the recording before navigating away from the page? We don't want the user to forget to finalise their recording.

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
```diff
@@ -0,0 +1,470 @@
...
+                  border: 1,
+                  borderColor: 'divider',
+                  borderRadius: 1,
+                  p: 1,
+                }}
+              >
+                <Box sx={{display: 'flex', alignItems: 'center', gap: 1}}>
+                  <MicIcon fontSize="small" color="action" />
+                  <Typography variant="body2" sx={{flexGrow: 1}}>
+                    {file.data?.metadata?.filename ?? `Recording ${idx + 1}`}
```
stevecassidy: When we export this data we'll construct a filename based on the record HRID, so best not to display a filename here since it won't persist. Just use 'Recording N' I think.

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
```diff
@@ -0,0 +1,470 @@
...
+          return;
+        }
+      }
+      // Cancel any leftover recording from a previous session
+      try {
+        await CapacitorAudioRecorder.cancelRecording();
+      } catch {
+        // No active recording to cancel — safe to ignore.
+      }
+      await CapacitorAudioRecorder.startRecording({
```
stevecassidy, line 413: I can see that having more than one audio field in a form causes problems since these recording controls are global - you can't start a second one when the first one is going.  We need to see if we can disable the field if another field is already recording which might need some persistent state - but see if you can check the recording status when you render and disable the start button if recording is underway (might be tricky if you don't know which field is actively recording).   If we need some state then we'll have to work out a mechanism for fields to access form level state which they don't at present.

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
```diff
@@ -0,0 +1,470 @@
...
+      }
+      // Ensure a valid MIME type before handing to the attachment service.
+      const typedBlob =
+        blob.type && blob.type.length > 0
+          ? blob
+          : new Blob([blob], {type: AUDIO_MIME});
+
+      const newId = await addAttachment({
+        blob: typedBlob,
+        contentType: typedBlob.type || AUDIO_MIME,
```
stevecassidy: we need to be sure that this type is correct, I got webm audio from my browser based test.  Need to check across a few devices.  Would be good if we could document what format the audio will be in.

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
stevecassidy: And I don't think these are correct, I got webm audio in my test.

stevecassidy (changes requested):
Looking really good, tests on IOS worked ok.  Three main areas to work on:

- when I navigate away from the form (click the finish button) we should end the recording if it is still going, TakePhoto has some code showing how to ensure things are cleaned up 
- more than one audio field is problematic, look at how to disable other fields if one is recording
- need to handle export of audio attachments, they currently go out as .dat files because the export code doesn't know about audio types

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
Caedous: I think this is now fixed in the newest commit.

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
Caedous: I found out what the issue was, it wasn't assigning the default file type correctly. This has been fixed in the newest commit.

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
Caedous: I followed your advice and used useMemo(), made more sense

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
Caedous: Have added some comments explaining this section of code

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
Caedous: Fixed this in both the field data entry view and the form view

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
Caedous: Have fixed this and it now ensures that no secondary audio field can start recording whilst another recording is ongoing.

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
Caedous: Have fixed this to be able to assign correct audio formats. Should now be able to check different audio formats

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
Caedous: Also have now set up a mechanism using some things from TakePhoto to save on exit, but otherwise block navigation away from the form unless the recording is stopped and saved, ensuring that the user does not forget the recording.

Caedous:
Have managed to rectify a lot of the requested changes and fixes, let me know if there is something else I should look at or if everything is up to scratch!

Caedous:
I have also modifed the recording bar, so it moves down the page as you scroll, keeping it visible for the user the entire time until they stop the recording.  Steve mentioned this may be a more useful way of having the recording function.

PeterBaker0:
@Caedous if you navigate to another section, does the recording persist? You mention it prompting you to finalise the recording, but what if we want to start the recording, then go to other sections (to complete the form) while it records in the background. I think this will be quite challenging as the component completely unmounts when changing sections. We can do that as another PR later if need be. 

Once @stevecassidy  is happy I think all good with me.

Caedous:
> @Caedous if you navigate to another section, does the recording persist? You mention it prompting you to finalise the recording, but what if we want to start the recording, then go to other sections (to complete the form) while it records in the background. I think this will be quite challenging as the component completely unmounts when changing sections. We can do that as another PR later if need be.
> 
> Once @stevecassidy is happy I think all good with me.

So it allows you to navigate the form and fill in other sections whilst recording, it just wont let you navigate away from the form (ie. going to another form or the form list) without saving the recording first.  It still lets you interact with everything inside the form as you go, just as long as you dont navigate away from the page. At least in the single test case I did with a text field it worked. Probably more strenuous edge case testing might be required.

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
```diff
@@ -0,0 +1,684 @@
...
+    )
+  )
+    return true;
+
+  // Tab panels / section content within the form
+  if (target.closest('[role="tabpanel"]')) return true;
+
+  return false;
+}
+
```
stevecassidy: I don't like this approach, it is too invasive and doesn't actually work (at present). I get a warning when interacting with another field.  I think the best option is just to finalise the recording if the field is unmounted - this will unfortunately stop the recording if we navigate to another section but we don't have a mechanism to keep it going in the background as yet.  We might leave that for a future update.

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
```diff
@@ -0,0 +1,684 @@
...
+// ============================================================================
+// Preview
+// ============================================================================
```
stevecassidy, line 118: Instead of cancel here, finalise the recording and save the attachment so that we don't lose anything.  This won't trigger if we maintain the setAttachmentSaving setting because that will prevent unmounting, but if we work out a new mechanism then this would be a safe fallback.

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
```diff
@@ -0,0 +1,684 @@
...
+  }, []);
+
+  const handleStart = useCallback(async () => {
+    setError(null);
+
+    if (!acquireRecordingLock(fieldId)) {
+      setError('Another audio field is currently recording.');
+      return;
+    }
+
```
stevecassidy, line 396: spread operator is redundant here since you are creating a new object anyway

stevecassidy, line 396: We need a way to render this component during recording whatever section of the form we are on.  I'd suggest refactoring this control as a separate component as one step towards this. Then we need a hook to have it displayed persistently while the form is in scope - will require a hook passed in from EditableFormManager.

stevecassidy (changes requested):
I've left some comments re. the navigation blocking during recording. I think we'll need a more complex solution which we might not get to with this PR.  Please do what you can within the AudioRecorder component and we can discuss the way forward.

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
Caedous: I believe I've managed this now, I followed a similar way to the TakePhoto and made it its own component.

Caedous: Fixed, no conditional spreads and no double-spreading

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
Caedous: Audio will no longer be lost should we have some mechanism that allows for unmounting.

library/forms/lib/fieldRegistry/fields/AudioRecorder/index.tsx
Caedous: Completely changed to rely on setAttachmentSaving and removed all other references to the old click handler

PeterBaker0 (approved):
Have tested in the app and it is working well. Just have two requests:
1. Could you please add documentation to the field documentation areas for the new field?
2. Could you add a new optional flag to the field configuration for displaying or not displaying the download button?

One of the requirements in our usage of the app is that it's somewhat difficult to extricate data from the system. Having a visible download button is a very quick way for users to get data out, so we'd like to disable that.

stevecassidy:
I would say remove the download option, we don't allow that for photos so no need for it here either.

stevecassidy (approved):
All good now.

PR #2070, opened 2026-05-06
feat: public templates and visibility controls

PeterBaker0:
### Summary

Adds **public/private template visibility** end to end: storage (`isPublic`), permissions for reading public templates vs private access, an API to toggle visibility, and UI to manage and explain it.

Motivation for this is to make the 'shared template' workflow easier to maintain and scale. Currently distributing templates to teams as starting points requires having duplicates created within each team. This then means that updating a 'shared template' requires updating every instance of the duplicated template. This allows ops admins to make templates shared so that users can access/use them directly, restricting the update space to one spot. There is still the issue where users create a customised template, and update that, which won't inherit the upstream changes - but that is out of scope for this PR.

**Requires DB migration - non-breaking**.

### Changes

- **Data model & migrations**: Persist `isPublic` on template documents; extend permissions with actions such as `READ_PUBLIC_TEMPLATE_DETAILS` and `CHANGE_TEMPLATE_VISIBILITY`, plus `userCanReadTemplateDocument` (private access or public + public-read permission). Adds `globalRolesGrantAction` for cases without a resource id.
- **API**: List/get/references use `userCanReadTemplateDocument` instead of only `READ_TEMPLATE_DETAILS` on the template id. New `PUT /api/templates/:id/set-visibility` updates `isPublic` only. Template responses can include `ownedByTeamDisplayName` for list/detail without extra team calls. Create template accepts optional `isPublic` (default false). Swagger updated.
- **Web**: Template visibility dialog (make public/private, copy, confirmations), hooks for `set-visibility`, and template tables/forms/team views updated for public state and display.
- **Tests**: API template tests and data-model migration tests expanded.

### Notes

- Public templates are intended as **read-only for general users**; changing visibility remains restricted to authorized roles.
- Our use case at the moment only wants operation admins to be able to access this feature, as making templates public is something that should be done with scrutiny. Users may accidentally leak IP or other sensitive materials. Open to discussion in opening this up moving forward, but would like time to treat this as 'admin only' for the time being.

### DIFF
--- a/api/src/api/templates.ts
+++ b/api/src/api/templates.ts
@@ -28,9 +28,11 @@ import {
   PostCreateTemplateInputSchema,
   PostCreateTemplateResponse,
   PostRestoreTemplateResponse,
+  PutTemplateSetVisibilityInputSchema,
   PutUpdateTemplateInputSchema,
   PutUpdateTemplateResponse,
   Role,
+  userCanReadTemplateDocument,
 } from '@faims3/data-model';
 import express, {Response} from 'express';
 import {z} from 'zod';
@@ -43,13 +45,15 @@ import {
   getTemplate,
   getTemplates,
   restoreTemplateFromArchive,
+  setTemplateVisibility,
   updateExistingTemplate,
+  withOwnedByTeamDisplayName,
+  withOwnedByTeamDisplayNames,
 } from '../couchdb/templates';
 import * as Exceptions from '../exceptions';
 import {
   isAllowedToMiddleware,
   requireAuthenticationAPI,
-  userCanDo,
 } from '../middleware';
 
 import {saveExpressUser} from '../couchdb/users';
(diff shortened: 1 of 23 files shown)


api/src/api/templates.ts
```diff
@@ -203,6 +231,28 @@ api.post(
   }
 );
 
+/**
+ * PUT set public/private visibility only.
+ */
+api.put(
+  '/:id/set-visibility',
```
stevecassidy: I wonder if we might want to support more metadata per template in future. We could make this an endpoint to update only the template metadata `/:id/metadata` rather than just the one property.  Payload could just allow setting isPublic for now but we could expand later if needed.

stevecassidy: Ok, I see that it's not part of metadata...

stevecassidy: Actually, ownedByTeam is set via POST to /:id below, we should be consistent and include this new property there as well. It deals with only updating one property.

stevecassidy (changes requested):
Updates to the API surface as discussed.

stevecassidy (approved):

api/src/api/templates.ts
PeterBaker0: As discussed - changing /set-visibility to /visibility PUT. Leaving team stuff as is for now - but should eventually spin off PUT /team or similar and move the team specific side effects to there.

For consistency, we should follow this approach moving forward

POST / - create it
PUT /:id - update minor/metadata properties - uses patch semantics i.e. only those properties provided in the payload are updated. In essence (with checks etc), `newObject = {...old, ...updatePayload}`
PUT /:id/<property> - update special/significant properties

Examples of special properties:

template team
template visibility
survey status
survey team 

etc.

PR #2076, opened 2026-05-12
feat: global cross-database migrations and migration docs

PeterBaker0:
This change adds **global migrations** that can coordinate upgrades across **more than one Couch DB in a single batch** (e.g. per-project **metadata** and **data** DBs, or other typed participants), with explicit **participant multiplicity** (`single` vs `all-of-type`) and supporting **resolver / registry / network invariant** helpers plus tests.

**`migrationService`** is extended so **`migrateDbs`** can run these global steps safely alongside existing per-DB behaviour; **`migrationsDB` types** and **API/Couch wiring** (`index`, `notebooks`) are updated where needed.

**Documentation** is expanded: **`CouchMigrations.md`** (global flow, batch semantics, operator guidance), smaller updates to **`DATABASE.md`**, **`NotebookMigrations.md`**, and **`DeployingAWSStack.md`**. **`metadata-design.md`** is kept aligned with the direction to use **global migrations** for metadata/UI layout cutover.

**Tests** cover the new global paths and invariants (`migrationService`, `migrationNetworkInvariants`).

### DIFF
--- a/docs/developer/docs/source/markdown/CouchMigrations.md
+++ b/docs/developer/docs/source/markdown/CouchMigrations.md
@@ -2,18 +2,23 @@
 
 ## Overview
 
-In `library/data-model/src/data-storage/migration` is a structured framework for managing database schema evolution across multiple PouchDB databases in our application. It provides a systematic way to apply incremental changes to database structures while maintaining data integrity and tracking migration history.
+The Couch / Pouch migration framework lives under:
+
+`library/data-model/src/data_storage/migrations/`
+
+It manages **schema evolution** across the many CouchDB databases FAIMS uses (people, projects, per-project data and metadata, etc.). It tracks a **numeric version per physical database** in a dedicated **migrations** database, applies **per-document** transforms, and supports **global** steps that need several DB connections in one atomic operation.
 
 Includes:
 
-- Version-controlled database schemas
-- Incremental, one-way migrations
-- Comprehensive logging of migration activities
-- Batch processing to handle large datasets efficiently
-- Error handling with detailed issue reporting
-- Health status tracking for each database
+- Version-controlled database schemas (`DB_TARGET_VERSIONS` per `DatabaseType`)
+- **Individual** migrations: one step per physical DB (`DB_MIGRATIONS`, always `from → from+1`)
+- **Global** migrations: coordinated jumps across types (`GLOBAL_MIGRATIONS` in `globalMigrations.ts`)
+- Comprehensive logging (`migrationLog` on each DB’s migration document; globals may set `globalMigrationId`)
+- Batch processing inside each DB for large `allDocs` scans
+- Error handling with detailed issue reporting; `not-healthy` status when a DB’s migration fails
+- Static **network invariants** (`migrationNetworkInvariants.ts`) so there is a unique path from default to target version and globals cannot overlap individuals in conflicting ways
 
-The system maintains a separate migration metadata database that tracks the version and migration history of each application database, ensuring that migrations are applied exactly once and in the correct sequence.
+The migrations **metadata** database stores one document per `(DatabaseType, dbName)` pair, recording `version`, `status`, and `migrationLog` history.
 
 ## Core Concepts
 
@@ -31,6 +36,7 @@ export enum DatabaseType {
   PEOPLE = 'PEOPLE',
   PROJECTS = 'PROJECTS',
   TEMPLATES = 'TEMPLATES',
+  TEAMS = 'TEAMS',
 }
 ```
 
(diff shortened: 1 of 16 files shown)


library/data-model/src/data_storage/migrations/globalMigrationTypes.ts
```diff
@@ -0,0 +1,128 @@
...
+ *   `from`. That one DB is the participant. If two (or zero) match, the global
+ *   does not apply—avoids “which directory did we mean?”.
+ * - `all-of-type`: **Every** physical DB of that `dbType` in the batch that is
+ *   still behind target must be at **the same** `from` version, and **all** of
+ *   them are upgraded together in this step. Use when the migration must run
+ *   once per logical “slice” of DBs (e.g. every `METADATA` DB in this batch).
+ *
+ * **How that maps to FAIMS `DatabaseType`**
+ *
+ * Types that are normally **one Couch DB per server** (directory, projects list,
```
stevecassidy, line 21: Is there really a case where we'd use 'single' for eg. the project databases? I think the choice of single/all is bound to the particular database rather than being a choice per migration.

library/data-model/src/data_storage/migrations/globalMigrationTypes.ts
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Every {@link LoadedDbHandle} taking part in this global step, deduped by
+   * `(dbType, dbName)`.
+   */
+  allHandles(): LoadedDbHandle[];
+  /**
+   * The full list passed into `migrateDbs` for this invocation, including DBs
+   * that are not part of this global migration (e.g. other projects’ DBs).
+   */
+  allBatchHandles(): LoadedDbHandle[];
```
stevecassidy, line 94: I don't think I understand the difference between 'batch' and not batch here or from other documentation.  What is a batch, is it smaller than the set returned by 'allHandles' or larger?

library/data-model/src/data_storage/migrations/migrationService.ts
```diff
@@ -283,196 +575,104 @@ export async function migrateDbs({
   migrationDb: MigrationsDB;
   userId?: string;
 }): Promise<void> {
-  // Process each database one by one
-  for (const {dbType, dbName, db} of dbs) {
-    // Track migration start time
-    const migrationStartTime = Date.now();
+  const handles: LoadedDbHandle[] = [];
```
stevecassidy, line 578: This needs some documentation. I've traced back to here to understand the difference between batchHandles and handles in runGlobalMigrationStep. So here I can see that handles is an array of _all databases_ with references to the db and migration document.

library/data-model/src/data_storage/migrations/migrationService.ts
```diff
@@ -143,6 +149,291 @@ export function identifyMigrations({
...
+    }
+  } catch (logError) {
+    console.error(
+      `Failed to log migration failure for ${dbName} (${dbType}):`,
+      logError
+    );
+  }
+}
+
+async function runGlobalMigrationStep({
```
stevecassidy, line 354: documentation please...

library/data-model/src/data_storage/migrations/globalMigrationResolver.ts
```diff
@@ -0,0 +1,235 @@
...
+        );
+      }
+      return arr[0];
+    },
+  };
+}
+
+/**
+ * Returns the first global migration definition that matches the current batch,
+ * or null. Matching is order-dependent: earlier definitions win.
```
stevecassidy, line 191: also returns the map of dbtype->dbHandle for those databases involved in the migration

library/data-model/src/data_storage/migrations/migrationService.ts
```diff
@@ -283,196 +575,104 @@ export async function migrateDbs({
...
-
-          // Save the updated migration document
-          await migrationDb.put(migrationDoc);
-        }
-      } catch (logError) {
-        // At this point, we've failed to migrate and also failed to log the failure
-        console.error(
-          `Failed to log migration failure for ${dbName} (${dbType}):`,
-          logError
+    const globalMatch = findApplicableGlobalMigration(handles, GLOBAL_MIGRATIONS);
```
stevecassidy, line 645: Re. comment above about 'handles' - here you effectively filter the larger 'handles' array to get just those databases that will be touched by the selected migration, so globalMatch is a really bad name for this. This is a subset of the overall set of databases plus a migration spec.

library/data-model/src/data_storage/migrations/migrationService.ts
```diff
@@ -143,6 +149,291 @@ export function identifyMigrations({
...
+  allHandles,
+  migrationDb,
+  userId,
+}: {
+  match: NonNullable<ReturnType<typeof findApplicableGlobalMigration>>;
+  allHandles: LoadedDbHandle[];
+  migrationDb: MigrationsDB;
+  userId: string;
+}): Promise<'success' | 'failure'> {
+  const {definition, matchedByDbType} = match;
```
stevecassidy, line 365: matchedByDbType is the collection of db handles that we will be migrating, allHandles is the complete set of db handles

library/data-model/src/data_storage/migrations/globalMigrationResolver.ts
```diff
@@ -0,0 +1,235 @@
...
+    const key = `${h.dbType}:${h.dbName}`;
+    if (!uniqueMap.has(key)) {
+      uniqueMap.set(key, h);
+    }
+  }
+  return Array.from(uniqueMap.values());
+}
+
+/**
+ * Builds the {@link GlobalMigrationRunContext} passed into
```
stevecassidy, line 119: I think the 'match' terminology is what is confusing me here because at this point we're creating a context for a single global migration - the matching process is outside of this and not really relevant.  What is key is that 'matchedByDbType' is the set of databases that we're going to migrate and 'allBatchHandles' is the complete set of databases handles. Maybe calling them 'targetDBMap' and 'allDBHandles' might be clearer.

stevecassidy (changes requested):
Have left some comments relating to trying to understand the code but the main observation is that we could get rid of individual migrations and just have the global migrations framework - single migrations are just a special case where there is only one database involved.  

I'd really like to see an example migration in this framework, I think writing one would be useful to test out the spec.  You could port the old single db migrations to this framework to achieve that.  

I'd really like to see migrations structured as single modules as well - yes I know it's just moving code around but it would make it much easier to follow a migration and see what it changes.  It might encourage us to better document each migration as well as a header comment in the migration module.

docs/developer/docs/source/markdown/CouchMigrations.md
```diff
@@ -430,17 +458,262 @@ export const dataV2toV3Migration: MigrationFunc = doc => {
...
+      from: 7,
+      to: 8,
+      multiplicity: 'single',
+    },
+  ],
+  run: async ctx => {
+    const projectsParticipant = ctx.getUnique(DatabaseType.PROJECTS);
+    const templatesParticipant = ctx.getUnique(DatabaseType.TEMPLATES);
+
+    const peopleInBatch = ctx.allHandlesForType(DatabaseType.PEOPLE);
```
stevecassidy, line 652: `batch` is confusing here because this db comes from the complete set of db handles, if it doesn't exist then there is no people DB.

docs/developer/docs/source/markdown/CouchMigrations.md
```diff
@@ -430,17 +458,262 @@ export const dataV2toV3Migration: MigrationFunc = doc => {
...
+    const adminId = ctx.userId; // or resolve from env / config
+
+    type DocRow = PouchDB.Core.ExistingDocument<Record<string, unknown>>;
+    const enrich = async (
+      db: DatabaseInterface,
+      pickUserId: (doc: DocRow) => string | undefined
+    ) => {
+      const res = await db.allDocs({include_docs: true});
+      for (const row of res.rows) {
+        if (!row.doc || row.id.startsWith('_design/')) continue;
```
stevecassidy, line 669: Would we ever use this mechanism to update design documents in a db? I don't think so.  Maybe we don't need the unlimited power in this design - if what I want is to be able to reference other databases when updating documents in my target database we could achieve that with a more restrictive design that just needs me to write per-document migration function but one that can have access to other databases.  

This example is not really global, it's just applying the same migration fn (enrich) to two databases where that migration fn queries the people db.   

Do we have a use case for more complex global migrations in mind?

Issue #1718, opened 2025-10-27
Upgdate @mui modules to more recent versions


stevecassidy:
While moving to PNPM I note that the @mui/XXX modules are all quite old.  We should move to the most recent versions as it's now hard to find the documentation for the versions we're using and we're not benefiting from bug fixes or improvements.

PR #2081, opened 2026-05-15
Chore/migrate to mui v9

stevecassidy:
# Chore/migrate to mui v9

## Ticket

#1718

## Description

Upgrade MUI to the most recent release (9.0.1).

## Proposed Changes

__AI driven__ upgrade to the most recent @mui/material. 

As part of the migration, some styles were moved into the theme rather than being left as customisations on elements.  

## How to Test

Appearance of app/ and web/ should be the same as before.  Build should be clean, tests should pass.

## Checklist


app/src/gui/components/notebook/record_table.tsx
```diff
@@ -470,7 +472,7 @@ export function buildColumnFromSystemField({
         type: 'boolean',
         flex: 0,
         minWidth: 125,
```
Caedous: When I built this branch locally and opened a notebook with records, the records table failed to render. The error boundary caught a `TypeError: Cannot read properties of undefined (reading 'row')` thrown from this `valueGetter` (line 462 hits the same error).

Repro: ran `pnpm build` and `npx vite`, logged in, opened a notebook with 3 records. Full stack trace in the review summary.

app/package.json
```diff
@@ -55,15 +55,16 @@
...
     "@ionic/pwa-elements": "^3.2.2",
-    "@mui/icons-material": "5.15.8",
-    "@mui/lab": "5.0.0-alpha.129",
-    "@mui/material": "5.15.14",
-    "@mui/x-data-grid": "6.20.4",
+    "@mui/icons-material": "9.0.1",
+    "@mui/lab": "9.0.0-beta.3",
+    "@mui/material": "9.0.1",
+    "@mui/material-pigment-css": "9.0.1",
+    "@mui/x-data-grid": "8.28.6",
```
Caedous: `@mui/x-data-grid` is at `8.28.6` and `@mui/material` is at `9.0.1`. I wasn't sure if those versions are meant to go together.

app/package.json
```diff
@@ -55,15 +55,16 @@
     "@capacitor/status-bar": "^7.0.1",
     "@capgo/capacitor-speech-recognition": "^8.0.4",
     "@emotion/react": "11.11.4",
-    "@emotion/styled": "11.10.8",
+    "@emotion/styled": "11.14.1",
```
Caedous, line 59: `@emotion/react` is `11.11.4` and `@emotion/styled` is `11.14.1`. I'd seen those usually kept on matching versions but I'm not sure if it's an issue here.

web/package.json
```diff
@@ -28,9 +28,10 @@
     "@hookform/resolvers": "^3.9.1",
     "@lexical/react": "^0.12.4",
     "@mdxeditor/editor": "^1.13.5",
-    "@mui/icons-material": "^5.14.3",
-    "@mui/lab": "^5.0.0-alpha.137",
-    "@mui/material": "^5.14.2",
+    "@mui/icons-material": "9.0.1",
+    "@mui/lab": "9.0.0-beta.3",
+    "@mui/material": "9.0.1",
+    "@mui/material-pigment-css": "9.0.1",
```
Caedous: I noticed `@mui/material-pigment-css` was added in both `app/package.json` and here in `web/package.json`. I had a quick look but couldn't see where it was in the Vite config. Just wanted to check whether it's needed or if it was added by mistake.

library/forms/package.json
```diff
@@ -34,11 +34,11 @@
     "@capacitor/geolocation": "^7.1.5",
     "@capgo/capacitor-speech-recognition": "^8.0.4",
     "@emotion/react": "11.11.4",
-    "@emotion/styled": "11.10.8",
+    "@emotion/styled": "11.14.1",
```
Caedous, line 38: Same thing I mentioned in `app/package.json` — `@emotion/react` and `@emotion/styled` are on different versions here. Wasn't sure if it matters.

app/src/constants/privateRouter.tsx
```diff
@@ -126,7 +126,7 @@ const OfflineLoginBanner = ({loginPath, visible}: OfflineLoginBannerProps) => {
         }}
       >
         <Alert
-          icon={<ErrorOutlineIcon sx={{fontSize: 28}} />}
+          icon={<ErrorOutlineIcon fontSize="large" />}
```
Caedous, line 129: Small thing I noticed: this changed from `sx={{fontSize: 28}}` to `fontSize="large"`. I think `large` is a bit bigger than 28 by default, so the icon might end up slightly larger than before. The PR notes say appearance should be unchanged, so flagging just in case.

app/src/constants/privateRouter.tsx
```diff
@@ -1,4 +1,4 @@
-import ErrorOutlineIcon from '@mui/icons-material/ErrorOutline';
+import ErrorOutlineIcon from '@mui/icons-material/ErrorOutlineOutlined';
```
Caedous, line 1: The import went from `ErrorOutline` to `ErrorOutlineOutlined`. The doubled "Outline" felt odd to me — I would have expected just `ErrorOutlined`. Maybe worth a quick check that this is the right icon name?

I noticed similar doubled-suffix renames in `SpeechToTextButton.tsx` and a few of the web designer field editors. `MapField.tsx` looks like it was done the way I'd expect (`CheckCircleOutline` → `CheckCircleOutlined`) so there's some inconsistency.

Caedous (commented):
Built and ran the branch locally. The records table didn't render and caught a `TypeError` thrown from a `valueGetter` in `record_table.tsx`.

Repro: ran `pnpm build` then `npx vite` in the app folder, logged in as admin, opened a notebook with a few records.

A few other smaller things flagged inline (package versions, an icon import that looked odd, a couple of small visible changes).

app/src/gui/components/notebook/record_table.tsx
stevecassidy: updated calls to valueGetter to new API

app/src/constants/privateRouter.tsx
stevecassidy: no, this is the right icon

stevecassidy:
Parking for now so we can merge the major UI review PR first.

ranisa-gupta16 (approved):
Thanks Steve
Looks really good and  clean overall. 
One thing, branch is behind main,  lock file conflicts, the AudioRecorder field isn't on the branch and the audio-recorder dep got dropped. Don't see any issues as compared to changes we did to latets merges and designer PR. looks all good to me.

PR #2083, opened 2026-05-15
Fix docker build of api

stevecassidy:
# Fix docker build of api

## JIRA Ticket

None

## Description

Docker build workflow was failing in the last day.

## Proposed Changes

Seems to be due to some upstream change in the docker base image or corepack.  Updated
to avoid relying on pnpm paths by using `pnpm dlx` 

## How to Test

Local docker build should work again.   Workflow should succeed.

## Checklist


PeterBaker0 (approved):
Looks fine to me

PR #2091, opened 2026-05-26
fix overlay issue in android side nav 

ranisa-gupta16:
## JIRA Ticket

[[<JIRA_TICKET>](https://jira.csiro.com/browse/<JIRA_TICKET>)
](https://jira.csiro.au/browse/BSS-1105)

## Description

Fixes the hamburger drawer sing out item being hidden behind the android system navigation overlay on devices with nav buttons. 

## Checklist


PeterBaker0 (approved):
Confirmed this resolves issue on Android device. Could you double check we don't already define a constant for this elsewhere - maybe just search for existing usages and whether we extrapolate a constant which we can import and reuse - otherwise all good 👍

ranisa-gupta16:
there are css custom properties in app/src/index.css but not used anywhere in codebase. may be that's the reason why this issue never got resolved and nobody came across it.

PR #2094, opened 2026-05-27
Feat/add warning dialog on finish record

ranisa-gupta16:
# feat/fix/chore: pr title

## JIRA Ticket

[BSS-1145](https://jira.csiro.au/browse/BSS-1145)

As discussed and reviewed by Elana, this is the final update:

Adds a warning dialog when the user clicks Finish on a record that still has required fields empty or validation errors. The dialog shows a count of fields with errors; clicking Go back and review (or the count itself) closes the dialog and scrolls the form to the first problematic field. Also adds a reusable ConfirmDialog component in @faims3/forms that other dialogs can use going forward.

here are screenshots for reference: 

image
image

## Checklist


library/forms/lib/formModule/formManagers/EditableFormManager.tsx
```diff
@@ -574,32 +576,112 @@ export const EditableFormManager: React.FC<
...
+    // Validation errors from react-form's per-field meta.
+    for (const [name, meta] of Object.entries(fieldMeta)) {
+      if ((meta?.errors as unknown[] | undefined)?.length) {
+        problematic.add(name);
+      }
+    }
+
+    const firstField = problematic.values().next().value ?? null;
+    return {count: problematic.size, firstField};
+  }, [dataEngine.uiSpec, form, visibleMap]);
```
PeterBaker0: This is a bit scary for a couple of reasons. 

1. this is duplicating existing logic for errors / required fields 
2. the dependency array is pretty dubious - particularly the inclusion of 'form' - it's uncertain when this will trigger a rerun - should we be more specific e.g. form.state.fieldMeta and values? Even then it's going to fire a lot of times doing completely unnecessary work 
3. performance

I'd propose instead that when you click finish - it triggers this _once_ (after flush saving, settling happens first to ensure everything is up to date) to determine if we need to show the dialog - it passes into the dialog at this point this information about the missing fields/errors so the dialog can display it. This way we don't run this thousands of times during form editing, and avoid any confusing dependency arrays.

PeterBaker0 (changes requested):
Concerns about efficiency and reliability.

ranisa-gupta16:
thanks for reviewing 

Have considered above in mind to look at broader performance picture.
Have dropped the wrapping usecallback so it only runs on finish click.

resused completionresult so we get fieldname not just counts, 
fieldMeta.errors is the main source of truth so getting that form there. 

peroformance wise, the flow now run once per finish clikc inside guardfinish  after flushSave, so we see the settled state , not on render or keystroke.

On deps, form is still in guardFinish bcoz useForm() returns a stable reference  so it doesnt trigger re-creation on keystrokes. 

also added a ref so double click during flush so it doesnt fire  guard twice.

let me know if anything else is needed to be updated.

library/forms/lib/formModule/formManagers/EditableFormManager.tsx
```diff
@@ -574,32 +576,109 @@ export const EditableFormManager: React.FC<
...
-  }, [navigationButtons, isAttachmentSaving]);
+    const isFinishButton = (id: string) =>
+      id === 'finish-to-parent' || id === 'finish-to-list';
+    return navigationButtons.map(b => {
+      const base = isAttachmentSaving
+        ? {...b, disabled: true, loading: true, statusText: 'saving attachment…'}
+        : b;
+      return isFinishButton(b.id)
+        ? {...base, onClick: guardFinish(b.onClick)}
+        : base;
```
PeterBaker0: This approach is very fragile. Don't rely on the IDs. We construct the navigation button objects, so we can change how they are generated to indicate whether they require warning guard.

PeterBaker0 (changes requested):
The approach of relying on fragile ID references is not reliable - please augment the navigation button generation to flag if the button requires a completion warning

library/forms/lib/formModule/formManagers/EditableFormManager.tsx
ranisa-gupta16: Updated
The button type now carries an optional flag for whether it needs the warning dialog, the finish buttons set that flag when they're built, and the form manager just reads the flag
no morre mmatching the string ids

PeterBaker0 (approved):
Works well and new approach is much more efficient and maintainble. 

My only idea for enhancement we could do (later) is to improve the highlighting of errors after navigating back. We might want to treat the transition from the finish warning -> the targeted site as a trigger for the validation mode of the target section to be 'all'? (Similar to the trigger when we leave a section in typical navigation). Not critical though.

PR #2102, opened 2026-06-03
refactor: unify UiSpec field/component-parameter types under a Zod source of truth in @faims3/data-model

PeterBaker0:
## Description

Unifies the `uiSpecification` field typing story. Previously, field and
component-parameter shapes were duplicated and drifting across
`@faims3/data-model`, `@faims3/forms`, and the designer (`@web`), with the
designer maintaining a large hand-written `ComponentParameters` union. This PR
makes `@faims3/data-model` the single Zod-based source of truth, has
`@faims3/forms` define each field's complete props contract, and types the
designer on top of those — removing the duplication and a chunk of dead code.

## Proposed Changes

**`@faims3/data-model` (source of truth)**
- Added `BaseFieldParametersSchema` / `BaseFieldParameters` (the base parameter
  envelope shared by every field) at the lowest layer.
- Reshaped `FieldDefinition`: `component-parameters` is now
  `BaseFieldParametersSchema.passthrough()` instead of `z.record(z.any())`. The
  exported type is derived from the strict shape (no index signature) so it stays
  composable with `Omit`/intersection downstream, while the schema still passes
  unmodelled keys through at runtime.
- Adopted the standard-vs-compiled pattern for fields: `conditionFn` now lives on
  `CompiledFieldDefinition`/`CompiledFieldDefinitionSchema` (mirroring views)
  rather than being an optional field on the base definition.
- Removed the deprecated `is_logic` and `access` array and their runtime paths;
  historical data is handled by migrations. Removed associated dead code in
  `types.ts`, `storageFunctions.ts`, and `queries.ts`.

**`@faims3/forms` (complete per-field contracts)**
- All field modules and wrappers now import base params from `@faims3/data-model`
  (local `BaseFieldPropsSchema` removed; no type-alias re-exports).
- Extracted a single shared `ChoiceElementProps` schema reused by RadioGroup,
  Select, and MultiSelect.
- Extended several props schemas to be complete contracts (e.g. NumberField
  `InputProps`, TemplatedString `template`/`InputLabelProps`, RichText `content`).
- Each field exports its `*PropsSchema` + `*Props` at its definition site,
  surfaced through a new `fieldRegistry/fieldParams.ts` barrel.

**`@web` designer**
- Replaced the ~80-line `ComponentParameters` union with a one-line envelope:
  `ComponentParametersEnvelope = BaseFieldParameters & { [k: string]: unknown }`.
- Each field editor now narrows that envelope to the forms-defined props type for
  its field, rather than the designer redeclaring parameter shapes.
- Palette templates (`fields.tsx`) and fixtures updated to satisfy the base
  contract (`name` is required).

**Docs**
- Updated the forms developer docs (`quickstart`, `05-field-system`,
  `06-field-spec`, `07-validation`, `index`, special-fields) to reflect the new
  `BaseFieldParametersSchema` location, the standard-vs-compiled field pattern,
  the export-for-designer workflow, the current registry list/resolution, and
  corrected stale designer paths and legacy `validationSchema` examples.

**API**
- Updated a notebook test fixture in `exportDeletedRelationships.test.ts` to
  include the now-required `name`.

## How to Test

- `npx turbo build` — all packages build (data-model, forms, api, web, infra).
- `cd library/data-model && npm test` — 317 pass / 1 skipped.
- `cd library/forms && npm test` — 42 pass.
- `cd web && npx tsc --noEmit` — clean (aside from a pre-existing, unrelated
  `DesignerWidget.tsx` `project_id` error on this branch).
- In the designer: add each field type from the palette, edit its options, and
  confirm the field editors (choice options, map, advanced-select, templated
  string, rich text, number, address, date/time, speech) read and write
  parameters correctly, and that conditions referencing choice options still work.

## Additional Information

- Designer editors narrow the open envelope via `as <FieldProps>` casts; a couple
  (`MapFieldProps`, AdvancedSelect's `optiontree`) go through `as unknown as`
  because required schema fields don't structurally overlap the envelope's index
  signature. A future tightening could replace these with typed accessor helpers
  keyed on `component-name`.
- `RelatedRecord`'s `relation_linked_vocabPair` / `related_type_label` remain
  untyped passthrough metadata (not consumed by the renderer).

## Checklist


PeterBaker0:
It's a step closer to the single field registry, but there is some pretty deep circular dependencies preventing going further.
 
would require breaking things up quite differently - probably pulling a lot of the type layer into a separate base module
 
or putting more stuff into the data model which is currently spread across forms/designer/data-model
 
we could dynamically build a zod validation model for the uiSpec based on this work - would just need to upgrade the component-parameters to use the registered zod model for each field type in the registry - however it's not clear where/how this fits in just yet, and this is a necessary step in that direction

ranisa-gupta16 (approved):
Have tested the designer end to end, took a thorough look at this really like having one 
canonical Zod source for the UI spec types. The separation between the strict FieldDefinition type and the passthrough schema isgood.
I ran different notebook jsons to see how validations work, worked well for me across diff. scenarios like blank notebook, wrong types, missing req. field, legacy wrapped uispec.

library/data-model/src/uiSpecification/types.ts
```diff
@@ -1,31 +1,265 @@
...
+export type UiSpecFields = z.infer<typeof UiSpecFieldsSchema>;
+
+/** Compiled field definitions keyed by field name. */
+export const CompiledUiSpecFieldsSchema = z.record(
+  z.string(),
+  CompiledFieldDefinitionSchema
+);
+export type CompiledUiSpecFields = z.infer<typeof CompiledUiSpecFieldsSchema>;
+
+/** A viewset: a named form type composed of one or more views. */
```
stevecassidy: I wish we could just call them forms and sections. Not the time to make that change but maybe in this file make it clear that 'viewset === form' and 'view === section'.

stevecassidy (approved):
Looks good, very nice bit of refactoring setting us up for the future.

library/data-model/src/uiSpecification/types.ts
PeterBaker0: I have done all changes I can non-destructively and marked this for later transition - would be a possible migration with the new frameworks - but big changeset

PR #2108, opened 2026-06-05
update designer dialog for fixing progress upon navigation

ranisa-gupta16:
## JIRA Ticket

](https://jira.csiro.au/browse/BSS-1302)

## Description

Fixes the issue where progress was being lost in the Designer when 
switching browser tabs.

If you opened the Designer, added a field, started typing into its 
name, then switched to another tab and came back — the field you 
were typing into would reset to empty.

The Designer dialog was watching the notebook prop too eagerly. 
When you switched tabs and came back, the parent re-fetched the 
notebook from the backend. 

The dialog now snapshots the notebook only once at the moment 
you open it. Refetches that happen while the dialog is open are 
ignored, so your work stays put.

## Checklist


PeterBaker0 (approved):
Works as described.

PR #2113, opened 2026-06-09
feat: add directional record sync modes with large-notebook defaults and settings UX

PeterBaker0:
Replaces the binary sync on/off model with explicit record replication modes for activated notebooks, while keeping local-first PouchDB writes unchanged.

- **Sync modes:** `none` (local only), `push` (upload only), `pull` (download only, wired but not offered in UI), and `both` (two-way). Persisted `isSyncing` is migrated to `syncMode` via redux-persist **v2**.
- **Replication:** `createPouchDbReplication()` supports one- and two-way live replication; lifecycle is centralised through `setSyncMode` and `replaceProjectReplication`.
- **Activation defaults:** When online, fetches server `recordCount` and defaults to **upload only** above `VITE_SYNC_PUSH_ONLY_RECORD_THRESHOLD` (default 500); offline or API failure falls back to **upload and download**. Shows a titled snackbar (“Sync mode changed”)
- **Settings UX:** Sync mode dropdown (download-only hidden), help dialog, plain confirmation dialog with shared copy, attachment download gated on pull-capable modes. Uses configurable `{NOTEBOOK_NAME}` labels.
- **Record list banner:** Dismissible warning when two-way sync is active and record count exceeds threshold; links to Settings; dismissal stored in localStorage and cleared on deactivation.
- **Fixes:** fix notebook tab index mapping for Settings navigation.
- **Docs/tests:** User sync docs updated; tests for replication factory, persist migration v2, and activation sync defaults.

## Test plan

- [ ] Activate a small notebook online → defaults to upload and download; no push-only snackbar
- [ ] Activate a large notebook online (or set threshold very low e.g. 1 record) → upload only + “Sync mode changed” alert
- [ ] Activate offline → upload and download; record-list banner appears once `recordCount` is known
- [ ] Change sync modes in Settings; confirm dialog copy and help dialog match
- [ ] Change sync modes and validate behaviour with other connected client
- [ ] Dismiss record-list banner; de/re-activate → banner shows again
- [ ] Toggle “Get attachments from other devices” — enabled only under two-way sync
- [ ] Upgrade from persisted state with `isSyncing` → correct `syncMode` after rehydrate

### DIFF
--- a/app/src/buildconfig.ts
+++ b/app/src/buildconfig.ts
@@ -131,6 +131,31 @@ function pouch_batch_size(): number {
   }
 }
 
+// This number requires more tuning based on upcoming load testing results. 500
+// is probably okay for most devices to two-way sync, but this is a more
+// conservative estimate.
+const DEFAULT_SYNC_PUSH_ONLY_RECORD_THRESHOLD = 500;
+
+/*
+ * Record count above which activation defaults to push-only sync (when online).
+ * See VITE_SYNC_PUSH_ONLY_RECORD_THRESHOLD.
+ */
+function sync_push_only_record_threshold(): number {
+  const raw = import.meta.env.VITE_SYNC_PUSH_ONLY_RECORD_THRESHOLD;
+  if (raw === '' || raw === undefined) {
+    return DEFAULT_SYNC_PUSH_ONLY_RECORD_THRESHOLD;
+  }
+  try {
+    const parsed = parseInt(raw, 10);
+    return Number.isNaN(parsed)
+      ? DEFAULT_SYNC_PUSH_ONLY_RECORD_THRESHOLD
+      : parsed;
+  } catch (err) {
+    logError(err);
+    return DEFAULT_SYNC_PUSH_ONLY_RECORD_THRESHOLD;
+  }
+}
+
 /*
  * See batches_limit in https://pouchdb.com/api.html#replication
  */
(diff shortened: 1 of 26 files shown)


ranisa-gupta16 (approved):
I had a play with this end-to-end and overall it feels good, sync modes switch the way I'd expect and the new banner copy is clear.


Image

Image


I tried the small notebook activation, small one came in as "Upload and download" with no snackbar, which is what I expected. Then I dropped the threshold to 1 in my .env, activated a bigger notebook (the one I've got with ~40 records) and saw the blue "Sync mode changed" snackbar pop up a witrh the explanation that it's been set to upload only to reduce device stress. Snackbar copy reads really nicely, no notes there.


Image

Once I went into the notebook, the orange "Large survey" warning banner at the top of My Sites was there too with the "GO TO SYNC SETTINGS" and "DISMISS" buttons.


Image

Played around with the sync mode dropdow,
switched between all three options (Sync off, Upload only, Upload and download). 

Image

Also tried activating with airplane mode on , defaultedd to Upload and download without breaking, which is the right safe fallback i guess when we can't reach the server to check the record count. 
Switched airplane mode back off, sync picked up normally a few seconds later.

Confirm dialog opens each time with the matching mode description. Switching to sync off shows the explanation that records will only be saved to this device.

Image
 
Switching to upload and download with 42 records shows the bandwidth/storage/battery note, which is the right warning for a large notebook. 
Help dialog (info icon next to Sync mode) wording matches what's in the confirm dialog, so no copy drift between them.
One question and confusion below that could be unrelated but worth flagging

Image

caught a runtime error in the Sync Status panel showing "Cannot read properties of undefined (reading 'pending')" while I had 11 of 13 notebooks activated in mixed sync modes. Screenshot below. Might be a quick null check fix but worth chasing before merge since it crashes the sync status reader.
 

 *** Very minor request, can we make this dialog's action buttons match the styling we've been using on the other recently-updated dialogs (red Cancel on left, black Confirm on right) for DASS/bssTheme

We've got centralised design tokens for dialog buttons now, so updating the values in one theme file should auto-apply across all three themes (BSS / Fieldmark / default). For BSS specifically I'd expect: red Cancel on left, black Confirm on right to maintain conssitency.

Image

Image 

like this 

Image

assuming the tests in projectsPersistMigration cover the actual rehydrate case so I'm calling that one good based on code review

Nothing here is blocking. Happy to approve.

PR #2119, opened 2026-06-12
fix: undo functionality in helper and advancedhelper text fields

ranisa-gupta16:
# feat/fix/chore: pr title

## JIRA Ticket

https://jira.csiro.au/browse/BSS-1295
https://jira.csiro.au/browse/BSS-1326

## Description

In the field editor, typing in Helper Text or Advanced Helper Text enabled the Undo button, but clicking Undo (or Redo) did nothing visible — the text on screen never reverted.

Helper Text → now uses DebouncedTextField bound to the store value (the same pattern the Label/Annotation fields already use), so it re-syncs on undo/redo. Removed the dead local-state/debounce code.
Advanced Helper Text → added an effect that pushes the store value back into the editor via setMarkdown only when the change is external (undo/redo/field-switch), tracked against the exact markdown the editor last emitted. Our own edits are recognized as echoes and skipped — so the caret is never disturbed, there's no setMarkdown↔onChange flicker, and the undo history / redo stack stay intact.

Screenshot 2026-06-12 153053

Screenshot 2026-06-12 153053

Screenshot 2026-06-12 152654

Screenshot 2026-06-12 152648

## Checklist


PeterBaker0 (approved):
Seems like an overall simplification of the logic, and seems to work as tested with both the built in undo/redo inside the editor, as well as the global undo/redo.

PR #2130, opened 2026-06-16
feat: limit map take point button label length

xinyuhoulucy:
# feat/limit-map-take-point-button-label-length

## JIRA Ticket

[BSS-1110](https://jira.csiro.au/browse/BSS-1110)

## Description

This PR adds a character limit to the `MapFormField` and the `TakePoint` button label text field.

## Proposed Changes

- Limits the saved button label text to a maximum number of characters, with a current default of 40 (up from 25). Please feel free to let me know if this is too long or too short.
- Shows an error message below the text field when the user exceeds the limit.
- Keeps the full typed value visible in the text field, while only saving the first 40 characters to Redux on change.

## How to Test

- Input text for `MapFormField` and `TakePoint` Button Label Text field:
   - When the input characters exceed the current maximum limit of 40, an error alert should be shown.
   - If the user saves at this time, only the first 40 characters will be saved. (when opening the same field next time, it will only show the saved text within the allowed length.)

image

## Additional Information

- I have not added any truncation or visual changes to the UI/app button label. The current 40-character limit looks good to me, but happy to adjust it if needed.

image

## Checklist


PeterBaker0 (changes requested):
Looks good - I like the custom hook. I think it would make more sense to have the hook require a max length argument, and have the default be managed by the caller context. So have the text length limits be in the calling components and make the argument required. 

The length limit would be good to consult with Elana on that. Once the above change is made we can have a quick chat with her.

web/src/designer/components/Fields/TakePointFieldEditor.tsx
```diff
@@ -13,6 +13,11 @@ import {withUpdatedField} from '../../features/fields/shared/updateField';
 import {fieldUpdated} from '../../store/slices/uiSpec';
 import {BaseFieldEditor} from './BaseFieldEditor';
 import {SimpleFieldWrapper} from './SimpleFieldWrapper';
+import {useTextFieldLengthLimit} from '@/hooks/use-input-char-limit';
+import { TakePointFieldProps } from '@faims3/forms';
```
PeterBaker0: formatting inconsistency - please run the ox formatter over it with ./format-lint.sh or similar.

web/src/designer/components/Fields/TakePointFieldEditor.tsx
```diff
@@ -23,7 +28,7 @@ export const TakePointFieldEditor = ({fieldName}: {fieldName: string}) => {
   );
   const dispatch = useAppDispatch();
 
-  const buttonLabelText = field['component-parameters'].buttonLabelText ?? '';
+  const buttonLabelText = (field['component-parameters'] as unknown as TakePointFieldProps).buttonLabelText ?? '';
```
PeterBaker0: This type hacking is necessary? Maybe so if the base component parameters is not sufficiently overlapping :/

PeterBaker0 (approved):
Looks good - please sort out formatting and ensure the typing is as good as can be

web/src/designer/components/Fields/TakePointFieldEditor.tsx
xinyuhoulucy: that’s a good command to know! thanks!

web/src/designer/components/Fields/TakePointFieldEditor.tsx
xinyuhoulucy: Yea,, hmm, on my side I was getting the type error for `buttonLabelText`, saying something like Type '{}' is not assignable to type 'string' :(

PR #2140, opened 2026-06-18
fix(designer): scope condition field picker to current form

### LINKED ISSUE #2087: Designer: menu of fields should only include those in current form
In the Notebook Editor's condition editor we present a menu of fields - these should just be from the current form but the current list include fields from all forms.

The current list is derived by just grabbing all fields in the current notebook.   Need to filter by the current form to get the appropriate list.   

Note that the TemplatedStringField visual builder does the right thing here by offering only the fields from the current form as part of the template.  May be able to share code here or at least see how to solve this problem.

jayenashar:
# fix: scope condition field picker to current form

## JIRA Ticket

N/A (open-source contribution). Tracked in GitHub: Closes #2087

## Description

In the Notebook Editor (Designer), the condition/visibility field-picker listed fields from every form in the notebook. Conditions are evaluated against a single record's values (one form), so fields from other forms are never in scope and would be non-functional options. This scopes the picker to the current form.

## Proposed Changes

- Scope the field-picker in `FieldConditionControl` to the current form, resolved from the editor context (`props.field` for field-level conditions, `props.view` for section-level).
- When the context cannot be resolved to a form, return no options instead of falling back to all notebook fields; guard against a missing section definition.
- Applies to both field-level and section-level conditions.
- Add regression tests covering both contexts.

## How to Test

1. Open the Designer for a notebook with two or more forms, each with distinct fields.
2. On a field in form A, choose **Add condition** and open the **Field** dropdown: it lists only form A's fields; form B's fields do not appear.
3. On a section, choose **Add conditions to section**: only the form's other-section fields appear.
4. Automated: `web/src/designer/components/condition.test.tsx` covers both contexts (run the `web` test suite).

## Additional Information

Section conditions on a single-section form now show an empty picker: they exclude their own section's fields, and there are no other sections to reference. This is pre-existing behaviour surfaced by the fix. See #2087 for a suggested follow-up (empty-state hint).

## Checklist

- [x] I have added JSDoc style comments to any new functions or classes. (No new functions/classes; an existing component was updated.)
- [x] Relevant documentation such as READMEs, guides, and class comments are updated. (No relevant documentation affected.)

### DIFF
--- a/web/src/designer/components/condition.test.tsx
+++ b/web/src/designer/components/condition.test.tsx
@@ -17,7 +17,7 @@
  */
 
 import {vi, describe, expect, test} from 'vitest';
-import {act, fireEvent, render, screen} from '@testing-library/react';
+import {act, fireEvent, render, screen, within} from '@testing-library/react';
 import {ConditionControl} from './condition/ConditionControl';
 import {ConditionType} from '../types/condition';
 import {sampleNotebook} from '../test-notebook';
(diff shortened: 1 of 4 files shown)


stevecassidy:
Good catch on the single-section form issue, I think a suitable message in this case would be most useful.

jayenashar:
> Good catch on the single-section form issue, I think a suitable message in this case would be most useful.

i would have preferred to hinder/prevent the user from clicking "Add condition" but that seemed too hard so i simply changed the noOptionsText

stevecassidy (approved):
Made a change to the way that the single section/field message is shown. Otherwise good. Thanks.

PR #2141, opened 2026-06-18
feat: add ComputedField field type 

Caedous:
# feat: add ComputedField field type

## Description
Adds a ComputedField that derives a numeric value from an expression over other
numeric fields in the same form. Similar in scope to TemplatedStringField, but
evaluates an arithmetic/conditional expression instead of a Mustache template.

## Proposed Changes
- Expression engine (expressionEngine.ts) built on jsep with a whitelist
  evaluator: only literals, identifiers, unary/binary operators and the ternary
  are executed; function calls, member access and indexing are rejected at
  compile time, so there is no eval and no injection surface. Parsed once and
  cached by source.
- computedFields.ts recomputes computed fields at the form-manager level
  (recompute-all-and-diff), mirroring templatedFields.ts. Wired into both
  EditableFormManager (on save) and PreviewFormManager (live on change), run
  before the templated recompute so templates can read fresh values.
- Read-only ComputedField component + spec; returns faims-core::Number, stores
  number | null, blank until all referenced fields have numeric values.
- Registered in the field registry; added to the designer chooser under Numbers;
  added a ComputedFieldEditor so authors can enter the expression.
- New dependencies: jsep, @jsep-plugin/ternary.

## Behaviour notes
- Expressions reference other fields by their exact Field ID (case-sensitive).
  IDs are used rather than labels so expressions survive label edits.
- Supports + - * / %, parentheses, unary minus, comparisons, && / ||, and the
  ternary operator. Result must be a finite number or the field shows blank.
- Referencing another derived field (computed or templated) yields blank, to
  avoid evaluation-ordering problems in this version.

## How to Test
1. In the designer, add two Number fields and one Computed Value field (Numbers
   category).
2. Set the Computed Value expression to an expression over the number fields'
   Field IDs, e.g. `Width * Height`.
3. Preview the form (or create a record): the computed field stays blank until
   both inputs have values, then updates as they change.
4. Try `(value1 + value2) / 2`, `-depth * 12`, and a conditional such as
   `Width > 0 ? Width * Height : 0`.

## Testing
- pnpm build --filter=@faims3/forms and --filter=@faims3/web: clean.
- pnpm --filter @faims3/forms test: 51 passing, incl. ticket examples,
  conditionals, caching, blank-on-missing, and compile-time rejection of unsafe
  expressions.
- Manual: verified live in the form preview — Width * Height computes as field
  values are entered, stays blank until both are present, keys off field IDs.

## Screenshots
**Designer — field in chooser with expression editor**
computed value 1

**Designer — Area = Width * Height configured**
computed value 2

**Preview — live computation (5 × 5 = 25)**
computed value 3

## Follow-ups
- Surface available field IDs in the expression editor (picker/reference list),
  mirroring the templated-string visual builder, to avoid guessing exact IDs.
- Dependency-ordered evaluation so a computed field can reference another.
- Optional string-returning expressions (needs a returns-type decision).

## Checklist
- [x] Tests have been added or updated to cover the changes.


Caedous:
I havent finished fully testing the expressionEngine yet, I've just made this push, to be able to continue work on another machine. Will mark this PR for review once it's ready.

library/forms/lib/formModule/formManagers/computedFields.ts
```diff
@@ -0,0 +1,167 @@
...
+    return Number.isNaN(n) ? null : n;
+  };
+
+  let changes = false;
+  const updates: Record<string, number | null> = {};
+
+  for (const {fieldName, expression} of computedFields) {
+    let result: number | null = null;
+    try {
+      const compiled = compileExpression(expression);
```
stevecassidy: We shouldn't need to compile the expression every time we evaluate, look at how conditions are handled by `compileUiSpecConditionals` when the uiSpec is loaded. Maybe we can do something similar so that here we just need to call a compiled function to evaluate the expression against the values.

library/forms/lib/formModule/formManagers/expressionEngine.ts
```diff
@@ -0,0 +1,183 @@
...
+  } catch (e) {
+    throw new ExpressionError((e as Error).message);
+  }
+
+  const refs = new Set<string>();
+  validate(ast, refs);
+
+  const evaluate = (scope: Map<string, number>): number | null => {
+    try {
+      const result = evalNode(ast as unknown as EvalNode, scope);
```
stevecassidy: This approach requires parsing chasing the expression every time we want to evaluate it with lots of string comparisons.  Instead, build a function to evaluate this expression and return that.   Look at how conditional expressions are compiled for inspiration. Should be able to make this a lot more efficient.

web/src/designer/components/Fields/ComputedFieldEditor.tsx
```diff
@@ -0,0 +1,92 @@
...
+                value={state.helperText || ''}
+                helperText="Help text shown along with the field"
+                onChange={e => updateProperty('helperText', e.target.value)}
+              />
+            </Grid>
+          </Grid>
+        </Card>
+      </Grid>
+
+      <Grid size={12}>
```
stevecassidy: This is workable but I think some kind of expression builder like the templated field string editor would be very useful - at least for insertion of field names.

stevecassidy (changes requested):
Looks good but a few things to work on.

library/forms/lib/formModule/formManagers/computedFields.ts
Caedous: Done. Expressions are now compiled once when the notebook loads (in `compileUiSpecConditionals`) and the compiled function is stored on the field. `recomputeComputedFields` just calls it now instead of compiling each time.

library/forms/lib/formModule/formManagers/expressionEngine.ts
Caedous: Done. The engine now compiles each expression into a function once instead of walking the AST on every evaluation. It's moved into data-model next to the conditionals compiler, since that's where compilation happens at load.

web/src/designer/components/Fields/ComputedFieldEditor.tsx
Caedous: Added. The editor now shows the other fields as chips below the expression box; clicking one inserts its field ID. Kept it simple rather than a full builder since the expression is just one line.

computedfield 4

PeterBaker0: How well does this scale up to many fields - we have forms with dozens of fields like 40-50? Would a search be a better option. If you build a nice fuzzy search component over fields, you could update the condition builder to use it at the same time. I like this library and I think we may already use it https://github.com/farzher/fuzzysort - maybe version pin since supply chain is not super secure for a smaller package like that.

PeterBaker0: @Caedous

PeterBaker0: https://github.com/leeoniya/uFuzzy could also check out this lib

web/src/designer/components/Fields/ComputedFieldEditor.tsx
```diff
@@ -0,0 +1,150 @@
...
+                  color="text.secondary"
+                  sx={{display: 'block', mb: 1}}
+                >
+                  Insert a field ID:
+                </Typography>
+                <Box sx={{display: 'flex', flexWrap: 'wrap', gap: 1}}>
+                  {formFieldIds.map(id => {
+                    const label =
+                      (
+                        allFields[id]?.['component-parameters'] as
```
stevecassidy: Check the return type of the field and only include number fields in this list.

web/src/designer/components/Fields/ComputedFieldEditor.tsx
stevecassidy: Nice idea, we'll leave that for a future PR.

web/src/designer/components/Fields/ComputedFieldEditor.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  // Appends a field ID to the expression, space-separated.
+  const insertFieldId = (id: string) => {
+    const current = state.expression || '';
+    const next = current === '' ? id : `${current} ${id}`;
+    updateProperty('expression', next);
+  };
+
+  return (
+    <Grid container spacing={2}>
```
stevecassidy: Use BaseFieldEditor to get the standard field editor components. See eg. ControlledNumberFieldEditor for an example.

library/data-model/src/uiSpecification/expressions.ts
```diff
@@ -0,0 +1,182 @@
...
+
+// A node compiled into a closure over the scope. Compilation does the type and
+// operator dispatch once; evaluation is just a function call.
+type NodeEvaluator = (scope: Map<string, number>) => number;
+
+const cache = new Map<string, CompiledExpression>();
+
+// Binary operator implementations. Comparisons and logical operators return
+// 1/0 so they compose with arithmetic. Selected once at compile time.
+const binaryOps: {[op: string]: (l: number, r: number) => number} = {
```
stevecassidy, line 62: I found a problem - field names can contain - (Wet-Soil-Mass-g) which is being parsed as subtraction. We'll need to quote identifiers in some way to allow parsing expressions.   Maybe {identifier} which is what we have for templated string field.

stevecassidy (changes requested):
Passing this back with a few comments and one big issue - hyphens in field names.

library/data-model/src/uiSpecification/expressions.ts
Caedous: Fixed. References are now wrapped in braces, e.g. `{Wet-Soil-Mass-g} * {Number-of-Samples}`, so the field ID's hyphens are treated as part of the reference rather than subtraction. The engine swaps each {...} for a safe placeholder before parsing and maps it back afterwards; unbraced references and malformed braces are rejected at compile time. Screenshot shows a hyphenated field computing correctly (4 × 2 = 8).

computedfield value 1
computedfield value 3

web/src/designer/components/Fields/ComputedFieldEditor.tsx
Caedous: Done. The editor now uses BaseFieldEditor with the expression input and field chips as children, matching ControlledNumberFieldEditor.

computedfield value 2
computedfield value 5

web/src/designer/components/Fields/ComputedFieldEditor.tsx
Caedous: Done. The chip list is now filtered to numeric fields only. In the screenshot the form also has a Text field (Location) which is correctly excluded from the list.

computedfield value 5
computedfield value 6

web/src/designer/components/Fields/ComputedFieldEditor.tsx
Caedous: Yep. Apologies for the delay in replying to this, will definitely look into this further and assess the best option.

stevecassidy (approved):
Very nicely done. Thanks.

Issue #1594, opened 2025-08-02
Enable notebook and team storage stats


stevecassidy:
We can use the /{db} route on CouchDB (not reflected in PouchDB) to get the size of any database. Use this to implement a per-notebook size statistic and consequently a per-team size statistic. 

Eventually we may want to be able to limit activity based on this size (in the commercial offering) but in the first instance we can just display this in the overview of the notebook or team.

jayenashar:
do i need to worry about projects with no team?

stevecassidy:
I guess you could compute the storage stats for the project but they wouldn't then be assigned to a team.

jayenashar:
for a commercial offering, how does the project/team storage stats feed into the customer's stats?  i'm assuming a multi-tenancy system for this.  if it's multiple single-tenant systems then could the storage limit be placed at the couchdb instance level and then we wouldn't need to calculate per-project?

i have the per-project calculation done and will aggregate it for a team next.

stevecassidy:
It's both multi-tenancy and per-enterprise servers, but even for enterprise deployments it will be useful to have per-team usage statistics.

PR #2146, opened 2026-06-21
Feat: enable notebook and team storage stats

jayenashar:
# feat: enable notebook and team storage stats

## JIRA Ticket

#1594

## Description

Add some usage statistics to the Team page in web/ and extend the simple record count in notebook pages.

## Proposed Changes

per-notebook size statistic and consequently a per-team size statistic

## How to Test

- Open a project page, details tab, see the new Current byte count section
- Open a team page, details tab, see the new projects section

## Additional Information

- `getUserProjectsDetailed` now has a Promise.all.  not sure if there will be problems when a team has a lot of projects.
- moved `statusDisplay` and added JSDoc

## Checklist

### DIFF
--- a/api/public/swagger.json
+++ b/api/public/swagger.json
@@ -2086,6 +2086,11 @@
             "is_admin": {
               "type": "boolean",
               "description": "True when the caller has project admin on this notebook"
+            },
+            "byteCount": {
+              "type": "integer",
+              "minimum": 0,
+              "description": "Optional size of the notebook’s data in bytes (projects DB v4+)"
             }
           },
           "required": ["is_admin"]
--- a/api/public/swagger.json
+++ b/api/public/swagger.json
@@ -2136,13 +2136,17 @@
       ]
     },
     "GetNotebookResponse": {
-      "description": "GET /notebooks/{id} — project document plus optional record count.",
+      "description": "GET /notebooks/{id} — project document plus optional record count and byte count.",
       "allOf": [
         {"$ref": "#/definitions/ProjectDocument"},
         {
           "type": "object",
           "properties": {
             "recordCount": {
+              "type": "integer",
+              "minimum": 0
+            },
+            "byteCount": {
               "type": "integer",
               "minimum": 0
             }
--- a/api/public/swagger.json
+++ b/api/public/swagger.json
@@ -2090,7 +2090,7 @@
             "byteCount": {
               "type": "integer",
               "minimum": 0,
-              "description": "Optional size of the notebook’s data in bytes (projects DB v4+)"
+              "description": "Size of the notebook’s data in bytes (projects DB v4+)"
             }
           },
           "required": ["is_admin"]
--- a/api/src/couchdb/notebooks.ts
+++ b/api/src/couchdb/notebooks.ts
@@ -687,6 +687,23 @@ export async function countRecordsInNotebook(
   }
 }
 
+export async function getByteCount(
+    project_id: ProjectID
+): Promise<number> {
+  const dataDB = await getDataDB(project_id);
+  // I can see what I want in sizes.active in CouchDB Fauxton, but it's not clear how to get it via the PouchDB API. This view is an approximation that sums the sizes of all attachments, which is likely to be the dominant contributor to database size for most projects.
+  try {
+    const res = await dataDB.query('index/attachmentSizeSum');
+    if (res.rows.length === 0) {
+      return 0;
+    }
+    return res.rows[0].value;
+  } catch (error) {
+    console.log(error);
+    return 0;
+  }
+}
+
 /*
  * For saving and loading attachment with type faims-attachment::Files
  */
--- a/api/src/couchdb/notebooks.ts
+++ b/api/src/couchdb/notebooks.ts
@@ -61,7 +61,12 @@ import {
   CompiledNotebookUiSpec,
   compileUiSpecConditionals,
 } from '@faims3/data-model';
-import {initialiseDataDb, localGetProjectsDb, verifyCouchDBConnection} from '.';
+import {
+  getNanoDataDb,
+  initialiseDataDb,
+  localGetProjectsDb,
+  verifyCouchDBConnection,
+} from '.';
 import {COUCHDB_PUBLIC_URL, MIGRATE_NOTEBOOKS_ON_STARTUP} from '../buildconfig';
 import * as Exceptions from '../exceptions';
 import {userCanDo} from '../middleware';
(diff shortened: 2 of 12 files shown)


web/src/components/tabs/project/details.tsx
```diff
@@ -118,6 +85,12 @@ const ProjectDetails = ({projectId}: {projectId: string}) => {
         label: 'Current record count',
         getValue: p => (p.recordCount != null ? String(p.recordCount) : '—'),
       },
+      {
+          key: 'byteCount',
+          label: 'Current byte count',
+          getValue: p => (p.byteCount != null ? `${p.byteCount} bytes (${formatFileSize(p.byteCount)})` : '—'),
```
stevecassidy: I think reporting just the formatted size is best, no-one really needs to know how many bytes.

web/src/hooks/queries.ts
```diff
@@ -199,14 +199,16 @@ export const useTemplateSurveyReferences = ({
...
   teamId,
   user,
+  includeArchived = false,
 }: {
   user: User | null;
   teamId: string;
+  includeArchived?: boolean;
 }) =>
   useQuery({
     queryKey: ['projectsbyteam', user?.token, teamId],
```
jayenashar: The `queryKey` `['projectsbyteam', user?.token, teamId]` does not include `includeArchived`. `team-details.tsx` calls this with `includeArchived: true` while `team-projects.tsx` uses the default `false` for the same `teamId`, so both share one cache entry. Whichever query loads first populates the cache and the other silently serves the wrong (filtered or unfiltered) result. Add `includeArchived` to the queryKey.

library/data-model/src/utils.ts
```diff
@@ -28,6 +28,17 @@ export const slugify = (str: string) => {
...
 
+/**
+ * Formats file size in human-readable format
+ */
+export function formatFileSize(bytes: number): string {
+  if (bytes === 0) return '0 Bytes';
+  const k = 1024;
+  const sizes = ['Bytes', 'KB', 'MB', 'GB'];
+  const i = Math.floor(Math.log(bytes) / Math.log(k));
+  return Math.round((bytes / Math.pow(k, i)) * 100) / 100 + ' ' + sizes[i];
```
jayenashar, line 44: `formatFileSize` has two unhandled edge cases now that it runs on network-supplied `byteCount` values:
- Negative or `NaN` input: `Math.log` yields `NaN`/`-Infinity`, so `sizes[i]` is `undefined` and the output is `"NaN undefined"`.
- `bytes >= 1024^4` (1 TB): `i` becomes 4 but `sizes` only has indices 0-3, so the output is `"X.XX undefined"`.

Suggest guarding non-finite/negative input and either extending `sizes` (TB, PB) or clamping `i` to `sizes.length - 1`.

api/src/couchdb/notebooks.ts
```diff
@@ -687,6 +693,27 @@ export async function countRecordsInNotebook(
...
+ */
+export async function getByteCount(
+  project_id: ProjectID
+): Promise<number> {
+  try {
+    const dataDb = await getNanoDataDb(project_id);
+    const info = await dataDb.info();
+    return info.sizes.active;
+  } catch (error) {
+    console.log(error);
```
jayenashar: Catching every error, logging via `console.log`, and returning `0` makes an I/O or connectivity failure indistinguishable from a genuinely empty database, so the UI shows "0 bytes" for a project that actually errored. Suggest `console.error` (so it surfaces in logs) and returning `undefined` propagated through the schema, so callers can tell "unknown" apart from "empty".

library/data-model/src/api.ts
```diff
@@ -245,12 +245,14 @@ export type PostExchangeTokenResponse = z.infer<
 /** GET /api/notebooks list row (project summary + access flags; no uiSpecification). */
 export const APINotebookListSchema = ProjectListItemSchema.extend({
   is_admin: z.boolean(),
+  byteCount: z.number(),
```
jayenashar, line 248: `byteCount` is declared `z.number()` (required) here and on `GetNotebookResponseSchema` (line 255), but the web UI treats it as optional (`p.byteCount ?? 0` in team-details, `p.byteCount != null` in project details). If it can be absent (pre-v4 projects DB, or an errored size lookup), declare it `z.number().optional()` like `recordCount`; otherwise the `?? 0` / `!= null` guards are dead code. Pick one so the contract and usage agree.

api/src/couchdb/notebooks.ts
```diff
@@ -283,7 +288,7 @@ export const getUserProjectsDetailed = async (
       });
     });
 
-  return userProjects.map(project => {
+  return await Promise.all(userProjects.map(async project => {
```
jayenashar: This `Promise.all` fans out one CouchDB `info()` call per project (an N+1 over the user's notebooks), on a path that runs for every notebook-list render. For a team or user with many notebooks this adds N concurrent round-trips and can pressure CouchDB connection limits. Consider a bounded-concurrency map (e.g. p-limit) or batching, or at least documenting the trade-off. (Noted in the PR description as a concern.)

web/src/components/tabs/teams/team-details.tsx
```diff
@@ -3,9 +3,12 @@ import {ListItem, ListLabel, ListDescription} from '@/components/ui/list';
 import {Skeleton} from '@/components/ui/skeleton';
 import {List} from '@/components/ui/list';
 import {Card} from '@/components/ui/card';
-import {useGetTeam} from '@/hooks/queries';
+import {useGetProjects, useGetProjectsForTeam, useGetTeam} from '@/hooks/queries';
```
jayenashar: `useGetProjects` is imported here but never used in this file. Remove it to avoid an unused-import lint error.

web/src/components/tabs/teams/team-details.tsx
```diff
@@ -56,6 +59,38 @@ const TeamDetails = ({teamId}: {teamId: string}) => {
...
+              {projects ? `${projects.length} - ${formatFileSize(projects.reduce((acc, p) => acc + (p.byteCount ?? 0), 0))}` : 'Unknown...'}
+                </summary>
+              <List className="pt-2 pl-4">
+                {[ProjectStatus.OPEN, ProjectStatus.ARCHIVED, ProjectStatus.CLOSED].map(status => {
+                  const projectsWithStatus = projects ? projects.filter(p => p.status === status) : [];
+                  const byteCount = projectsWithStatus.reduce((acc, p) => acc + (p.byteCount ?? 0), 0);
+                  return (
+                    <ListItem key={status}>
+                      <ListLabel>{statusDisplay(status)}</ListLabel>
+                      {isProjectsPending ? (
```
jayenashar: This inner `isProjectsPending` ternary is dead code: it is already inside the `false` branch of the outer `isProjectsPending` check (line 64), so the `<Skeleton/>` arm can never render here. Render the `ListDescription` unconditionally.

web/src/components/tabs/project/details.tsx
```diff
@@ -118,6 +85,12 @@ const ProjectDetails = ({projectId}: {projectId: string}) => {
         label: 'Current record count',
         getValue: p => (p.recordCount != null ? String(p.recordCount) : '—'),
       },
+      {
+          key: 'byteCount',
```
jayenashar: This new `byteCount` row is indented with extra spaces relative to the sibling `recordCount` entry above it. Align to the surrounding 2-space indentation to match the file's style.

api/public/swagger.json
```diff
@@ -2086,6 +2086,11 @@
             "is_admin": {
               "type": "boolean",
               "description": "True when the caller has project admin on this notebook"
+            },
+            "byteCount": {
+              "type": "integer",
+              "minimum": 0,
+              "description": "Optional size of the notebook’s data in bytes (projects DB v4+)"
```
jayenashar: The description marks `byteCount` as "Optional" and it is absent from this object's `required` array, but the Zod schema (`APINotebookListSchema` in library/data-model/src/api.ts) declares it required. Align the swagger spec and the schema so the documented contract matches.

stevecassidy (approved):
This looks good for a start on this feature, one minor comment and this is good to merge.

api/src/couchdb/notebooks.ts
jayenashar: opted to make it -1 because i just made it required a few mins ago

jayenashar:
made some updates.  mainly looking for a review of the tests

stevecassidy (approved):
Looks good, very happy with this. Thanks.

PR #2147, opened 2026-06-21
chore: add watch option for libs and use it in localdev.sh

jayenashar:
# chore: add watch option for libs and use it in localdev.sh

## JIRA Ticket

none

## Description

Currently running localdev.sh doesn't rebuild the library modules on changes.

Also when localdev.sh is run it wipes any local changes in .env that might be being tested.

## Proposed Changes

- exposed dev libs to docker
- added watcher scripts
- run watcher scripts from localdev.sh
- don't overwrite .env files if they already exist to preserve local changes.

## How to Test

`./localdev.sh --all` and change a data model and be able to use it in app, web, and api without manually rebuilding

## Checklist


stevecassidy (approved):
Thanks, I added another little tweak that has annoyed me in the past with localdev.sh.

PR #2150, opened 2026-06-22
Fix: duplicate dialog not showing when field is collapsed

stevecassidy:
# Fix: duplicate dialog not showing when field is collapsed

## Ticket

None

## Description

When a field is collapsed, clicking the duplicate button doesn't do anything.

## Proposed Changes

Remove stopPropagation which results in the field opening up and then showing the dialog.

## How to Test

Click the duplicate field button when a field is closed.

## Checklist


web/src/designer/components/field-editor.tsx
```diff
@@ -267,7 +267,11 @@ const FieldEditorComponent = ({
   };
 
   const handleOpenDuplicateDialog = (event: React.SyntheticEvent) => {
-    event.stopPropagation();
+    // if the field is collapsed, we need to open it up so that the
+    // dialog will show. If it is open already, we don't want to do anything
```
jayenashar: i think this comment makes sense in the context of the issue, but without that context, the inverse makes more sense.  going forward the issue will not present itself, so for a new person looking at the code, i think the comment could be "we want the final state to be expanded + dialog.  propagating while expanded means it will collapse, so we stop the collapse so the final state is always expanded"

jayenashar (approved):

web/src/designer/components/field-editor.tsx
jayenashar: maybe the better way of phrasing that: i see the current comment as "why the conditional" but i think it could be "why stop propagation"

PR #2151, opened 2026-06-22
A component to show record history (audit trail)

jayenashar:
# feat: A component to show record history (audit trail)

## JIRA Ticket

FAIMS/GroundSight-FAIMS#15

## Description

A component to show record history (audit trail)

## Proposed Changes

- added a component to show record history
- added a component to show revision metadata

## How to Test

view a record and click on the new history tab.

## Additional Information

unpolished.  will be updated.

## Checklist


app/src/gui/pages/viewRecord.tsx
```diff
@@ -356,6 +357,139 @@ const ViewTabContent: React.FC<ViewTabContentProps> = ({
...
+/**
+ * Content for the History tab - displays revision history and metadata
+ * @param props.recordId - ID of the record to fetch history for
+ * @param props.dataEngine - DataEngine instance for fetching data
+ */
+const HistoryTabContent: React.FC<{
+  recordId: RecordID;
+  dataEngine: DataEngine;
+}> = props => {
+  const {recordId, dataEngine} = props;
```
stevecassidy: minor style note, use destructuring in function argument ({recordId, dataEngine})

app/src/gui/pages/viewRecord.tsx
```diff
@@ -356,6 +357,139 @@ const ViewTabContent: React.FC<ViewTabContentProps> = ({
...
+    return (
+      <Box sx={{p: 2}}>
+        <Typography color="error">Revision data not found.</Typography>
+      </Box>
+    );
+  }
+
+  return (
+    <Stack spacing={2}>
+      <Typography variant="h6">Revision ID: {revisionId}</Typography>
```
stevecassidy: Revision id is not really a useful human readable value so let's just use the created date/created by as a header, eg. 'Revision created by user@here.com at 24/06/2026 15:28:33`.   See formatTimeStamp for time formatting or maybe use relative time (https://polypane.app/blog/the-intl-api-the-best-browser-api-youre-not-using/#intlrelativetimeformat). So then we could have 'Revision created by user@here.com yesterday'.

stevecassidy (changes requested):
Some minor changes but looks ok.

jayenashar:
sorry i should not have assigned you as reviewer as i wasn't happy with it yet and i didn't review it myself.

this is the current state:

Screen Shot 2026-06-24 at 20 07 32

will be showing the field names next.  then changing it from parents[0] to look like a graph for multi-parent revisions.

app/src/gui/pages/viewRecord.tsx
jayenashar: used formatTimeStamp though imho toLocaleDateString looks better.  didn't use relativetimeformat because i'd have to calculate the difference manually.  could i have used momentjs?  i see it in app/package.json.  any plan to move away from momentjs?

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -1407,6 +1408,73 @@ class FormOperations {
...
+          context: {revision},
+        } = await this.getExistingFormData({recordId, revisionId});
+        return revision;
+      })
+    );
+
+    // Index by revision ID so each revision can look up its parent's AVPs.
+    const revisionsById = new Map(revisions.map(rev => [rev._id, rev]));
+
+    return revisions.map((revision): RevisionHistoryEntry => {
```
jayenashar: bug: `record.revisions` is stored lexicographically sorted by `frev-<uuid>` (see updateRecordHeads at engine.ts:1961), so this returns the audit trail in effectively random order, not chronological. Sort the result by `created` (e.g. newest first) before returning.

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -1407,6 +1408,73 @@ class FormOperations {
...
+    recordId: string;
+  }): Promise<RevisionHistoryEntry[]> {
+    const {
+      context: {record},
+    } = await this.getExistingFormData({recordId});
+
+    // Hydrate every revision once so we can report authorship and diff each
+    // revision's AVP map against its parent to find the fields that changed.
+    const revisions = await Promise.all(
+      record.revisions.map(async revisionId => {
```
jayenashar: performance: `getExistingFormData` fully hydrates every AVP (and re-fetches the record doc) for each revision, so a record with N revisions costs O(N) full hydrations. History only needs `avps`, `parents`, `created`, `createdBy` -- use `this.core.getRevision(revisionId)` here (and `this.core.getRecord` above) instead.

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -1407,6 +1408,73 @@ class FormOperations {
...
+    const revisionsById = new Map(revisions.map(rev => [rev._id, rev]));
+
+    return revisions.map((revision): RevisionHistoryEntry => {
+      const parentsIds = revision.parents.length
+        ? revision.parents
+        : [undefined]; // Treat no parents as a single parent so all fields are considered new from 1 parent
+      return {
+        revisionId: revision._id,
+        created: revision.created,
+        createdBy: revision.createdBy,
```
jayenashar: important: `revision.createdBy` is the author of the revision document, which `updateRevision` preserves from the original (engine.ts:1348) even when later in-place edits pass a different `updatedBy`. Per-field authorship lives on the AVPs; consider surfacing that for an accurate audit trail.

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -1407,6 +1408,73 @@ class FormOperations {
...
+        changedFields: Object.fromEntries(
+          parentsIds.map((parentId) => {
+            const parent = parentId ? revisionsById.get(parentId) : undefined;
+            const parentAvps = parent?.avps ?? {};
+
+            const changed = new Set<string>();
+            for (const [field, avpId] of Object.entries(revision.avps)) {
```
jayenashar: important: this diff compares AVP ids only. AVPs updated in place (see avpsToUpdate in createOrReuseAvps, e.g. annotation-only edits) keep the same id, so such changes will not appear in `changedFields`. Worth documenting as a known limitation or comparing AVP contents.

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -1407,6 +1408,73 @@ class FormOperations {
...
+      })
+    );
+
+    // Index by revision ID so each revision can look up its parent's AVPs.
+    const revisionsById = new Map(revisions.map(rev => [rev._id, rev]));
+
+    return revisions.map((revision): RevisionHistoryEntry => {
+      const parentsIds = revision.parents.length
+        ? revision.parents
+        : [undefined]; // Treat no parents as a single parent so all fields are considered new from 1 parent
```
jayenashar: important: using `undefined` as the parent key means `Object.fromEntries` coerces it to the string "undefined", which leaks into the API contract and the zod schema. Use an explicit sentinel (e.g. "_root") for the no-parent case.

app/src/gui/pages/viewRecord.tsx
```diff
@@ -356,6 +361,100 @@ const ViewTabContent: React.FC<ViewTabContentProps> = ({
...
+      dataEngine.form.getHistoryData({
+        recordId,
+      }),
+    networkMode: 'always',
+    refetchOnMount: 'always',
+    staleTime: 0,
+    gcTime: 0,
+  });
+
+  if (isPending || isRefetching) {
```
jayenashar: important: with `gcTime: 0` / `staleTime: 0` / `refetchOnMount: always`, gating on `isRefetching` replaces the whole list with a full spinner on every revisit. Keep showing cached `historyData` with a subtler indicator, or drop `isRefetching` from this guard and raise `gcTime`.

app/src/gui/pages/viewRecord.tsx
```diff
@@ -356,6 +361,100 @@ const ViewTabContent: React.FC<ViewTabContentProps> = ({
...
+        const parentFields = Object.entries(entry.changedFields);
+        return (
+          <Stack key={entry.revisionId} spacing={2}>
+            <Typography variant="body1" id={entry.revisionId}>
+              Revision created by <span style={{textDecoration: 'underline'}}>{formatRevisionMetadata(entry)}</span>
+            </Typography>
+            <Stack sx={{pl: 2}}>
+              {parentFields.map(([parentId, fields]) => 
+                <Typography variant="body1" key={parentId}>
+                  Fields changed{parentFields.length > 1 ? <> compared to <Link href={`#${parentId}`}>{formatRevisionMetadata(revisionIdsRevision.get(parentId))}</Link></> : ''}: {fields.map(
```
jayenashar: bug: when `fields` is empty (a revision whose AVP ids match this parent, e.g. one side of a clean merge), this renders "Fields changed: " with nothing after the colon. Skip entries with no changed fields, or show an explicit "None".

app/src/gui/pages/viewRecord.tsx
```diff
@@ -64,13 +66,16 @@ import RecordMeta from '../components/record/meta';
 import UGCReport from '../components/record/UGCReport';
 import BackButton from '../components/ui/BackButton';
 import {theme} from '../themes';
+import { formatTimestamp } from '../../utils/formUtilities';
+import { Tag } from '@mui/icons-material';
```
jayenashar: cleanup: `Tag` is imported but never used; remove it to satisfy lint/CI.

library/data-model/tests/engine.form.test.ts
```diff
@@ -1212,4 +1214,36 @@ describe('Form Operations', () => {
       expect(formData.data['Fourth-1'].attachments).toHaveLength(1);
     });
   });
+
+  describe('getHistoryData', () => {
+    test('should return revision authorship and changed fields', async () => {
```
jayenashar, line 1225: important: this only covers a single root revision. The core new logic -- multi-parent merge revisions, per-parent `changedFields`, and ordering -- is untested. Add cases for a merge revision with two parents and for history ordering.

api/src/scripts/seedTestDataset.ts
```diff
@@ -428,6 +567,14 @@ const main = async () => {
       console.log(`  ✓ ${spec.email} (${spec.name})`);
     }
 
+    // ── Phase 6: Records with merge history ─────────────────────────────────────
+    console.log('\nPhase 6: Seeding a record with a merge revision...');
```
jayenashar: cleanup: Phase 6 is auxiliary demo data but runs unguarded; if `seedRecordWithMergeRevision` throws (e.g. no form has two editable fields), the entire seed aborts. Wrap in try/catch and warn, or skip gracefully.

api/src/scripts/seedTestDataset.ts
```diff
@@ -240,6 +247,138 @@ function clampDescription(description?: string): string | undefined {
...
+  fieldB: string;
+} {
+  for (const [formId, viewset] of Object.entries(uiSpec.viewsets)) {
+    const fieldIds = viewset.views.flatMap(view => uiSpec.views[view].fields);
+    // Most-preferred first: plain text inputs, then any string-valued field,
+    // then anything. De-duplicate while preserving that preference order.
+    const ordered = Array.from(
+      new Set([
+        ...fieldIds.filter(
+          id => uiSpec.fields[id]['component-name'] === 'TextField'
```
jayenashar: cleanup: `uiSpec.fields[id]` can be undefined if a view references a field id not present in `fields`, which would throw here and abort seeding. Guard with optional chaining before reading `component-name` / `type-returned`.

library/data-model/src/databaseEngine/engine.ts
jayenashar: more of a display thing to me, so sorted on the frontend.

library/data-model/src/databaseEngine/engine.ts
jayenashar: ok that was a really hacky way to do it.

app/src/gui/pages/viewRecord.tsx
jayenashar: good catch.  i never meant to import that.

library/data-model/tests/engine.form.test.ts
jayenashar: was only intended as a basic test, not exhaustive.  decided to order on frontend not in engine.

api/src/scripts/seedTestDataset.ts
jayenashar: i think a throw is ok because then someone will fix the seed

api/src/scripts/seedTestDataset.ts
jayenashar: abort seeding is ok i think, because then someone will fix the seed.

library/data-model/src/databaseEngine/engine.ts
jayenashar: i'm pretty sure we don't have in-place edits.  also pretty sure authorship lives on the revision and there is no per-field authorship.

library/data-model/src/databaseEngine/engine.ts
jayenashar: if an avp could be updated in place, then the same id would still point to the same edited avp.  if there's an annotation on a revision that doesn't create a new revision, then i think it's ok to ignore those.

app/src/gui/pages/viewRecord.tsx
jayenashar: dropped isRefetching+staleTime+gcTime

library/data-model/src/databaseEngine/engine.ts
jayenashar: well that's a million times better now

app/src/gui/pages/viewRecord.tsx
```diff
@@ -356,6 +360,100 @@ const ViewTabContent: React.FC<ViewTabContentProps> = ({
...
+  ]));
+
+  const formatRevisionMetadata = (entry?: RevisionHistoryEntry) => {
+    return entry ? `${entry.createdBy} at ${formatTimestamp(new Date(entry.created).getTime())}` : 'unknown';
+   }
+
+  return (
+    <Stack spacing={4}>
+      <Typography variant="h5">Revision History</Typography>
+      {historyData.sort((a, b) => b.created.localeCompare(a.created)).map(entry => {
```
jayenashar: `historyData.sort(...)` sorts in place, mutating the array returned by `useQuery` (the React Query cache object). Mutating cached data during render is a render-phase side effect and can reorder the array seen by later renders or other consumers. Use `[...historyData].sort(...)` or `.toSorted(...)`. (Flagged independently by all three reviewers.)

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -1407,6 +1408,65 @@ class FormOperations {
...
+        createdBy: revision.created_by,
+        changedFields:
+          revision.parents.length === 0
+            ? { root: Object.keys(revision.avps).sort() }
+            : Object.fromEntries(
+                revision.parents.map((parentId) => {
+                  const parent = parentId
+                    ? revisionsById.get(parentId)
+                    : undefined;
+                  const parentAvps = parent?.avps ?? {};
```
jayenashar, line 1472: When a parent revision id is not found in `revisionsById`, `parentAvps` silently falls back to `{}`, so every field in this revision is reported as changed. That yields a misleading diff rather than surfacing incomplete history. Consider fetching the missing parent or signalling the gap instead of treating it as a full-field change.

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -1407,6 +1408,65 @@ class FormOperations {
...
+   * @returns Array of revision history entries (revisionId, created,
+   * createdBy, changedFields)
+   */
+  async getHistoryData({
+    recordId,
+  }: {
+    recordId: string;
+  }): Promise<RevisionHistoryEntry[]> {
+    const record = await this.core.getRecord(recordId);
+    const revisions = await Promise.all(
```
jayenashar, line 1453: This fires one `getRevision` per revision in parallel with no batching or cap. For records with long edit histories that is N round-trips and scales poorly. Consider a bulk fetch (e.g. `allDocs` by revision id) or pagination. (Flagged by all three reviewers.)

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -1407,6 +1408,65 @@ class FormOperations {
...
+      record.revisions.map((revisionId) => this.core.getRevision(revisionId)),
+    );
+
+    // Index by revision ID so each revision can look up its parent's AVPs.
+    const revisionsById = new Map(revisions.map(rev => [rev._id, rev]));
+
+    return revisions.map((revision): RevisionHistoryEntry => {
```
jayenashar: `revision.deleted` marks that the *record* was deleted: `deleteRecord` appends a new head revision with `deleted: true` that copies the parent's AVPs unchanged. In the history that revision therefore diffs to zero changed fields and currently renders as an ordinary edit showing "None", not as a deletion. Carrying `revision.deleted` through into `RevisionHistoryEntry` (a minimal change) would let the History tab label it as a delete event.

library/data-model/tests/engine.form.test.ts
```diff
@@ -1212,4 +1214,36 @@ describe('Form Operations', () => {
       expect(formData.data['Fourth-1'].attachments).toHaveLength(1);
     });
   });
+
+  describe('getHistoryData', () => {
```
jayenashar, line 1224: This only covers a single root revision. The motivating cases for the feature are a child revision diffed against its parent and a multi-parent merge revision (the new seed script builds one). A test asserting per-parent `changedFields` on a merge revision would guard the diff logic the History tab relies on.

api/src/scripts/seedTestDataset.ts
```diff
@@ -428,6 +567,14 @@ const main = async () => {
       console.log(`  ✓ ${spec.email} (${spec.name})`);
     }
 
+    // ── Phase 6: Records with merge history ─────────────────────────────────────
+    console.log('\nPhase 6: Seeding a record with a merge revision...');
+    const mergeRecordId = await seedRecordWithMergeRevision({
```
jayenashar: Phase 6 runs unconditionally, and `pickTwoEditableFields` throws if the Red Notebook has no form with at least two fields. A throw here aborts the whole seed run after earlier phases have already written data, leaving a partially seeded dataset. Consider guarding this phase or making its failure non-fatal.

api/src/scripts/seedTestDataset.ts
jayenashar: `uiSpec.fields[id]` is indexed without a guard, so a view that references a field id missing from `fields` would throw here (and on the `type-returned` check below). Optional chaining (`uiSpec.fields[id]?.[...]`) keeps the seed robust against partial specs. (Also flagged by another reviewer.)

app/src/gui/pages/viewRecord.tsx
```diff
@@ -356,6 +360,100 @@ const ViewTabContent: React.FC<ViewTabContentProps> = ({
...
+    );
+  }
+
+  const revisionIdsRevision = new Map<string, RevisionHistoryEntry>(historyData.map(entry => [
+    entry.revisionId,
+    entry,
+  ]));
+
+  const formatRevisionMetadata = (entry?: RevisionHistoryEntry) => {
+    return entry ? `${entry.createdBy} at ${formatTimestamp(new Date(entry.created).getTime())}` : 'unknown';
```
jayenashar: `new Date(entry.created).getTime()` is redundant: `formatTimestamp` already accepts a string and guards NaN, so `entry.created` can be passed straight through.

app/src/gui/pages/viewRecord.tsx
```diff
@@ -356,6 +360,100 @@ const ViewTabContent: React.FC<ViewTabContentProps> = ({
...
+
+  return (
+    <Stack spacing={4}>
+      <Typography variant="h5">Revision History</Typography>
+      {historyData.sort((a, b) => b.created.localeCompare(a.created)).map(entry => {
+        const parentFields = Object.entries(entry.changedFields);
+        return (
+          <Stack key={entry.revisionId} spacing={2}>
+            <Typography variant="body1" id={entry.revisionId}>
+              Revision created by <span style={{textDecoration: 'underline'}}>{formatRevisionMetadata(entry)}</span>
```
jayenashar: The author and timestamp render as a single underlined span ("Revision created by <createdBy at timestamp>"), which reads a little awkwardly. Consider separating author and timestamp (e.g. emphasise the author, plain timestamp) for a clearer audit line.

library/data-model/src/databaseEngine/types.ts
```diff
@@ -744,6 +744,22 @@ const initialFormData = z.object({
 
 export type InitialFormData = z.infer<typeof initialFormData>;
 
+/**
+ * A single entry in a record's revision history, for displaying an audit
+ * trail of who changed a record and when. Carries the revision identity,
+ * authorship, and the IDs of the fields that changed relative to the
+ * revision's parent (added, value-changed, or removed fields).
+ */
+const revisionHistoryEntry = z.object({
```
jayenashar: `revisionHistoryEntry` is defined but only used to infer the type; `getHistoryData` builds and returns these objects without parsing them through the schema, so the values are never validated at runtime. Parse the output through it (e.g. `return revisionHistoryEntry.array().parse(result)`) so the validation actually runs.

app/src/gui/pages/viewRecord.tsx
jayenashar: opted for slice().sort()

library/data-model/src/databaseEngine/engine.ts
jayenashar: only root's parent would be missing and that's a full-field change

library/data-model/src/databaseEngine/engine.ts
jayenashar: this is local in pouchdb so not a round trip.  if was between api and db, at least it's better than the first version of this PR!

library/data-model/tests/engine.form.test.ts
jayenashar: acceptable

api/src/scripts/seedTestDataset.ts
jayenashar: already accepted

api/src/scripts/seedTestDataset.ts
jayenashar: already accepted

app/src/gui/pages/viewRecord.tsx
jayenashar: formatTimestamp accepts a number string not a ISO date string.

app/src/gui/pages/viewRecord.tsx
jayenashar: acceptable

library/data-model/src/databaseEngine/types.ts
jayenashar: ok learning how the schema is supposed to be used

library/data-model/src/databaseEngine/engine.ts
jayenashar: carried

api/src/scripts/seedTestDataset.ts
```diff
@@ -240,6 +247,138 @@ function clampDescription(description?: string): string | undefined {
...
+      [fieldA]: {data: 'Base value A'},
+      [fieldB]: {data: 'Edited by Bob'},
+    },
+    mode: 'parent',
+    updatedBy: 'seed-bob',
+  });
+
+  // Automerge the two heads into a single revision with both as parents.
+  const fullyMerged = await mergeHeads({projectId, recordId: record._id, dataDb});
+  if (!fullyMerged) {
```
jayenashar: `mergeHeads` returns `true` even when there was nothing to merge: with a single head the `while (working_heads.length > 1)` loop in `merging.ts` is skipped, `fully_merged` stays `undefined`, and the tail coerces it to `true`. So `if (!fullyMerged)` does **not** catch the case where branching failed to produce two heads, and Phase 6 would log "Seeded merge record" without ever creating a multi-parent merge revision (silently defeating the whole point of this seed).

Suggest a post-condition that confirms a real merge happened, e.g. reload the record and assert the single head revision has `parents.length >= 2`, throwing otherwise.

app/src/gui/pages/viewRecord.tsx
```diff
@@ -356,6 +360,100 @@ const ViewTabContent: React.FC<ViewTabContentProps> = ({
...
+        const parentFields = Object.entries(entry.changedFields);
+        return (
+          <Stack key={entry.revisionId} spacing={2}>
+            <Typography variant="body1" id={entry.revisionId}>
+              {entry.deleted ? 'Record deleted by ' : 'Revision created by '}<span style={{textDecoration: 'underline'}}>{formatRevisionMetadata(entry)}</span>
+            </Typography>
+            <Stack sx={{pl: 2}}>
+              {parentFields.map(([parentId, fields]) => 
+                <Typography variant="body1" key={parentId}>
+                  Fields changed{parentFields.length > 1 ? <> compared to <Link href={`#${parentId}`}>{formatRevisionMetadata(revisionIdsRevision.get(parentId))}</Link></> : ''}: {fields.map(
```
jayenashar: The "compared to <parent>" link only renders when `parentFields.length > 1` (merge revisions). For ordinary single-parent edits no parent is linked, and because the list is sorted by `created` (not lineage), in a branched history the entry directly above is not necessarily this revision's parent. That makes a non-merge edit's lineage hard to trace, which is the same audit-trail clarity the merge links are meant to provide.

Consider linking every non-root parent (e.g. gate on `parentId !== 'root'` instead of `parentFields.length > 1`) so each revision points at its actual parent regardless of merge-ness.

api/src/scripts/seedTestDataset.ts
jayenashar: it's fine.  i just wanted the seed so i could develop the merge, and it works for that.

app/src/gui/pages/viewRecord.tsx
jayenashar: opted to link every parent which is not the next chronologically

api/src/scripts/seedTestDataset.ts
```diff
@@ -240,6 +247,138 @@ function clampDescription(description?: string): string | undefined {
...
+ *
+ * Builds a base revision, branches two heads that each edit a different field
+ * (so the changes automerge cleanly), then automerges the heads into one
+ * revision with two parents.
+ *
+ * @param projectId The notebook/project to seed the record into
+ * @param createdBy User recorded as the author of the base revision
+ * @returns The seeded record id
+ */
+async function seedRecordWithMergeRevision({
```
stevecassidy: As mentioned, we currently can't generate merge revisions in practice but this is good test data to have.  This module is just generating sample users/templates/notebooks, there is another module that generates sample data for a notebook (api/src/couchdb/devtools.ts) and this would be better integrated into that code.

stevecassidy: It might be most useful to have this functionality available for testing in the longer term too.  Maybe move it into the devtools module for now and we can make use of it later.

stevecassidy (approved):
Looks good, happy with the component, just a note on the generation of test data but once that is resolved this can be merged.

api/src/scripts/seedTestDataset.ts
jayenashar: i thought we can generate merge revisions, but we have last-wins instead of manual conflict resolution.

jayenashar:
not sure if i need to request re-review or not.  you said "once that is resolved this can be merged"

api/src/couchdb/devtools.ts
```diff
@@ -285,3 +291,135 @@ export const validateProjectDatabase = async (project_id: ProjectID) => {
 
   return {errors: errors};
 };
+
+/**
```
stevecassidy: Good to have this here but not sure how to use it yet.  Maybe add it in to the main random record generation function so that we always generate a merge record when we do that.

PR #2154, opened 2026-06-24
feat: filter forms which are visible in the overview map

PeterBaker0:
Adds a new setting the form settings to show/hide elements from the overview map. 

Later this would be part of a more expansive spatial settings for forms (and fields). Right now you can't control anything, this allows form level visibility, but future work would change iconography, conditional visibility, per-field visibility etc.


stevecassidy (approved):
Looks good, tests ok.

PR #2156, opened 2026-06-25
feat: geopackage export and improved random record generation pipeline

PeterBaker0:
## Geopackage exporting

Our stakeholders have requested native geopackage exports. 

The existing geospatial export pipeline is pretty well setup to handle this. 

## Approach - gdal callout

Using GDAL appears to be the most robust option for this workflow - pumping geoJSON (easy in Node) into Geopackage as a subprocess callout to the ogr2ogr CLI. This is a reliable approach but does add in some weight to the images and some maintenance overhead. 

I propose it's worth it. 

## layer approach

Pumping the default geoJSON -> Geopackage gives a pretty poor structure - the feature merging between all forms results in a very nullish sparse structure with a lot of redundant columns. 

This approach buckets the geojson as its streamed then appends layer by layer grouped by a) form b) geometry type. This gives a much cleaner format for direct ingestion into QGIS. 

**Note**: we have become aware that ArcGIS has limitations with field and column text lengths preventing importing in some cases. QGIS is our test target but future work might be needed to enhance compatibility. 

## Side note

Improves the bulk record generation dev pipeline with 

a) improves spatial field generation to randomly place across Australian bounding box
b) some basic optimisations + parallel generation since embarrassingly parallel 
c) option to not add attachments to speed up this kind of testing
d) improved UI in web


api/src/couchdb/export/gdal.ts
```diff
@@ -0,0 +1,128 @@
+/**
+ * GDAL / ogr2ogr helpers for GeoPackage export.
+ *
+ * FAIMS does not embed a GeoPackage writer; conversion is delegated to the system
+ * `ogr2ogr` binary. Callers produce layered GeoJSON FeatureCollections on disk,
+ * then this module assembles them into a single multi-layer `.gpkg` file.
+ *
+ * GDAL must be on PATH (`gdal-bin` on Debian/Ubuntu, `gdal-tools` on Alpine).
```
stevecassidy: `brew install gdal` on MacOS

api/src/couchdb/export/geospatialExport.ts
```diff
@@ -532,79 +823,189 @@ function createInitialStats(
...
+    streamFinalizeTasks.push(waitForStream(kmlStream));
+  }
+
+  if (gpkgCollector && geopackagePath && tempDir) {
+    try {
+      const layers = await gpkgCollector.finalize();
+      result.geopackage!.featureCount = gpkgCollector.featureCount;
+
+      if (gpkgCollector.featureCount > 0) {
+        await convertLayeredGeoJsonToGeoPackage({
```
stevecassidy, line 965: If we don't have gdal installed this will throw an error but it won't make it back to the user or into logs.

api/src/couchdb/devtools.ts
```diff
@@ -138,14 +309,19 @@ const generateValue = (field: any) => {
     return [options[randomInt(options.length)]];
   }
 
+  if (field['component-name'] === 'MapFormField') {
+    return generateMapFormFieldValue(field);
```
stevecassidy, line 326: Note you should use this to generate Location values below as well (for TakePoint fields)

stevecassidy (approved):
Just a few comments but overall seems to work and is neatly done (apart from the external binary dependency which seems unavoidable). Can't tell if the resulting file is in a useful shape but have confirmed it is readable by the viewer.

PR #2159, opened 2026-06-26
fix: restoring map layout after MUI v9

PeterBaker0:
The map layout got off-centred, excessive padding, and cutoff the attribution on smaller phones. Just doing some css to try and get it looking the same as before. Seems good now.


stevecassidy (approved):
Looks much neater. Thanks.

PR #2161, opened 2026-06-29
feat: adds search to designer and improved field selection

PeterBaker0:
## Summary

Adds fuzzy search across the notebook designer and replaces plain field dropdowns with a reusable, scoped field picker. Large notebooks are easier to navigate, and field references in conditions and computed expressions are easier to find.

### Global design search
- New `DesignerGlobalSearch` bar in the design panel header searches **forms**, **sections**, and **fields** across the notebook.
- Results are ranked with weighted fuzzy matching (`fuzzysort`) over labels, IDs, and helper text, with match highlighting in the dropdown.
- Selecting a result navigates to the correct form tab and section (`?section=` / `?field=` URL params), scrolls the target into view, and expands the field accordion when applicable.
- Press **`/`** to focus search when not typing in an editable control - this is becoming industry standard for shortcut to search over complex interfaces

### Reusable field search (`FieldSearchAutocomplete`)
- Shared autocomplete component for picking fields with fuzzy search, configurable **scope** (`all`, `viewset`, or condition **context**), and optional **filters** (exclude IDs, component types, custom predicates).
- Replaces the plain MUI `Autocomplete` in **condition editors** — same scoping rules as before (same-form fields only; excludes self or current section fields).
- Replaces the chip list in **computed field editor** with a searchable picker limited to numeric fields; shows an info alert when none are available.

### Navigation & deep linking
- New navigation layer (`scrollToForm` / `scrollToSection` / `scrollToField`) with `data-designer-form`, `data-designer-section`, and `data-designer-field` anchors.
- `DesignPanel` keeps the active tab in sync when navigation happens outside the tab bar.
- `FormEditor` resets to the first section on form change but still honours `?section=` from search/deep links.
- Hidden fields scroll into view before autofocus when a new field is added.

### Refactors & tests
- Extracts `resolveAddedFieldKey` so field-list UI and `fieldAdded` reducer share the same key-generation logic (including HRID handling for templated string fields).
- Adds unit tests for design search, field search, field ID resolution, and the field reducer.
- Adds `fuzzysort` dependency.

---

## Test plan

- [ ] Open a notebook with multiple forms/sections/fields; use global search to jump to a form, section, and field — verify correct tab, section step, scroll position, and field accordion expansion.
- [ ] Press `/` outside inputs to focus search; confirm it does not trigger while typing in text fields.
- [ ] Edit a **field condition** — verify the field picker only shows same-form fields, excludes the current field, and fuzzy search works on labels/IDs.
- [ ] Edit a **section condition** — verify section fields are excluded from the picker.
- [ ] Edit a **computed field** — verify only numeric fields appear in the insert picker; confirm info message when none exist; insert a field reference into the expression.
- [ ] Add a new field (including templated string / HRID) — verify it expands and focuses correctly.


web/src/designer/domain/notebook/ids.ts
```diff
@@ -62,3 +62,42 @@ export const buildUniqueFieldName = (
...
+  fieldType: string,
+  viewId: string,
+  sectionFieldNames: string[],
+  existingFieldNames: string[]
+): string => {
+  let fieldLabel = slugify(fieldName);
+
+  if (fieldType === 'TemplatedStringField') {
+    // First HRID in a section uses the canonical `hrid{viewId}` id.
+    const hasHRID = sectionFieldNames.some(
```
stevecassidy: Do we rely on this at all now? I thought we'd gone beyond this.

web/src/designer/components/field-selector/FieldSearchAutocomplete.tsx
```diff
@@ -0,0 +1,220 @@
...
+/** Renders fuzzysort highlight markup; falls back to plain text when no match. */
+const renderHighlighted = (
+  match: Fuzzysort.Result | null | undefined,
+  fallback: string
+) => {
+  if (!match) return fallback;
+  return (
+    <span
+      dangerouslySetInnerHTML={{
+        __html: match.highlight('<strong>', '</strong>'),
```
stevecassidy: If the match occurs in help text then this shows the help text and not the field name.

stevecassidy (approved):
Just one issue with the search result.

web/src/designer/components/field-selector/FieldSearchAutocomplete.tsx
PeterBaker0: Fixed :)

web/src/designer/domain/notebook/ids.ts
PeterBaker0: Removed the additive hrid prefix logic

PR #2162, opened 2026-06-29
fix: attachments stuck on saving... and navigation no longer locked

PeterBaker0:
## Summary

Fixes attachment field save/display bugs and navigation race windows affecting TakePhoto, FileUploader, and AudioRecorder. Maintains contract that **fields own `data`** while **`handleAddAttachment` only manages PouchDB storage and `attachments` metadata**.

## Problem

After the Node v24 update (though uncertain if related or coincedental), seeing:

- **"Saving…" stuck indefinitely** (header, nav buttons, or photo thumbnail overlay)
- **Photos disappearing** from the gallery after capture, despite successful PouchDB writes
- **Brief window where section navigation was allowed** immediately after closing the Capacitor camera or file picker, before the save lock engaged

Debug tracing confirmed attachment storage (`putAttachment`) and `performSave` were completing successfully — the failures were in **UI state management and field value merging**, not PouchDB.

## Root cause

Three separate issues:

### 1. Stale read-modify-write on field values (photos vanishing)

Attachment fields store two related pieces of state on each field:

- `attachments` — metadata used by `useAttachments` to load/display files
- `data` — field-specific value (for these fields, a `string[]` of attachment IDs)

The flow was:

1. `handleAddAttachment` → `form.setFieldValue({ …, attachments })`
2. Field → `setFieldData([…, newId])` immediately after `await addAttachment()`

TakePhoto **did** wait for `addAttachment`. The bug was that `setFieldData` spread **`field.state.value` from the last render**, which didn't yet include the new `attachments`. That second write **overwrote** the first, dropping `attachments` from field state. The file remained in PouchDB but the UI lost the reference and the gallery went empty.

### 2. Non-reactive save indicators (header / nav "Saving…")

The header used `pendingValuesRef.current` directly in JSX, and nav status text read `hasPendingChanges()` inside a `useMemo` keyed only on nav-local state. Refs don't trigger re-renders, so the UI could remain on "Saving…" (or fail to update) after save completed.

### 3. Navigation lock applied too late

`setAttachmentSaving(true)` was called **after** camera/file-picker returned and sometimes **after** blob conversion, EXIF tagging, etc. During that gap, section Next/Back was enabled. AudioRecorder had the inverse bug — it released the lock **before** `addAttachment` on stop.

### Field value merging (`Field.tsx`, `EditableFormManager.tsx`)

- **`setFieldData` / `setFieldAttachment` / `setFieldAnnotation`** now use TanStack Form **functional updaters** (`field.handleChange(prev => …)`) so merges always apply to the latest pending value.
- **`handleAddAttachment`** writes **`attachments` only** via `form.setFieldValue(fieldId, prev => …)`.
- **Fields** restore responsibility for writing into `data` after `await addAttachment()`, using functional `setFieldData(prev => …)` updaters.

### Save status UI (`EditableFormManager.tsx`, `useNavigationLogic.tsx`)

- Added `hasPendingSave` React state to drive the header save indicator.
- Pass `hasPendingSave` + `isFormSaving` into navigation logic instead of reading refs inside a stale memo.

### TakePhoto pending overlay (`TakePhoto/index.tsx`)

- "Saving…" overlay only shown while `attachmentId === null` (during PouchDB write).
- Pending preview kept until `useAttachments` confirms load (avoids eternal overlay without removing the photo prematurely).

### Navigation lock timing (all attachment fields)

| Field | Change |
|-------|--------|
| **TakePhoto** | Lock set **before** `Camera.getPhoto` (after permissions) |
| **FileUploader** | Lock on `onFileDialogOpen`; drag-and-drop path sets lock at drop |
| **AudioRecorder** | Lock held from record start through `addAttachment` on stop; released in `finally` |

### Debug tracing (`logging.ts`, gated by `DEBUG_APP`)

- Added opt-in `[attachment-save-trace]` logging across the save pipeline, toggled via `setAttachmentSaveTraceEnabled(DEBUG_APP)` at app startup.

## Test plan

- [ ] Take photo on native (Capacitor): photo stays visible, header shows Saved, cannot tap Next during camera/processing/save
- [ ] Upload file via picker and drag-and-drop: files appear in list, navigation blocked during picker/save
- [ ] Record and stop audio: attachment saved, navigation blocked through save
- [ ] Confirm no regression on form debounced save ("Saving…" clears ~1s after last edit)
- [ ] With `VITE_DEBUG_APP=false`, no attachment trace noise in console
- [ ] With `VITE_DEBUG_APP=true`, traces show full save → load → performSave pipeline

### DIFF
--- a/app/src/logging.tsx
+++ b/app/src/logging.tsx
@@ -20,6 +20,7 @@
 
 import Bugsnag from '@bugsnag/js';
 import BugsnagPluginReact from '@bugsnag/plugin-react';
+import {setAttachmentSaveTraceEnabled} from '@faims3/data-model';
 import {FormLogger, LoggingService} from '@faims3/forms';
 import DashboardIcon from '@mui/icons-material/Dashboard';
 import {Button, Grid, Typography} from '@mui/material';
@@ -262,9 +263,11 @@ const createBugsnagFormLogger = (
 const registerFormsLogger = (): void => {
   const formLogger = createBugsnagFormLogger(Bugsnag, BUGSNAG_KEY);
   LoggingService.register(formLogger);
+  setAttachmentSaveTraceEnabled(DEBUG_APP);
   console.debug('[FormLogger] Registered with LoggingService', {
     bugsnagEnabled: !!BUGSNAG_KEY,
     debugEnabled: DEBUG_APP,
+    attachmentSaveTraceEnabled: DEBUG_APP,
   });
 };
 
--- a/library/data-model/src/databaseEngine/engine.ts
+++ b/library/data-model/src/databaseEngine/engine.ts
@@ -1,3 +1,4 @@
+import {attachmentSaveTrace} from '../logging';
 import {isEqualFAIMS} from '../datamodel';
 import {DatabaseInterface} from '../types';
 import {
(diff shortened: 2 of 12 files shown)


stevecassidy (approved):
Thanks for the catch and the fix.

PR #2163, opened 2026-06-30
docs: add Computed Value field user documentation

Caedous:
## Description
Adds user documentation for the Computed Value field (merged in #2141),
following the existing field-doc format.

## Proposed Changes
- New user doc at `docs/user/field-types/number-fields/computed-value-field.md`,
  modelled on the Templated String field doc.
- Registered in the number-fields `index.md` (toctree + field list).
- Two screenshots: adding the field, and the configured editor.

## Notes
Covers what the field does, adding and configuring it, the brace-wrapped
expression syntax (e.g. `{Width} * {Height}`), the insert-field chips, and the
blank-until-complete behaviour.

## Checklist
- [x] Relevant documentation is updated.


stevecassidy (approved):
Thanks, looks good.

PR #2165, opened 2026-07-01
feat: add admin user impersonation in API, web, and app

PeterBaker0:
## Summary

Adds admin user impersonation so operations admins can sign in as another user to troubleshoot or verify permissions.

- New `IMPERSONATE_USER` action and `POST /api/users/:id/impersonate` endpoint with audit claim (`impersonatingUserId`) and short-lived refresh tokens
- **Control Centre:** impersonate from the admin users table; session swap with banner and “Return to your account”
- **Field app:** impersonate via user picker dialog; same session swap + banner flow
  -  Note: this implementation won't scale that well to hundreds of users due to the one big request + client side filtering - known limitation with user DB implementation at the moment
- Safety rails: cannot impersonate yourself, disabled accounts, or top level system admins
- Tests for data-model permissions, API endpoint, and app auth slice

## Test plan

- [ ] Sign in as an operations admin in Control Centre → Users → impersonate a normal user → confirm UI loads as that user → return to admin account
- [ ] Sign in as admin in the field app → Impersonate a user → confirm banner and behaviour → return to admin account
- [ ] Verify impersonation is blocked for self, disabled users, and `GENERAL_ADMIN` accounts


stevecassidy (changes requested):
There is an issue surfaced here that we've had for a while - changing users in the app leaves the previous user's list of notebooks on display.   If admin has no notebooks then switching to userA and refreshing the app gives them the list of userA's notebooks.  Then switch back I still see them, switch to userB and I still see them until I refresh when I see userB's notebooks.  

It works ok in other parts of the app - if I'm in a record then I'm sent back to the notebook list where I see the notebook but can't navigate to a record.  

Probably just need to refresh notebooks after user auth?

stevecassidy (commented):
I still see persistent projects listed when switching users - you are pulling the new directory but that is not updating the active projects that are displayed.  In fact it is still trying to probe those projects and getting 401 responses.  The Refresh button doesn't remove the old active notebooks now either (I think it did before).

stevecassidy (approved):
All good now, thanks!

PR #2166, opened 2026-07-02
feat: styling improvements

PeterBaker0:
This returns the white background in the bssTheme, adds subtle offwhite backdrop the stepper, reduces excessive padding and stack spacing, removes unwanted stacked margins from the multi select, and other misc tidy ups. 

Before:

image

After:
image


ranisa-gupta16 (approved):
thanks for the fix, looks fine to me, 
The changes are as described and shows clean white background and padding fixed.

PR #2168, opened 2026-07-02
fix: notebook crashes if loading route when de-activated, or after removal

PeterBaker0:
## Summary

Prevents crashes and infinite spinners when a notebook disappears from the local store (remote archival/deletion) or is opened before activation.

- **Remote removal:** Directory sync probes missing notebooks (`GET /api/notebooks/:id`) instead of pulling archived listings; removes archived notebooks immediately, debounces transient 404s, and navigates off stale routes via `handleRemoteProjectRemoval` + `useProjectRouteGuard`. 
  - The change to not pull archived notebooks is mainly for performance - there is no reason to repeatedly fetch the potentially many archived notebooks again and again - only do targeted checks for specific notebooks
- **Route resilience:** Notebook/record pages show `NotebookUnavailable` or redirect home instead of throwing; `editRecord` / `viewRecord` fix Rules-of-Hooks violations and gate DB access with `tryLocalGetDataDb`.
- **Unactivated notebooks:** Deep links to an unactivated notebook redirect to `/` rather than crashing in `useRecordList` / `useRecordAudit`.
- **Query safety:** `useRecordList` and `useRecordAudit` skip queries when no local Pouch DB exists; shared `queryClient` + `appRouter` support cleanup/navigation outside React.

## Test plan

- [ ] Archive or delete a notebook while it is open → redirect to home, no crash, no repeated alerts
- [ ] Server has archived surveys the user never activated → refresh does not alert or remove anything spurious
- [ ] Deep-link to an unactivated notebook route → redirect home, no `localGetDataDb` error
- [ ] Open a record on a notebook that is then removed → clean redirect, no ErrorBoundary
- [ ] Normal activated notebook flow (list, view, edit records) unchanged


stevecassidy (approved):
Looks good, thanks.

PR #2169, opened 2026-07-02
build: pre-commit hooks with Husky

PeterBaker0:
Uses husky to auto format staged files pre-commit. Run pnpm prepare (or pnpm i) to setup.


stevecassidy (approved):
Looks like it will be useful

PR #2170, opened 2026-07-02
feat: removing redundant status display and improving team element of template form + permissions fix

PeterBaker0:
**Templates list**
- Replace Status column with Last updated (archived templates are hidden anyway)

**Create Template dialog**
- Tighter layout; team ownership shown in a compact highlighted callout
- Fix create button for users with global `CREATE_TEMPLATE` (no team required)
- If team is mandatory and there’s only one option → callout; if global create is allowed → optional team dropdown with clear

**Create project from template**
- Same team callout/dropdown rules as template creation
- Default team to the template’s owner when the user can create there
- Refresh auth token after create so the new project shows on the template’s Projects tab


stevecassidy (approved):
Nice cleanup, thanks.

PR #2171, opened 2026-07-02
chore: map control refactor

PeterBaker0:
Greatly simplifies the layout, styling and complexity of the map controls. Previous approach of using ol controls just had so many css, flex, integration challenges. This is much simpler. Uses react overlays and hooks in via pre-defined controls/events. Uses theming for color control.


stevecassidy (approved):
Much neater, thanks!

PR #2173, opened 2026-07-06
feat: select recommended offline map region for survey with app prompts to download during activation

PeterBaker0:
## Summary

- Adds a per-notebook **recommended offline map region** (EPSG:4326 polygon) stored on the project, exposed via `PUT /api/notebooks/:id/offlineMapRegion` and a new **Offline map** tab in the web project settings.
- On the mobile app, prompts users to download the region after activation (or when the plan region changes), with progress, cancellation, size estimates, and a notebook settings panel to manage download status.
- has an update batch which manages cases where multiple updates occur from the server e.g. 2 new regions to download, one region changed
- cleans up old no longer used regions due to server updates
- permissions are enforced in the web. users without edit permission can see the region but not change it (matching API spec)

## Test plan

- [ ] Web: configure an offline map region on a project (save and clear)
- [ ] App: activate a notebook with a configured region — confirm post-activation prompt, download progress, and successful completion.
- [ ] App: change the region on the server and sync — confirm stale download is removed and re-download prompt appears.
- [ ] App: cancel an in-progress download from the prompt and from notebook settings; confirm partial tiles are cleaned up.
- [ ] App: re-activate a notebook with an already-downloaded matching region — confirm prompt is skipped.


api/src/couchdb/notebooks.ts
```diff
@@ -564,6 +565,29 @@ export const updateProjectUiSpecification = async (
...
+  const updated: ProjectDocument = {
+    ...project,
+    ...(offlineMapRegion == null
+      ? {offlineMapRegion: undefined}
+      : {offlineMapRegion}),
+    updatedAt: nowIso(),
+  };
+  if (offlineMapRegion == null) {
+    delete updated.offlineMapRegion;
+  }
```
PeterBaker0, line 588: this is a bit redundant - if null it sets to undefined then deletes, just unroll it then delete

app/src/context/slices/projectSlice.ts
```diff
@@ -780,24 +810,23 @@ const projectsSlice = createSlice({
...
+    setPendingOfflineMapDownloadPrompt: (
+      state,
+      action: PayloadAction<PendingOfflineMapDownloadPrompt>
+    ) => {
+      state.pendingOfflineMapDownloadPrompt = action.payload;
+    },
+
+    clearPendingOfflineMapDownloadPrompt: state => {
+      state.pendingOfflineMapDownloadPrompt = undefined;
+    },
```
PeterBaker0, line 867: Comments for these

PeterBaker0 (commented):
Self-review

web/src/components/tabs/project/offline-map.tsx
```diff
@@ -0,0 +1,539 @@
...
+        return {
+          variant: 'info' as const,
+          message: canSetOfflineMapRegion
+            ? 'Click Draw Area to select a map area for offline download.'
+            : `No offline map area has been selected for this ${NOTEBOOK_NAME}.`,
+          actions: (
+            <RestrictedOfflineMapButton allowed={canSetOfflineMapRegion}>
+              <Button
+                variant="default"
+                onClick={handleDrawArea}
```
stevecassidy, line 397: If this is disabled I don't get a useful message - it just doesn't do anything.  Maybe don't show this if we can't set offline map regions.

app/src/gui/components/maps/projectOfflineMap.ts
```diff
@@ -0,0 +1,277 @@
...
+}): Promise<void> {
+  const tileStore = getSharedTileStore(config);
+  await tileStore.tileStore.initDB();
+
+  const extent3857 = offlineMapRegionToExtent3857(region);
+  const setName = projectOfflineMapSetName(projectId);
+
+  await tileStore.createTileSet(extent3857, setName, undefined, undefined, {
+    projectId,
+    label: projectName,
```
stevecassidy: Maybe call it `{projectName} default offline map`?

stevecassidy (approved):
Nice feature, thanks.

PR #2174, opened 2026-07-07
feat: scroll new field name input into view on creation

### LINKED ISSUE #2148: Designer: scroll to view new field after adding it
When I add a new field in designer on my laptop screen, in some cases (number field) the bottom of the field accordion is visible but not the top where the field name is.  

On field creation, scroll to view the field name entry box which has focus so I can see what I'm typing without scrolling.

Caedous:
# feat: scroll new field name input into view on creation (#2148)

## JIRA Ticket

N/A (GitHub issue #2148)

## Description

When adding a field in the designer on a laptop-height screen, the new field's accordion expanded and its Label input received focus, but the input was often left off-screen — for taller fields (e.g. Number) only the bottom of the accordion was visible while the field name box sat below the fold, so the user was typing into a box they could not see.

While fixing this, a pre-existing rename loop in `BaseFieldEditor` was found and fixed as well (see below). It is a separate defect but was surfaced and made visible by the scroll work, so both are included here. Happy to split it into its own issue/PR if preferred.

## Proposed Changes

- `web/src/designer/components/field-editor.tsx`
  - After a new field is added, focus its Label input and scroll it to the centre of the viewport. The scroll waits for the accordion's expand animation to finish (watching the scroll container's height) so the input can reach the centre rather than being pinned to the bottom edge, and reads the focus callback via a ref and runs once per activation so the effect cannot re-fire in a loop.

- `web/src/designer/components/Fields/BaseFieldEditor.tsx` (pre-existing defect)
  - The Label to Field ID auto-sync could rename fields in a loop when several default "New Field" fields were open at once: blurring an untouched field renamed it, and the `[fieldName]` effect re-armed auto-sync after that rename because the deduped `New-Field-N` id still matched the "fresh field" pattern. Guarded so a field only syncs when the user actually edited its label, and auto-sync is not re-armed after our own rename.

## How to Test

1. Open the designer on a laptop-height window and open a form section.
2. Add a Number field low in the section. Its Label box should scroll to the centre of the viewport, focused and selected.
3. Add several fields in a row without typing labels, leaving them open. The Field IDs should stay `New-Field`, `New-Field-1`, `New-Field-2`, ... with no cycling or back-and-forth scrolling.
4. Type a label on one field, pause, and confirm its Field ID auto-syncs once to the slug of what you typed.

## Additional Information

Closes #2148

## Checklist

### DIFF
--- a/api/src/api/notebooks.ts
+++ b/api/src/api/notebooks.ts
@@ -47,6 +47,7 @@ import {
   PutChangeNotebookStatusInputSchema,
   PutChangeNotebookTeamInputSchema,
   PutUpdateNotebookMetadataInputSchema,
+  PutUpdateNotebookOfflineMapRegionInputSchema,
   PutUpdateNotebookResponse,
   PutUpdateNotebookUiSpecificationInputSchema,
   removeProjectRole,
@@ -98,6 +99,7 @@ import {
   getUiSpecModel,
   getUserProjectsDetailed,
   updateProjectMetadata,
+  updateProjectOfflineMapRegion,
   updateProjectUiSpecification,
 } from '../couchdb/notebooks';
 import {getTemplate} from '../couchdb/templates';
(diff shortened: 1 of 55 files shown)


Caedous:
Just looking into why the Build & Lint failed, it appears to be two files unrelated to this PR. Not sure what you need me to do or if I need to fix these up myself? I didn't interact with either of these.

stevecassidy (approved):
Nice fix, thanks!

PR #2175, opened 2026-07-07
chore: build config refactor to typed config singleton with shared Zod helpers

PeterBaker0:
## Description

Refactors runtime configuration across `api`, `app`, and `web` to a typed `config` singleton instead of scattered env-derived exports. Parsing is a **one-pass** Zod pipeline: each env key is declared once with coercion/defaults, unknowns are stripped, and a final `.transform()` renames ENV_KEYS into the camelCase `config` shape (plus a few cross-field values).

Shared Zod schema factories live in `@faims3/data-model` as `configHelpers`. The monorepo is standardised on Zod v4.

## Proposed Changes

- **Shared helpers**: `library/data-model/src/config/` (`configHelpers`) used by api/app/web/designer schemas
- **api / app `buildconfig.ts`**, **web `constants.ts`**, **designer `buildconfig.ts`**: one-pass Zod schema → `config` singleton (`Prefer config.<field>`)
- **Call sites**: migrate imports from individual constants (e.g. `CONDUCTOR_PUBLIC_URL`, `EMAIL_SERVICE`) to `config.<field>` / `emailService` / `keyService` across api, app, and web
- **web fetches**: replace remaining `import.meta.env.VITE_API_URL` usages with `config.apiUrl`
- **`.env.dist` sync**: fix `VITE_POUCHDB_DEBUG` → `VITE_DEBUG_POUCHDB`; document missing schema vars (API URLs, `KEY_SOURCE`, web theme / token hints / template protections)
- **Docs**: `api/README.md` (`emailService`), Passport auth provider setup, test-dataset seeding CouchDB vars, `AGENTS.md` runtime-config section, `CHANGELOG.md`
- **Zod v4**: bump consumers (including AWS CDK `z.record` / `error.issues`)

No intended runtime behaviour change — config values and defaults are preserved. API `config` is intentionally **not** frozen so tests can override individual fields.

## How to Test

- [ ] `pnpm run lint` and package tests pass in `api`, `app`, and `web`
- [ ] Start stack with existing `.env` files; confirm api boots and serves on expected port
- [ ] App (localhost:3000): login, notebook list, about-build page shows correct server/version info
- [ ] Web (localhost:3001): login, project/template pages, designer loads; API calls still hit Conductor
- [ ] Verify a few env overrides still apply (e.g. custom `VITE_*` / `CONDUCTOR_*` values)
- [ ] Confirm `VITE_DEBUG_POUCHDB` (not the old `VITE_POUCHDB_DEBUG` name) is honoured when set


app/src/buildconfig.ts
```diff
@@ -40,180 +53,375 @@ export const DEFAULT_CONDUCTOR_URL = 'http://localhost:8154';
...
-  return bugsnag_key;
-}
+// ---------------------------------------------------------------------------
+// Pass two: build the typed configuration values.
+// ---------------------------------------------------------------------------
+
+const ConfigSchema = z.object({
+  debugPouchdb: boolFromEnv(false, {label: 'VITE_DEBUG_POUCHDB'}),
+  // Note: badly-defined VITE_DEBUG_APP historically defaults to true.
+  debugApp: boolFromEnv(false, {label: 'VITE_DEBUG_APP', badFallback: true}),
```
stevecassidy: Do we really need badFallback here - it's on us if we get this wrong and we can work it out, this is the only place you use this option

app/src/buildconfig.ts
```diff
@@ -40,180 +53,375 @@ export const DEFAULT_CONDUCTOR_URL = 'http://localhost:8154';
...
+  VITE_AUTOSUGGEST_SOURCE: import.meta.env.VITE_AUTOSUGGEST_SOURCE,
+  VITE_AUTOSUGGEST_MAPBOX_KEY: import.meta.env.VITE_AUTOSUGGEST_MAPBOX_KEY,
+  VITE_MAPBOX_ADDRESS_COUNTRY: import.meta.env.VITE_MAPBOX_ADDRESS_COUNTRY,
+  VITE_AUTOSUGGEST_MAPTILER_KEY: import.meta.env.VITE_AUTOSUGGEST_MAPTILER_KEY,
+  VITE_MAPTILER_ADDRESS_COUNTRY: import.meta.env.VITE_MAPTILER_ADDRESS_COUNTRY,
+  NODE_ENV: import.meta.env.NODE_ENV,
+});
+
+// ---------------------------------------------------------------------------
+// Reusable pass-two field builders.
```
stevecassidy: Please document more or rename these - I found them hard to read but the trick is that they are returning zod schema rather than doing any validation themselves.  So intFromEnv doesn't actually look at the ENV at all, it's just a shortcut for an int schema.  Maybe it's intWithDefault(12), boolWithFallback({default: false, label: "foo", badFallback: true})

app/src/buildconfig.ts
```diff
@@ -40,180 +53,375 @@ export const DEFAULT_CONDUCTOR_URL = 'http://localhost:8154';
...
-  // Used in the server, as opposed to COUCHDB_USER and PASSWORD for testing.
-  const username = import.meta.env.VITE_DIRECTORY_USERNAME;
-  const password = import.meta.env.VITE_DIRECTORY_PASSWORD;
+// ---------------------------------------------------------------------------
+// Pass one: read and validate the raw environment into (optional) strings.
+//
+// The raw object uses explicit static `import.meta.env.VITE_*` accesses so that
+// Vite performs its build-time string replacement for each variable.
+// ---------------------------------------------------------------------------
+
```
stevecassidy, line 126: I don't like how much repetition is needed here - I need to change in three (or more) places if I add a new variable. You could probably derive both EnvSchema and RawEnv from a single array of env variables.

app/src/buildconfig.ts
```diff
@@ -40,180 +53,375 @@ export const DEFAULT_CONDUCTOR_URL = 'http://localhost:8154';
...
-    bugsnag_key === '' ||
-    bugsnag_key === undefined ||
-    bugsnag_key === 'false'
-  ) {
-    return undefined;
-  }
-  return bugsnag_key;
-}
+// ---------------------------------------------------------------------------
+// Pass two: build the typed configuration values.
```
stevecassidy: Again here there is repetition, define the very specific schema and then instantiate it. Can we do this in one step? A function that returns both the ConfigSchema and parsedConfig given suitable input?

stevecassidy (commented):
Overall a great move but wondering if we can do better with buildconfig?

library/forms/lib/formModule/formManagers/EditableFormManager.tsx
PeterBaker0, line 1: Concerned about these new type coersions

AGENTS.md
PeterBaker0, line 1: @stevecassidy  this is an artifact of cursor's agentic cloud - happy to remove if its borthersome.

PeterBaker0 (commented):
I'm happy with this now

api/src/buildconfig.ts
```diff
@@ -32,565 +43,456 @@ import {getKeyService, IKeyService, KeySource} from './services/keyService';
...
+// ---------------------------------------------------------------------------
+
+const EnvSchema = z
+  .object({
+    /** Public base URL of this Conductor server. */
+    CONDUCTOR_PUBLIC_URL: configHelpers.stringDefault(
+      DEFAULT_CONDUCTOR_URL,
+      'CONDUCTOR_PUBLIC_URL'
+    ),
+    /** Public URL of the Field Mark web/data-collection app. */
```
stevecassidy: not 'Field Mark'

stevecassidy (approved):
Looks good, thanks.

PR #2176, opened 2026-07-08
chore: query invalidations in web

PeterBaker0:
Adds some missing /web query invalidations.


ranisa-gupta16:
there are some formatting issues to be fixed, please run pnpm run format to resolve.

ranisa-gupta16 (approved):
Thanks for fixing these invalidations
Have done some tests : like 'projects', undefined now refreshes the projects list it was stale before, 
'templates', undefined also fixed stale templates list,
'teams', undefined fixes stale temas list.

'project-users' - hyphen typo is fixed too which was never matching before.

Small but v. relevant fixes.

PR #2177, opened 2026-07-08
feat: local android apk build helper script

PeterBaker0:
A script which helps build apks - can be run from the home path. Setup the .env file and then run it with `./app/scripts/build-android-apk.sh` you can then install it with `adb install -r app/android/app/build/outputs/apk/debug/app-debug.apk`. Tested and working for our dev build.


PeterBaker0:
Closing in favour of usage of normal cap build unless we see reason this script is helpful.

PR #2178, opened 2026-07-08
Feat: createRecord allows for initial data

stevecassidy:
# Feat: createRecord allows for initial data

## Ticket

[Groundsite #54](https://github.com/FAIMS/GroundSight-FAIMS/issues/54)

## Description

We need to be able to create records with initial data.

## Proposed Changes

Extends `NewFormRecord` to include an optional `initial` property containing `FormUpdateData`.  If present, this is used to update the initial revision immediately after creation.  The new updated revision is returned in this case. 

This allows for adding initial values, annotations and attachments on records at creation time.

Adds some tests for this and rewrites existing tests that were adding initial data to records via separate calls.

## How to Test

Not used yet in this codebase but tests in data-model should still pass including new tests of this functionality.

## Checklist

### DIFF
--- a/library/data-model/src/databaseEngine/engine.ts
+++ b/library/data-model/src/databaseEngine/engine.ts
@@ -1137,7 +1137,18 @@ class FormOperations {
     };
 
     // Create the new revision
-    const rev = await this.core.createRevision(revisionDoc);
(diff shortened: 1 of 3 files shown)


PeterBaker0 (approved):
Lovely - simple change :)

Issue #2186, opened 2026-07-13
Enable react plugin for oxlint and consider settings


stevecassidy:
We don't currently enable the oxlint react plugin and it might be useful.  Consider enabling https://oxc.rs/docs/guide/usage/linter/rules/react/rules-of-hooks for example.

PR #2187, opened 2026-07-13
feat: rule of hooks

PeterBaker0:
AI refactor to add rule of hooks oxlint enforcement #2186 and chase through issues. Also addresses other easy lint fixes. Circular dependency in data model due to form field stuff remains as it's tricky!

### DIFF
--- a/app/src/utils/customHooks.tsx
+++ b/app/src/utils/customHooks.tsx
@@ -430,7 +430,8 @@ export const useRecordList = ({
   projectId: string;
   filterDeleted: boolean;
   metadataRefreshIntervalMs?: number | undefined | false;
-  uiSpecification: CompiledNotebookUiSpec;
+  /** Required when `enabled` is true; may be undefined while the compiled spec loads. */
+  uiSpecification: CompiledNotebookUiSpec | undefined;
   enableProfiling?: boolean;
   /** When false, or when the project has no local Pouch DB yet, skips the query. */
   enabled?: boolean;
@@ -452,7 +453,7 @@ export const useRecordList = ({
   const activeUser = useAppSelector(selectActiveUser);
   const token = activeUser?.parsedToken;
   const dataDb = tryLocalGetDataDb(projectId);
-  const canQueryRecords = enabled && !!dataDb && !!token;
+  const canQueryRecords = enabled && !!dataDb && !!token && !!uiSpec;
 
   // First - just fetch a list of all unhydrated records
   const unhydratedRecordQuery = useQuery({
@@ -481,10 +482,10 @@ export const useRecordList = ({
       const queryFnStart = performance.now();
       profile('queryFn started');
 
-      if (!token || !dataDb) {
-        // Trying to run without token or before notebook activation.
+      if (!token || !dataDb || !uiSpec) {
+        // Trying to run without token, local DB, or compiled UI spec.
         console.warn(
-          'Trying to fetch record list without user token or local DB.'
+          'Trying to fetch record list without user token, local DB, or UI spec.'
         );
         return [];
       }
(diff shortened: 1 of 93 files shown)


web/src/hooks/auth-hooks.ts
```diff
@@ -69,11 +69,12 @@ export const userCanCreateTemplate = (user: User | null): boolean => {
...
-    return false;
-  }
-
-  return useMemo(() => userCanCreateTemplate(user), [user.token]);
+  return useMemo(() => {
+    if (!user || isExpired()) {
+      return false;
+    }
+    return userCanCreateTemplate(user);
+  }, [user, user?.token]);
```
jayenashar: should `isExpired()` be a dependency here?  i know eslint has a rule to catch dependency issues

app/src/utils/customHooks.tsx
```diff
@@ -769,19 +770,17 @@ export const useIsAuthorisedTo = ({
...
+  return useMemo(() => {
+    if (!activeUser) {
+      return false;
+    }
+    return isAuthorized({
+      decodedToken: activeUser.parsedToken,
+      action,
+      resourceId,
+    });
+  }, [action, resourceId, activeUser, activeUser?.token]);
```
jayenashar: should `activeUser?.token` be `activeUser?.parsedToken`?

library/forms/lib/fieldRegistry/registryApi.ts
```diff
@@ -0,0 +1,127 @@
...
+ * `namespace::name` key is missing from {@link FIELD_REGISTRY}.
+ *
+ * Called by {@link ./registry.ts} after registering canonical specs (with the
+ * text-field spec). Must run before any render path calls {@link getFieldInfo}.
+ *
+ * @param spec - Field info to use as the unknown-type fallback
+ */
+export function setRegistryFallback(spec: FieldInfo<any>): void {
+  fallbackFieldInfo = spec;
+}
```
jayenashar: my AI is saying this isn't needed

https://github.com/FAIMS/FAIMS3/pull/2189/changes/c97d68d1b62fa30b0757c962671cfedc630b6232

web/src/components/tabs/admin/global-invites.tsx
```diff
@@ -28,15 +25,19 @@ const GlobalInvites = () => {
 
   const columns = useGetGlobalInviteColumns({
     deleteInviteHandler: async (inviteId: string) => {
+      if (!user) throw new Error('Not authenticated');
```
jayenashar: can move this to a wrapper including useAuth: https://github.com/FAIMS/FAIMS3/pull/2189/changes/17b24298d2459ea28cd66872db1a12cb03aabe21#diff-faddfb4eb8d3b668e2e780142a26be94a77c8342d8af8419b328a493e77a32e1R23-R31

web/src/hooks/auth-hooks.ts
jayenashar: @PeterBaker0 did you answer this?

app/src/utils/customHooks.tsx
PeterBaker0: As discussed this usage is fine I believe.

jayenashar: peter answered this on slack but i think this section of code is confusing if there's a mismatch between the dependencies and the used variables.  will see if i can suggest a change to make this code less confusing in isolation.

web/src/hooks/auth-hooks.ts
PeterBaker0: The expired function is this

```typescript
  const isExpired = () => {
    return isUserExpired(user);
  };
```
Which is just

```typescript
  // Safe backout if no token present
  if (!user || !user.decodedToken || !user.token) return true;
  // Consider a token expired if it's within 1 minute of expiry
  return user.decodedToken.exp * 1000 < nowMs();
```

Pretty cheap - this re-running on each re-render is likely worthwhile. But we still need some provocation for a render to even occur. useAuth should trigger that, so probably okay. Will include.

PeterBaker0: Adding in

jayenashar (approved):

PR #2189, opened 2026-07-13
refactor: rule-of-hooks follow-ups — static registry fallback, useRequiredUser

jayenashar:
# refactor: rule-of-hooks follow-ups — static registry fallback, useRequiredUser

## Ticket

Follow-up to #2187 (targets its `feat/rule-of-hooks` branch).

## Description

Two refactors addressing review findings on #2187. Both keep the rules-of-hooks enforcement intact while removing the ceremony the mechanical fixes introduced.

## Proposed Changes

**1. `refactor(forms): import registry fallback directly instead of late-set mutable state`**

`registryApi.ts` used a mutable module-level `fallbackFieldInfo`, a `setRegistryFallback()` called from `registry.ts` at load time, and a runtime "fallback not initialised" throw in `getFieldInfo`. Any future entry point that loaded `DataView` without the `registry.ts` side-effect module (a unit test, a storybook story, a tree-shaken bundle) would have thrown at first render.

`registryApi.ts` now imports `textFieldSpec` directly as a constant fallback. This is cycle-safe: the text field's transitive imports (form types, fallback renderer types, MUI wrappers) never reach `DataView`, so the RelatedRecord → DataView → registry cycle the registryApi split breaks stays broken. Verified with `madge -c --extensions ts,tsx lib/` (no circular dependencies).

**2. `refactor(web): assert authenticated user once via useRequiredUser`**

Every component under `_protected` is unreachable without a user: `_protected.tsx` returns `<SessionExpiredOverlay />` *instead of* the `<Outlet />` whenever `isAuthenticated` (`!!user && !isExpired()`) is false, and `user` is initialised synchronously from localStorage. So the ~20 `if (!user) return <fallback/>` guards that #2187 had to reorder hooks around are dead code, as are the `if (!user) throw new Error('Not authenticated')` narrowing closures it added to the delete-invite handlers.

This adds a single `useRequiredUser()` hook that asserts the invariant once, converts those components to it, deletes the dead guards/throws/`return []` branches, and reverts the `useGetUsersForTeam` `User | null` + `enabled: !!user` widening (its only caller now passes a non-null user). Net −88 lines, and new components (e.g. plan views) get a non-null `User` without re-litigating where the guard goes.

## How to Test

- `pnpm turbo run build --filter=@faims3/forms` and `pnpm --filter @faims3/forms test` (51 tests pass)
- `library/forms`: `pnpm run lint` (includes the now-fatal `madge` circular-dep check)
- `web`: `pnpm run typecheck` and `pnpm run lint`
- In the web app: browse teams/invites/tokens pages while logged in (unchanged behaviour); log out or let the session expire — the `_protected` layout shows the session-expired overlay before any converted component renders.

## Additional Information

Reviewed pre-push by a multi-model panel (Claude, Codex, Cursor, Antigravity): no findings across circular-dependency, mount-reachability, rules-of-hooks, behavioural-regression, and memo-deps checks.

## Checklist

### DIFF
--- a/library/forms/lib/fieldRegistry/registry.ts
+++ b/library/forms/lib/fieldRegistry/registry.ts
@@ -26,7 +26,6 @@ import {emailFieldSpec, textFieldSpec} from './fields/TextFields';
 import {
   buildRegistryKey,
   FIELD_REGISTRY,
-  setRegistryFallback,
   splitRegistryKey,
 } from './registryApi';
 import {FieldInfo} from './types';
@@ -119,7 +118,6 @@ for (const alias of LEGACY_FIELD_ALIASES) {
     alias.spec
   );
 }
-setRegistryFallback(textFieldSpec);
 
 /**
  * Validate the registry on load. Legacy alias keys are intentionally allowed
--- a/library/forms/lib/fieldRegistry/registryApi.ts
+++ b/library/forms/lib/fieldRegistry/registryApi.ts
@@ -1,15 +1,18 @@
+import {textFieldSpec} from './fields/TextFields';
 import {FieldInfo} from './types';
 
 /**
  * Lookup surface for the field registry.
  *
- * Kept free of field-spec imports so modules like DataView can resolve
+ * Kept free of the full field-spec list so modules like DataView can resolve
  * renderers without forming a static cycle:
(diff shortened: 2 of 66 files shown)


PeterBaker0 (approved):
Good change - better to centralise that assumption then chase it along everywhere.

PR #2191, opened 2026-07-14
build: improved e2e tests for web, app and API

PeterBaker0:
## Summary
- Expands the WebdriverIO e2e suite across smoke, Control Centre (`web`), Conductor auth, Fieldmark (`app`), and cross-surface journeys, with shared helpers, page objects, and a documented tier inventory (`e2e/SUITE.md`).
- Adds stable `data-testid` hooks in web/app/API auth views and wires a headless Chromium CI workflow (`.github/workflows/e2e.yml`) that seeds the test dataset, runs smoke → web → app, and uploads HTML/JUnit/screenshot artifacts.
- Supports reliable local/CI auth flows via `RATE_LIMITER_ENABLED` (HTTP + CouchDB-backed limits), seed-persona env, and an `e2e-minimal` notebook for record CRUD.

## Test plan
- [ ] `pnpm --filter=@faims3/e2e test:e2e:headless:ci` against a seeded local stack (`RATE_LIMITER_ENABLED=false`)
- [ ] Confirm CI E2E job passes and artifacts (`index.html`, JUnit, failure shots) upload correctly
- [ ] Spot-check headed smoke/web/app if debugging failures locally


api/src/couchdb/emailReset.ts
```diff
@@ -82,6 +82,11 @@ export const checkCanCreateEmailCode = async ({
   reason?: string;
   nextAttemptAllowedAt?: number;
 }> => {
+  // Shared with HTTP limiter via RATE_LIMITER_ENABLED (local/e2e often disables).
+  if (!config.rateLimiterEnabled) {
```
jayenashar: - `RATE_LIMITER_ENABLED` was documented HTTP-only before this PR; this early return (same in `verificationChallenges.ts:114`) now also disables the per-user email-code / verification-challenge attempt limits
- a deployment already running `RATE_LIMITER_ENABLED=false` (per-IP limiting handled upstream) silently loses password-reset + verification-code brute-force protection on upgrade
- keeping these attempt limiters always-on and letting e2e raise windowMs/maxAttempts via config would preserve the flag's HTTP-only contract

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,304 @@
...
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: e2e-artifacts-${{ github.run_id }}
+          path: ${{ runner.temp }}/e2e-upload/
+          if-no-files-found: warn
+          retention-days: 14
+
+      - name: Comment e2e result on PR
+        if: always() && github.event_name == 'pull_request'
```
jayenashar: - fork PRs get a read-only `GITHUB_TOKEN` under `pull_request` (job-level `pull-requests: write` can't elevate it), so `createComment` 403s and this `always()` step fails the job even when the suite passed

```suggestion
        if: always() && github.event_name == 'pull_request' && !github.event.pull_request.head.repo.fork
```

e2e/test/specs/conductor/password-reset.e2e.ts
```diff
@@ -0,0 +1,75 @@
...
+        return success || !url.includes('reset-password');
+      },
+      {timeout: 20000, timeoutMsg: 'Expected password reset to complete'}
+    );
+    await captureStep({
+      surface: 'conductor',
+      label: 'reset-complete',
+    });
+
+    // Restore original seed password for later suites.
```
jayenashar: - restore only runs if everything above succeeds; a failure after the reset submit (e.g. the 20s `waitUntil` timing out) leaves shared persona `user` on the temp password, and `register-invite.e2e.ts` then can't log in until a reseed
- a `try/finally` (or `after` hook) around the reset would restore on failure too

api/src/scripts/seedTestDataset.ts
```diff
@@ -22,8 +22,10 @@
...
  *                        Defaults to "TestPassword123!".
  *   TEST_SEED_NOTEBOOKS  Comma-separated paths to notebook JSON files.
- *                        Defaults to "./notebooks/Field-Sampler.json,
+ *                        Defaults to "./notebooks/e2e-minimal.json,
  *                        ./notebooks/sample_notebook.json"
+ *                        (Red = minimal one-field notebook for app record
+ *                        CRUD e2e; Blue = sample survey).
  *
  * WARNING: This is a destructive operation.  All existing data in the
  * configured CouchDB will be wiped and replaced with seed data.
```
jayenashar: - this wipe claim doesn't hold: `initialiseAndMigrateDBs({force: true})` only force-rewrites design/init docs, never deletes people/team/project documents
- rerunning against an already-seeded couch accretes duplicate Red/Blue teams/templates/notebooks (phases 2-4), then exits 1 in phase 5 (`seed-admin@faims.test` already exists), leaving data that breaks first-row-selecting specs
- destroying/recreating the DBs here, or making the script idempotent, would match the docs

e2e/test/helpers/auth.ts
```diff
@@ -0,0 +1,186 @@
...
+}
+
+/**
+ * Best-effort logout for Control Centre (user menu).
+ * Falls back to clearing cookies/storage + navigating away.
+ */
+export async function logoutWeb(): Promise<void> {
+  const webUrl = getWebUrl();
+  try {
+    const logout = await $('[data-testid="web-nav-logout"]');
```
jayenashar: - `web-nav-logout` is a Radix `DropdownMenuItem`, only mounted while the dropdown is open, so this `isExisting()` check (before the menu is opened) is always false and the UI logout path never runs
- the fallback clears cookies/storage on the web origin only; the conductor session on :8080 survives, so a later navigation can silently re-authenticate (masked today because the sole caller also runs `reloadSession()`)
- opening the user menu first, then waiting for the item, makes the UI path live

web/src/routes/_protected.tsx
```diff
@@ -203,7 +203,10 @@ function RouteComponent() {
             </div>
           </header>
 
-          <div className="flex flex-1 flex-col gap-4 p-4 pt-0">
+          <main
```
jayenashar: - `SidebarInset` already renders `<main>` (`web/src/components/ui/sidebar.tsx:318`), so this nests `<main>` inside `<main>`: non-conforming HTML and duplicate main landmarks (axe `landmark-main-is-top-level`)
- keeping this element a `div` with the `data-testid`, or moving the testid onto the SidebarInset main, avoids it

e2e/wdio.web.conf.ts
```diff
@@ -1,18 +1,39 @@
...
+ * Suite label is `web` (includes Control Centre, journeys, and Conductor auth).
  *
  * Set WEB_URL env var to override the dashboard URL.
  * Set VIEWPORT=desktop|wide (default: desktop) to control the screenshot size.
  */
 export const config = {
   ...baseConfig,
-  specs: ['./test/specs/web/**/*.ts'],
-  baseUrl: process.env.WEB_URL || 'http://localhost:3001',
+  specs: [
```
jayenashar, line 21: - this conf adds the auth-heavy `conductor/**` + `journeys/**` specs but inherits `maxInstances: 10` from the base conf; the headless twin sets 1, and README/SUITE.md say to keep 1 for auth-heavy suites (token-exchange races)
- headed `test:e2e:web` will flake locally where CI passes; `maxInstances: 1` here fixes it

e2e/wdio.conf.ts
```diff
@@ -1,293 +1,71 @@
...
-  // order to group specific specs to a specific capability.
-  //
-  // First, you can define how many instances should be started at the same time. Let's
-  // say you have 3 different capabilities (Chrome, Firefox, and Safari) and you have
-  // set maxInstances to 1; wdio will spawn 3 processes. Therefore, if you have 10 spec
-  // files and you set maxInstances to 10, all spec files will get tested at the same time
-  // and 30 processes will get spawned. The property handles how many capabilities
-  // from the same test should run tests.
-  //
+  specs: ['./test/specs/app/**/*.ts'],
```
jayenashar, line 28: - narrowing the base glob to `app/**` silently changed the legacy entry points (`test:e2e`, `test:e2e:all`, the viewport variants, `test:e2e:headless:all`) to app-only, and no conf now runs web/conductor specs at mobile/tablet viewports
- retiring/renaming the legacy scripts, or noting mobile web coverage as out of scope in SUITE.md, would make the narrowing explicit

api/src/couchdb/emailReset.ts
PeterBaker0: Adding a separate flag and wiring through

.github/workflows/e2e.yml
PeterBaker0: Good call

e2e/test/specs/conductor/password-reset.e2e.ts
PeterBaker0: Wrapped in try/finally

api/src/scripts/seedTestDataset.ts
PeterBaker0: making it idempotent so that it 'restores' the state back to the intended state. This removes the duplication issue but I don't want it to wipe the DB beyond this, for risk of accidental side effects.

e2e/test/helpers/auth.ts
PeterBaker0: Implemented suggested fix

e2e/wdio.web.conf.ts
PeterBaker0: Done

e2e/wdio.conf.ts
PeterBaker0: Just doing doc changes for now, needs further work to unify with the older e2e work.

web/src/routes/_protected.tsx
PeterBaker0: fixed

api/src/buildconfig.ts
```diff
@@ -208,8 +208,20 @@ const EnvSchema = z
...
+     */
     RATE_LIMITER_ENABLED: configHelpers.equalsTrueBool(true),
+    /**
+     * Whether per-user email-code / verification-challenge attempt limits
+     * are enabled (`true` / other). Independent of the HTTP IP limiter so
+     * deployments that disable RATE_LIMITER_ENABLED upstream still keep
+     * password-reset and verification brute-force protection. Default on;
+     * e2e may set false for repeated auth flows.
+     */
+    AUTH_ATTEMPT_LIMITER_ENABLED: configHelpers.equalsTrueBool(true),
```
jayenashar: **Security toggle fails open on a typo.** `equalsTrueBool(true)` resolves *unset* → `true` (safe), but any *defined* value that isn't exactly `true` (`1`, `yes`, or an empty string left after editing) → `false`, which makes `checkCanCreateEmailCode` / `checkCanCreateVerificationChallenge` return `{canCreate: true}` unconditionally and removes per-user password-reset / verification-code brute-force protection — the exact control this field's doc comment says to keep on in production.

`boolWithDefault` (which rejects unrecognized input) would fail-closed on the typo. Consistent with the existing `RATE_LIMITER_ENABLED` handling, so this may be intentional convention — flagging because this one gates a brute-force protection.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,310 @@
...
+
+      - name: Run e2e (smoke + web + app)
+        id: e2e-tests
+        working-directory: e2e
+        env:
+          CHROME_BIN: ${{ steps.setup-chrome.outputs.chrome-path }}
+          CHROMEDRIVER_PATH: ${{ steps.setup-chrome.outputs.chromedriver-path }}
+          WDIO_CHROME_PATH: ${{ steps.setup-chrome.outputs.chrome-path }}
+          WDIO_CHROME_DRIVER: ${{ steps.setup-chrome.outputs.chromedriver-path }}
+        run: pnpm test:e2e:headless:ci
```
jayenashar, line 176: **`&&`-chaining aborts most of the "comprehensive" gate on any smoke hiccup.** `test:e2e:headless:ci` runs `smoke && web && app`, so a single failing or flaky smoke spec exits non-zero here and the web + app suites never run — no JUnit/screenshots for them, and the whole run collapses to one `steps.e2e-tests.outcome` boolean reported as "failed" with web/app coverage silently skipped.

Consider `;` between suites (or separate steps with `continue-on-error`) so all suites run and all artifacts upload.

api/src/scripts/seedTestDataset.ts
```diff
@@ -240,6 +290,243 @@ function clampDescription(description?: string): string | undefined {
...
+  uiSpecification: Record<string, unknown>;
+  teamId: string;
+}): Promise<ExistingTemplateDocument> {
+  const templatesDb = getTemplatesDb();
+  const normalizedUiSpecification =
+    normalizeUiSpecificationOrThrow(uiSpecification);
+  const now = nowIso();
+
+  try {
+    const existing = await getTemplate(id);
```
jayenashar, line 366: **Existence check can't tell a 404 from a transient error, breaking the "safe to re-run" guarantee.** `getTemplate` (and `getProjectById` at line 428) wrap *every* read failure into `ItemNotFoundException`, so `isNotFoundError()` returns `true` on a transient CouchDB/network error against an *existing* seed doc. The code then takes the create branch and `put`s with a fixed `_id` and no `_rev` → 409 conflict → the seed aborts mid-run (teams upserted, users not), leaving a partially-seeded DB despite the new idempotent docstring.

A cheap guard: fetch the raw doc via `.get(id)` directly here (which surfaces the real status) instead of the exception-flattening wrappers, or check for the doc via a list/allDocs lookup.

app/src/gui/components/notebook/add_record_by_type.tsx
```diff
@@ -153,6 +154,7 @@ export default function AddRecordButtons({
                     key={viewset_name}
                     startIcon={<AddCircleSharpIcon />}
                     variant="contained"
+                    data-testid="app-record-add-button"
```
jayenashar: **Duplicate `data-testid` across the multi-type map.** This map branch reuses `app-record-add-button` (same testid as the single-type branch at ~line 144), so a notebook with more than one visible record type renders N buttons with the identical testid. `byTestId('app-record-add-button')` then resolves to the first element and always actuates the first record type. The current single-type `e2e-minimal` notebook doesn't hit this, but it's a latent hole in the selector scheme — suggest suffixing with `viewset_name` here.

jayenashar (commented):
Automated review of the e2e/API changes. Six inline findings below — #1–3 are the ones worth resolving before merge; #4–6 are lower severity. Refuted candidates (BaseMuiTextField testid, finish-button testid, createTeam/createTextRecord persistence) are omitted after checking the page objects and specs actually compensate.

e2e/test/pageobjects/app-records.ts
```diff
@@ -0,0 +1,119 @@
...
+      }
+    );
+    await waitForTestId('app-record-field-notes', {timeout: 20000});
+    const input = await this.notesField.$('input, textarea');
+    await input.waitForDisplayed({timeout: 10000});
+    await input.setValue(notes);
+    await this.finishButton.waitForClickable({timeout: 15000});
+    await this.finishButton.click();
+    // Confirm finish-anyway dialog if validation warns
+    const finishAnyway = await $('button*=Finish anyway');
```
jayenashar, line 112: **Auto-clicking "Finish anyway" masks validation failures.** This force-confirms the validation-override dialog whenever it appears, so a record that fails validation is saved anyway and still counts as created. That means the record-CRUD spec can't catch validation regressions — a genuinely invalid record and a valid one both pass. Fine as a deliberate fallback, but worth asserting no validation dialog appeared for the happy-path case rather than always clicking through.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,310 @@
...
+              `| **Run** | [View workflow run](${runUrl}) |`,
+              '',
+              'After downloading and extracting the .zip file, open `artifacts/index.html` (failures listed first).',
+              '',
+              `_Updated for commit ${context.sha.slice(0, 7)}_`,
+            ].join('\n');
+
+            const {owner, repo} = context.repo;
+            const issue_number = context.issue.number;
+            const {data: comments} = await github.rest.issues.listComments({
```
jayenashar, line 276: **Comment upsert isn't paginated.** `listComments({per_page: 100})` reads only the first page, so on a long-lived branch with >100 comments the `<!-- e2e-artifact-bot -->` marker falls beyond page 1, `existing` is `undefined`, and every run `createComment`s a fresh comment instead of updating — spamming the PR. Use `github.paginate(github.rest.issues.listComments, {...})` or fetch newest-first.

.github/workflows/e2e.yml
PeterBaker0: that's the intended path - if smoke fails why waste time running the rest?

api/src/buildconfig.ts
PeterBaker0: Fixed - throws if invalid, defaults to true.

api/src/scripts/seedTestDataset.ts
PeterBaker0: Eh ... it's fine.

app/src/gui/components/notebook/add_record_by_type.tsx
PeterBaker0: Fair enough

e2e/test/pageobjects/app-records.ts
PeterBaker0: That's fine.

.github/workflows/e2e.yml
PeterBaker0: Eh...

.github/workflows/e2e.yml
jayenashar: - agreed on smoke; one nuance to be aware of: the `&&` fail-fast also applies mid-chain, so a web failure skips the app suite too (a flaky web spec costs app coverage + artifacts for that run); fine if intended

stevecassidy (approved):
Great work!

jayenashar (commented):
Round 3: multi-agent review of the current head (a77db8e), findings verified individually; 8 confirmed, worst first in the inline comments.

e2e/wdio.conf.ts
```diff
@@ -1,293 +1,71 @@
...
+  exclude: [],
   maxInstances: 10,
```
jayenashar, line 30: **Base conf still runs headed app suites at `maxInstances: 10`.** `test:e2e`, `test:e2e:app`, and the four viewport variants all run this conf, and all 5 app specs log in during setup, so up to 5 workers race the token exchange: the same race `wdio.web.conf.ts`, `wdio.smoke.conf.ts`, and the headless confs pin `maxInstances: 1` to avoid. Headed app runs will flake where CI passes; `maxInstances: 1` here matches the rest of the suite.

e2e/test/specs/web/project-invites.e2e.ts
```diff
@@ -0,0 +1,102 @@
...
+  const inviteName = `E2E Proj Invite ${Date.now()}`;
+
+  before(async () => {
+    await browser.reloadSession();
+    await loginWebPersona('managerBlue');
+  });
+
+  it('should open Invites tab and create a project invite', async () => {
+    await openFirstProject();
+    const invitesTab = await $('button*=Invites');
```
jayenashar, line 62: **Text selectors where this PR already ships the testids.** 12+ selectors across 9 web specs use `*=` visible-text matching (`button*=Invites` here; also permissions-matrix 83/102/120, team-invites 63/75, users-admin 147/162, project-status-archive 25, template-visibility-archive 24, designer-basic 35, teams 66, templates-create 25) while the matching testids exist (`web-project-tab-invites`/`-actions`/`-details`, `web-nav-templates`) and SUITE.md declares testids the required mechanism. A label edit or i18n silently breaks these, and substring matching can hit an unintended button; `byTestId('web-project-tab-invites')` etc. fixes.

web/src/components/data-table/data-table.tsx
```diff
@@ -171,6 +172,7 @@ export function DataTable<TData, TValue>({
                 <TableCell
                   colSpan={columns.length}
                   className="h-24 text-center"
+                  data-testid="web-data-table-empty"
```
jayenashar, line 175: **Shared DataTable rows got no testid, so specs scrape hrefs.** teams.e2e.ts:58 (`a[href*="/teams/"]`) and designer-basic.e2e.ts:25 (`a[href*="/templates/"]`) navigate by href shape, which any route change breaks, and first-match anchor picking is nondeterministic. The app side already established the pattern this needs: `app-notebook-row` on tab-grid.tsx:137, clicked first-match by the page object. One `data-testid` on the body `<TableRow>` (~line 143) covers both specs.

e2e/test/helpers/auth.ts
```diff
@@ -0,0 +1,211 @@
...
+ */
+export async function loginWeb(user: Credentials): Promise<void> {
+  const webUrl = getWebUrl();
+  // Navigate to dashboard root (absolute URL — baseUrl may be app or web).
+  await browser.url(webUrl);
+
+  // The protected route redirects unauthenticated users to the API login page.
+  await waitForUrl('/login', {
+    timeout: 15000,
+    timeoutMsg:
```
jayenashar, line 76: **Full UI login runs in the `before` of nearly every describe.** 24 spec files / 77 `loginWeb*`/`loginApp*` call sites each drive the multi-redirect UI login (~5-15s each per CI run). The web session is a pure localStorage JWT (`auth-provider.tsx`), and a token is mintable headlessly (`POST /auth/local`, then `POST /api/auth/exchange`), so a login-once-per-persona token-injection helper would cut dozens of UI logins per run, keeping real UI logins only in the login-focused specs. (Note `getWebAuthSession` in seed.ts only reads a token after `loginWeb`; the minting helper would be new.)

e2e/wdio.headless.smoke.conf.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Tier 0 smoke suite headless (CI gate).
+ */
+export const config = {
+  ...baseConfig,
+  specs: ['./test/specs/smoke/**/*.ts'],
+  baseUrl: getAppUrl(),
+  maxInstances: 1,
+  autoXvfb: false,
+  capabilities: [...headlessChromeCapabilities],
```
jayenashar, line 18: **The three headless confs triplicate config that's also already inherited.** They repeat `maxInstances: 1`, `autoXvfb: false`, the capabilities array, and a byte-identical junit `reporters` block, and that reporters block is also inherited from `wdio.conf.ts:44-56` via `...baseConfig`, so it's declared four times. A shared headless base conf (the spread pattern the suite already uses) leaves only `specs`/`baseUrl` per file.

e2e/test/helpers/selectors.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Locate an element by exact data-testid.
+ */
+export function byTestId(testId: string) {
+  return $(`[data-testid="${testId}"]`);
+}
+
+/**
+ * Locate an element whose data-testid contains the given substring.
+ */
+export function byTestIdContaining(partial: string) {
```
jayenashar, line 13: **Dead helpers with zero call sites** (grepped at head): `byTestIdContaining` + `testIdSelector` here, `tryGetPersona` (env.ts:193), `captureRaw` (screenshot.ts:141), `createGlobalInvite` (seed.ts:225), the `SCREEN_WIDTH`/`SCREEN_HEIGHT` branch (viewport.ts:37), and `waitForNetworkIdle` (wait.ts:68). They widen the helper surface, and `byTestIdContaining` invites the substring matching the README discourages; suggest deleting them.

web/src/components/forms/create-global-invite-form.tsx
```diff
@@ -133,9 +133,16 @@ export function CreateGlobalInviteForm({
 
   return (
     <Form
-      fields={fields}
+      fields={fields.map(field =>
```
jayenashar, line 136: **testIds injected by name-matching post-processing.** `fields.map(f => f.name === 'name' ? {...f, testId} : ...)` string-matches field names, so renaming a field silently drops its testid. The fields array is a local literal (lines 46-79), and the five sibling forms in this PR set `testId` inline on the field definition; same here keeps one pattern.

e2e/test/helpers/report.ts
```diff
@@ -0,0 +1,667 @@
...
+  </header>
+  ${
+    thumbs
+      ? `<div class="thumbs">${thumbs}</div>`
+      : '<p class="muted">No step screenshots</p>'
+  }
+</section>`;
+}
+
+function slugAnchor(spec: string, test: string): string {
```
jayenashar, line 199: **`slugAnchor` duplicates `slugify`.** Same transform as `slugify` in artifacts.ts (already imported by this file) but with an 80 vs 100 length cap, so gallery anchor ids can diverge from artifact directory slugs; call `slugify` instead.

jayenashar (approved):

PR #2192, opened 2026-07-14
fix: intermittent api unit test failure

PeterBaker0:
Due to timestamp comparisons occurring from two different spots with a Date.now() call. Simple fix.

### DIFF
--- a/library/data-model/tests/migrations.test.ts
+++ b/library/data-model/tests/migrations.test.ts
@@ -1939,7 +1939,7 @@ const AUTH_V3_TO_V4_MIGRATION_TEST_CASES: MigrationTestCase[] = [
       userId: 'user123',
       code: 'hashed_code_abc',
       used: false,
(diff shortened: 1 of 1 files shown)


jayenashar:
- ✅ hoisted const kills the flake (single evaluation shared by input and expected)
- the custom `equalityFunction` now only checks `createdTimestampMs` is *defined*, so a regression writing e.g. `0` or copying the wrong field would still pass
- alternative: freeze the clock and assert exactly, which drops the custom `equalityFunction` in favour of plain `areDocsEqual`:

  ```ts
  const FIXED_NOW = Date.now();
  beforeAll(() => jest.spyOn(Date, 'now').mockReturnValue(FIXED_NOW));
  afterAll(() => jest.restoreAllMocks());
  ```

  expected doc then states `createdTimestampMs: FIXED_NOW`
- ❓ worth tightening here, or happy to merge as-is?

stevecassidy (approved):

PR #2194, opened 2026-07-14
feat: refactor condition editor

xinyuhoulucy:
# feat: refactor condition editor

## Ticket

[BSS-1281](https://jira.csiro.au/browse/BSS-1281)

## Description

Refactored the condition editor to implement the redesigned condition editor, including draggable condition groups and condition rule rows.

## Proposed Changes

- Added editor tree support for condition groups and rule rows.
- Added drag-and-drop support for groups and rules.
- Added operator tooltips and updated condition summary display.
- Updated related condition editor tests.

## How to Test

Test all condition editor functionality and use cases, including adding, editing, deleting, grouping, ungrouping, dragging condition groups and rule rows, and verifying the correctness of the condition summary text.

## Checklist


PeterBaker0:
image

Some overflow issues on the preview

PeterBaker0:
image

Missing drag target from a group or single item INTO a single item (to form a new group)

PeterBaker0:
image


Would be good to have grab handle aligned with others on right side? 

Maybe consult with Elana

PeterBaker0:
image

When creating a new group, rather than adding an empty placeholder field, just add a 'skeleton' or a 'prompt' visual feedback that the user can add new fields to this group. The reason being is that they may just want to make a group then drag a field into it. "Grouping" as an action being conflated with "Creating a field" is in my opinion, confusing.

PeterBaker0 (changes requested):
HUGE improvement already, looking amazing!

Some feedback to address, then let's have another look. 

If anything is too hard/not feasible, please push back and we can discuss :)

xinyuhoulucy:
> image
> When creating a new group, rather than adding an empty placeholder field, just add a 'skeleton' or a 'prompt' visual feedback that the user can add new fields to this group. The reason being is that they may just want to make a group then drag a field into it. "Grouping" as an action being conflated with "Creating a field" is in my opinion, confusing.

Good idea! let me try!!

xinyuhoulucy:
The issues above should all be fixed or improved now.

Also tried to align the button order in the condition groups and condition rule rows for a more consistent layout.

Personally I find it more convenient to keep all the group level actions together (Ungroup, Delete group, and Drag/move group), but still leaving “Add condition” at the bottom of the group. Since new conditions are added to the end of the list, that button position makes more sense to me.

Let me know if there are any other suggestions, or if the previous layout looked or felt better :)

PeterBaker0 (approved):
Working well now - as discussed some changes to do later with simple editor but this is a tremendous improvement. Well done.

PR #2195, opened 2026-07-14
refactor: replace axios with native fetch in api (#2050)

### LINKED ISSUE #2050: Replace single use of axios in api with fetch
We shouldn't need axios for a single request when we have fetch built in to node.

Caedous:
# refactor: replace axios with native fetch in api (#2050)
## JIRA Ticket
N/A (GitHub issue #2050)
## Description
The API used axios for a single request: the PUT that pushes the JWT public key into CouchDB's config during database initialisation. Node's built-in fetch handles this, so axios has been removed as a dependency of the API.
## Proposed Changes
- Rewrote `api/src/auth/keySigning/initJWTKeys.ts` to use native fetch instead of axios
- Built the basic auth header manually and added an explicit `response.ok` check, since fetch has no `auth` option and does not throw on HTTP error statuses the way axios does
- Removed `axios` from `api/package.json` and updated the lockfile

Two small additional fixes surfaced during testing:
- Stripped `\r` from the public key before writing to CouchDB config. This fixes a pre-existing bug (reproducible on the original axios code) where PEM key files with Windows (CRLF) line endings cause CouchDB to reject the PUT with `400 Invalid configuration value`. Happy to split this into a separate issue/PR if preferred.
- The catch block now logs the underlying error before rethrowing — the previous behaviour of discarding it made the above failure hard to diagnose.
## How to Test
1. From `api/`, run `pnpm migrate-with-keys` — output should include `JWT public key configured in CouchDB` and the migration should complete successfully
2. Verify the key landed: `curl -u <user>:<password> http://localhost:5984/_node/_local/_config/jwt_keys` should return an `rsa:<kid>` entry containing the public key with `\n` separators and no `\r` characters
3. Start the API and log in to the app — token validation works end to end
4. Negative path: set an incorrect `COUCHDB_PASSWORD`, re-run `pnpm migrate-with-keys` — it fails with a logged error rather than reporting success
## Additional Information
Closes #2050
## Checklist

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -35,7 +35,6 @@
     "@types/passport-openidconnect": "^0.1.3",
     "archiver": "^6.0.1",
     "aws-sdk": "^2.1664.0",
-    "axios": "^1.15.2",
     "body-parser": "1.20.3",
     "cache-manager": "^5.7.3",
     "cookie-session": "2.1.1",
--- a/api/src/auth/keySigning/initJWTKeys.ts
+++ b/api/src/auth/keySigning/initJWTKeys.ts
@@ -1,4 +1,3 @@
-import axios from 'axios';
 import {config, keyService} from '../../buildconfig';
 
 export async function initialiseJWTKey(): Promise<void> {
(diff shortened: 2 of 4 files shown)


PeterBaker0 (approved):
Lgtm - thanks for this change. Tested.

PR #2198, opened 2026-07-17
Add TEMPLATE_MANAGER role and give it to TEAM_MANAGER

stevecassidy:
# Add TEMPLATE_MANAGER role and give it to TEAM_MANAGER for all team templates

## Ticket

None

## Description

TEAM_MANAGER was able to create templates within a team but could not see existing templates. 

## Proposed Changes

Added a new TEMPLATE_MANAGER role that has what used to be the permissions of TEMPLATE_ADMIN without template deletion.  Made this a virtual role for TEAM_MANAGER so that team managers can see and manage templates within the team but can't delete them.   TEAM_ADMIN is required to delete templates via TEMPLATE_ADMIN. 

Adds tests around this new distinction and new users in the seed data with the admin and manager roles within teams.

## How to Test

After running the seed-test-dataset script, as user seed-manager-red you should be able to see the sample template in the red team.  You should not be able to delete it.

## Checklist

### DIFF
--- a/api/src/scripts/seedTestDataset.ts
+++ b/api/src/scripts/seedTestDataset.ts
@@ -160,15 +160,39 @@ const USER_SPECS: UserSpec[] = [
     },
   },
 
-  // ── seed-manager-cross ────────────────────────────────────────────────────
+  // ── seed-manager-blue ────────────────────────────────────────────────────
   // TEAM_MANAGER on Blue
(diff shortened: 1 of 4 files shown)


PeterBaker0 (approved):
Roger 👍 Any docs need updating in the main docs build? Happy to be deferred ticket if need be to get this working.

Issue #2197, opened 2026-07-14
Expanded computed field capability


PeterBaker0:
**Is your feature request related to a problem? Please describe.**

The current computed is limited to numerical types in input and output. Ideally we would offer more flexibility here to support use cases that e.g. Survey123/Fulcrum can support. 

**Describe the solution you'd like**

Ability to compute over strings, numbers and booleans and output strings, numbers and booleans. This would then allow us to build a more general expression engine based on this work, for informing things such as boundary conditions for sliders (e.g. XLSForm constraints). 

An example of a field (pseudo code) might be:

```
fireRisk := if vegType = 'hedge' then (if vegHeigh < 3 then 'Low' else if vegHeight <5 then 'Medium' else 'High') else 'Medium'
```

Caedous:
A couple of questions to help me work out the direction this needs to go:

1. With strings, numbers and booleans in play, what should mixed-type operations do? My inclination is to be strict and reject them at compile/eval time (e.g. `{vegType} * 2` where vegType is a string is an error, not `NaN`), rather than inherit JavaScript's coercion rules. Related: should `+` concatenate strings, or should concatenation be a separate operator so `+` stays unambiguously numeric?
2. The field currently declares `type-returned: faims-core::Number` statically in the spec. If an expression can return a string or boolean, we need to decide how that's expressed. Options I can see are (a) the author declares the return type in the field config and we validate the expression against it, or (b) the field's return type becomes a union. (a) keeps validation and downstream consumers well-defined, but adds a config step. Any preference?
3. The example uses `if ... then ... else`. The engine currently usesthe ternary (`{vegHeight} < 3 ? 'Low' : 'Medium'`) via jsep's grammar. Reusing the ternary is much cheaper than adding an if/then/else form (which means custom parsing). Is the ternary acceptable, or is the friendlier syntax a requirement?
4. Is there an idea that this engine eventually backs conditions/validations as well?

Also noting the current field references fields in braces (`{Field-ID}`) to handle IDs containing hyphens.

PeterBaker0:
@Caedous thanks - good qns

> With strings, numbers and booleans in play, what should mixed-type operations do? My inclination is to be strict and reject them at compile/eval time (e.g. {vegType} * 2 where vegType is a string is an error, not NaN), rather than inherit JavaScript's coercion rules. Related: should + concatenate strings, or should concatenation be a separate operator so + stays unambiguously numeric?

[...]

PR #2199, opened 2026-07-20
feat: typed computed fields - ComputedNumber and ComputedText (#2197)

Caedous:
## Ticket
#2197

## Description
Extends the computed field expression engine to handle strings, numbers and booleans in and out, and splits the field into ComputedNumber and ComputedText so each has a fixed `type-returned`. Typing is strict with no implicit coercion, per Peter's direction on the issue: ill-typed expressions are rejected when the notebook is designed rather than producing a silently wrong value, and the two-field split avoids union return types that would complicate export flows such as GeoPackage.

Operators are discrete and unambiguous: `+ - * / %` are numeric only, `&` is string concatenation, `&& || !` are boolean only, `< > <= >=` compare two numbers or two strings, `== !=` require matching types, and the ternary requires a boolean condition with same-typed branches. Mixed-type inputs are fine so long as the overall result matches the field's type — `{vegType} == 'test' || {flag} ? 3 : 5` is a valid ComputedNumber, while `... ? 3 : 'Not valid'` is rejected. Field references keep the brace syntax (`{Field-ID}`).

## Proposed Changes
The engine in data-model compiles each expression bottom-up into a typed closure, inferring every node's type at compile time and selecting operator implementations once. Field reference types come from the spec's `type-returned` values via a shared `FAIMS_TYPE_TO_EXPR_TYPE` map, built at notebook load in `compileUiSpecConditionals`, and each component's expression is validated against its required return type (number for ComputedNumber, string for ComputedText). Mismatches throw `ExpressionError` naming the types involved. The source cache is removed, since the same source legitimately compiles differently against different field types and compile-once-at-load makes it redundant.

In forms, ComputedField is renamed to ComputedNumber and ComputedText is added: one shared read-only component with per-type value schemas in a single `ComputedFields` folder, following the `DateFields` multi-spec pattern. The recompute resolves scope values per each referenced field's declared type instead of coercing everything to `Number`; an empty string counts as missing, so partially filled forms stay blank rather than showing a partial result.

The Designer gains two chooser cards (Computed Number under NUMBERS, Computed Text under TEXT) sharing one editor. The insert chips now list all referenceable fields — number, text and checkbox, with derived fields excluded — and the helper text shows a per-variant example. Invalid expressions are compiled as the user types and the error is shown inline under the expression input; if one still gets through via raw JSON, the uiSpec compilation logs it and the field stays blank.

inline errors

The expression tests are rewritten for the typed engine (26 cases) covering the ticket's fireRisk example, concatenation, string comparison, mixed-type inputs with a matching result type, and the compile-time error battery: type mismatches per operator, ternary rules, wrong return type, unknown fields, and brace validation. The user documentation is updated, replacing the Computed Value page with Computed Number and adding a Computed Text page under text fields.

## How to Test
1. Build data-model, forms and web; run the data-model and forms test suites.
2. In the Designer, open the ADD A FIELD dialog: Computed Number appears on the NUMBERS tab and Computed Text on the TEXT tab.

computed number

computed text

3. Regression: add two number fields and a Computed Number with an arithmetic expression; the chips insert braced references and the preview computes as before.

computed number configured

4. Recreate the ticket's fireRisk example as a Computed Text, with a text field `vegType` and a number field `vegHeight`, using the expression `{vegType} == 'hedge' ? ({vegHeight} < 3 ? 'Low' : ({vegHeight} < 5 ? 'Medium' : 'High')) : 'Medium'`. Entering hedge/2 shows Low, hedge/4 Medium, hedge/6 High, and tree with any height Medium.

fireRisk example

fireRisk example 2

fireRisk example 3

fireRisk example 4

5. Concatenation: a Computed Text with `{vegType} & '-test'` renders `tree-test` and stays blank until vegType has a value.

concat

6. Strict typing: enter `{vegHeight} & 'x'` as a Computed Text expression. The console reports `ExpressionError: & (concatenation) requires strings, got number and string` at compile time and the field stays blank.

invalid expression

## Checklist


docs/user/field-types/number-fields/computed-number-field.md
```diff
@@ -0,0 +1,109 @@
...
+:align: right
+:width: 100%
+```
+
+### Computed Number-Specific Settings
+
+The Computed Number's key feature is the **Expression** text area, which
+defines the calculated value. Each referenced field is written by
+wrapping its Field ID in braces, e.g.
+`{Wet-Soil-Mass-g} * {Number-of-Samples}`. Below the expression, a chip
```
PeterBaker0: "Chip is shown" is out of date

docs/user/field-types/number-fields/computed-number-field.md
```diff
@@ -0,0 +1,109 @@
...
+is shown for each referenceable field in the form; clicking a chip
+inserts that field's reference, which avoids typing the Field ID by
+hand.
+
+| Setting          | What It Does                                                                                                                                      |
+| ---------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **Expression**   | A text area where you define the calculation using literals, operators, and field references in single-brace syntax (e.g., `{Width} * {Height}`). |
+| **Insert field** | A searchable picker of the referenceable fields in the form. Selecting one inserts its braced reference into the expression.                      |
+
+The expression is typed and checked when the {{notebook}} is designed —
```
PeterBaker0, line 65: "rather than a silent wrong answer" reads like an agent correcting an earlier prompt

docs/user/field-types/number-fields/computed-number-field.md
```diff
@@ -0,0 +1,109 @@
...
+Configure any of the shared field options as needed.
+
+For settings shared across all field types — including Annotation,
+Uncertainty, Conditions, Copy value to new records, and Display in child
+records — see
+[Field Options](../shared-settings/field-options.md).
+
+## Tips
+
+- **Use the field picker to build expressions** rather than typing Field
```
PeterBaker0, line 98: out of date

docs/user/field-types/number-fields/computed-number-field.md
```diff
@@ -0,0 +1,109 @@
...
+  IDs by hand. Clicking a field inserts the exact braced reference, so you
+  avoid typos and do not need to remember a field's ID.
+- **The result stays blank until all referenced fields have values.**
+  This is intentional — a partially complete record shows no value rather
+  than a misleading partial calculation.
+- **Type mismatches are reported when the {{notebook}} is designed**, not
```
PeterBaker0: rewrite to not refer to previous lesser implementations - "not hidden at collection time" is not necessary

PeterBaker0 (changes requested):
Some documentation updates. 

We should also write a migration to move the old ComputedValue -> ComputedNumber. While there might not be any in the wild, would be good to do so to follow best practise here for backwards compatibility. There is existing hooks for 'migrate notebooks on startup' which run these migrations.

web/src/designer/components/Fields/ComputedFieldEditor.tsx
```diff
@@ -104,35 +129,44 @@ export const ComputedFieldEditor = ({fieldName, viewsetId}: PropType) => {
...
-            number, percentage, or other numeric fields to reference them in the
-            expression.
+            No referenceable fields in this form. Add number, text, or checkbox
+            fields to reference them in the expression.
           </Alert>
         )}
         <FormHelperText>
-          Reference other numeric fields by wrapping their ID in braces.
-          {numericFieldCount > 0 &&
+          Reference other fields by wrapping their ID in braces.
```
stevecassidy, line 214: I think some inline help would be useful on writing expressions, maybe just the list of operators you have in the docs might be enough.

stevecassidy (commented):
One thing that's missing is a check when fields are deleted that they don't occur in a computed expression. We have this for conditions, see findFieldDependencyReferences, and field delete checks for references before allowing deletion.  We should have the same for computed fields.

web/src/designer/components/Fields/ComputedFieldEditor.tsx
Caedous: Added — a tolerant `extractExpressionReferences` in data-model (works on malformed/in-progress expressions) feeds `findFieldDependencyReferences` as a `computed-expression` reference type, so deleting a referenced field is blocked with the same dialog as conditions and templated strings.

cannot delete

Added a collapsed "Expression syntax" expander under the input with the operator list from the docs and the variant's required result type.

syntax

PeterBaker0 (approved):
Happy with this now, as it's addressed migrations, improved docs, and has additional ref safety.

stevecassidy (approved):
All looks good, thanks for the work!

PR #2204, opened 2026-07-23
chore: couch db version upgrade guide and instrumentation

PeterBaker0:
## Summary

- Add an AWS CouchDB version-upgrade runbook plus helper scripts (baseline, EBS snapshot, stop/detach for replace).
- Force Couch EC2 replacement on UserData changes (`userDataCausesReplacement`) and document the post-replace keys/migrate flow. **NOTE** this will force an instance replacement at deploy time even if you don't change the version.
- Add `dump-user-emails` for maintenance notices; have `env-from-cdk-stack.sh` emit `AWS_DEFAULT_REGION`.

Tested out on our stage replica. Works well. Note in the above that if you don't have the userDataCausesReplacement you don't get an EC2 replacement with a new user data run (to build the new couch version) but it does mean for current builds you'll get an instance replacement which is a bit risky if not coordinated well. Future ones won't experience that unless the user data hash changes (which it shouldn't typically). 

**Note** if you don't detach the volume prior to running the instance replacement, you'll get a CFN error about the volume being already attached. Known CFN bug we have to work around.


infrastructure/aws-cdk/scripts/env-from-cdk-stack.sh
```diff
@@ -0,0 +1,278 @@
+#!/usr/bin/env bash
```
stevecassidy, line 1: Can this be in the AWS folder please.

api/src/scripts/dumpUserEmails.ts
```diff
@@ -0,0 +1,189 @@
+/* eslint-disable n/no-process-exit */
```
stevecassidy, line 1: This facility would be useful but better as an API endpoint for SYSTEM_ADMIN so that we can easily access it on a given server. Maybe a separate PR.

infrastructure/aws-cdk/scripts/notes.txt
```diff
@@ -0,0 +1,77 @@
+{
```
stevecassidy: Was this supposed to be here?

docs/developer/docs/source/markdown/CouchVersionUpgradeGuideAWS.md
```diff
@@ -0,0 +1,198 @@
+# CouchDB version upgrade on AWS
```
stevecassidy, line 1: This is AWS specific but the 'migrate-with-keys' step after updating seems general, might be worth pointing that out near the top along with any other general issues.

stevecassidy (approved):
Looks ok with a few comments.

api/src/scripts/dumpUserEmails.ts
PeterBaker0: Yep, separate PR for now - made ticket in our system

infrastructure/aws-cdk/scripts/notes.txt
PeterBaker0: Nope :/

PR #2212, opened 2026-07-26
refactor: extract record GIS feature hydration into a shared useRecordFeatures module

### LINKED ISSUE #2226: Record GIS feature hydration is locked inside OverviewMap
Record GIS geometry extraction lives inline in `OverviewMap.tsx`, so nothing else in the app can hydrate GeoJSON features from records, and it has no tests. The Map tab also has no e2e coverage (no spec ever opens the tab). Extract a shared, tested module and cover the tab.

🤖 Written with [Claude Code](https://claude.com/claude-code)

jayenashar:
# refactor: extract record GIS feature hydration into a shared useRecordFeatures module

## Ticket

Closes #2226.

## Description

Record GIS geometry extraction lived inline in `OverviewMap.tsx`, so nothing else could hydrate GeoJSON features from records. This moves it into its own module with tests; `OverviewMap` consumes it and renders exactly as before. The Map tab also gains its first e2e coverage (no spec ever opened the tab).

## Proposed Changes

- New `recordFeatures.ts` (`getGISFields`, `extractFeaturesFromRecord`, the `useRecordFeatures` hook) with 10 tests; `OverviewMap.tsx` consumes the hook with rendering, props, and mount contract untouched.
- New e2e spec `app/notebook-map.e2e.ts`; the seeded e2e notebook gains an optional `TakePoint` field, and a helper stubs `navigator.geolocation` (headless Chrome has no location provider).

Three decisions worth review:

1. **Extraction is notebook-wide, not per-form**: the designer's field moves can leave a record's geometry unreachable from its own form, and per-form scoping silently drops it.
2. **A missing revision or AVP skips only its own record or field**; any other read failure rethrows, so a systemic fault surfaces instead of an empty map.
3. **The hook returns named props rather than spreading the React Query result**, so `OverviewMap` stops re-rendering on tracked props it never reads.

The query key prefix renames to `record-features`; nothing invalidates the old key, so the rename is inert.

## How to Test

`pnpm --filter @faims3/app compile && pnpm --filter @faims3/app test && pnpm --filter @faims3/e2e typecheck && pnpm run format:check`; CI's Headless Chromium suite runs the new e2e spec. Manually: open the Map tab on a notebook with geometry (records plot by form type, pin popover opens); notebooks with no GIS fields or no geometry keep their empty states.

## Additional Information

The e2e's load-bearing assertion is the absence of the "No records with location data found" empty state, which renders exactly when hydration yields zero features. No basemap key is needed; OpenLayers builds its viewport whether or not a tile loads.

## Checklist


app/src/gui/components/notebook/OverviewMap.tsx
```diff
@@ -64,12 +66,7 @@ interface OverviewMapProps {
   records: {allRecords: MinimalRecordMetadata[]};
 }
 
-interface FeatureProps {
-  name: string;
-  record_id: string;
-  revision_id: string;
-  form_id: string;
-}
+type FeatureProps = RecordFeatureProps;
```
PeterBaker0: Don't redefine type here just use the imported one.

app/src/gui/components/notebook/OverviewMap.tsx
```diff
@@ -85,28 +82,7 @@ const FORM_TYPE_COLORS = [
...
- */
-const getGISFields = (uiSpec: NotebookUiSpec): string[] => {
-  const fields = Object.getOwnPropertyNames(uiSpec.fields);
-  return fields.filter(
-    (field: string) =>
-      uiSpec.fields[field]['component-name'] === 'MapFormField' ||
-      uiSpec.fields[field]['component-name'] === 'TakePoint'
-  );
-};
+type FeatureCollection = RecordFeatureCollection;
```
PeterBaker0: Same here

app/src/gui/components/notebook/recordFeatures.test.tsx
```diff
@@ -0,0 +1,293 @@
...
+  },
+  viewsets: {
+    Site: {views: ['site-section']},
+    Observation: {views: ['obs-section']},
+  },
+  visible_types: ['Site', 'Observation'],
+  settings: {showQrCodeButton: false},
+  schemaVersion: '1',
+  conditional_sources: new Set<string>(),
+} as unknown as CompiledNotebookUiSpec;
```
PeterBaker0: Why do we need to type cast as unknown here? This is risky for long term as the ui spec shape drifts through time - better to do a proper one?

app/src/gui/components/notebook/recordFeatures.test.tsx
```diff
@@ -0,0 +1,293 @@
...
+  type: string
+): MinimalRecordMetadata =>
+  ({
+    projectId: 'test-project',
+    recordId,
+    revisionId,
+    type,
+    conflicts: false,
+    deleted: false,
+  }) as MinimalRecordMetadata;
```
PeterBaker0: type casting?

app/src/gui/components/notebook/recordFeatures.ts
```diff
@@ -0,0 +1,260 @@
...
+  type: string;
+  geometry?: unknown;
+  properties?: RecordFeatureProps;
+};
+
+/** The extracted features of all records, as one GeoJSON FeatureCollection. */
+export type RecordFeatureCollection = {
+  type: 'FeatureCollection';
+  features: RecordGeoJSONFeature[];
+};
```
PeterBaker0: I'm pretty sure there are zod models for geojson features already part of the data model - maybe have a look around, we have definitely defined these before. Also possibly part of the TakePoint fields. Would be worth hoisting out to common types as beginning of effort to unify these.

app/src/gui/components/notebook/recordFeatures.ts
```diff
@@ -0,0 +1,260 @@
...
+      gisFields.join(','),
+      recordTypes.join(','),
+    ],
+    queryFn: fetchAllFeatures,
+    enabled: gisFields.length > 0 && mapRecords.length > 0,
+    staleTime: 5 * 60 * 1000, // Consider data fresh for 5 minutes
+    gcTime: 10 * 60 * 1000, // Keep in cache for 10 minutes
+    retry: 2,
+    retryDelay: attemptIndex => Math.min(1000 * 2 ** attemptIndex, 10000),
+  });
```
PeterBaker0: This seems okay, but also might be worth considering useQueries ? Then you would have per record query caching and inbuilt parallelism.

e2e/test/helpers/stubLocation.ts
PeterBaker0, line 1: can you improve filename - maybe stubLocation.ts

PeterBaker0 (changes requested):
Overall I think it's an improvement. 

What I'm wondering at a higher level is 

a) what belongs in the data engine
b) what belongs in the data model
c) what about geospatial export logic which overlaps a lot with this?

Not against merging once suggested changes made, but is it worth considering above?

app/src/gui/components/notebook/recordFeatures.test.tsx
jayenashar: ✅ 1fecda6 types the fixture as `NotebookUiSpec` and runs it through `compileUiSpecConditionals`, mirroring the app's `compiledSpecService` (same single cast, since the compiler attaches its artifacts in place). Spec-shape drift now fails the type check.

app/src/gui/components/notebook/recordFeatures.test.tsx
jayenashar: ✅ Dropped in 1fecda6; the fixture now carries full `MinimalRecordMetadata`.

app/src/gui/components/notebook/recordFeatures.ts
jayenashar: ✅ Switched in 7a87da6: one query per record with a `combine` that preserves the hook's contract. A record syncing in now re-reads only itself; unchanged records serve from cache (the old single query re-hydrated every record whenever any record changed).

Tradeoffs: one cache entry per record (10 min gcTime), and the batch-of-10 throttle is gone; per-record reads all fire together, serialized by the PouchDB adapter underneath.

app/src/gui/components/notebook/recordFeatures.ts
jayenashar: The GeoJSON zod live in `@faims3/forms` (`lib/components/maps/types.ts`, the Map field's value parse) rather than data-model; TakePoint stores the same Feature shape. ✅ 203e095 exports the inferred `GeoJSONFeature` type alongside the schema there and derives this module's types from it, so the hand-rolled shapes are gone.

Should full unification be a follow-up?  Hoist the schemas from forms into data-model as the common types, then share one spatial-extraction path in the data engine between this hook and the api's geospatial export (`geospatialExport.ts` has its own `ExtractedGeometry` and Feature/FeatureCollection branching today).

jayenashar:
> a) what belongs in the data engine b) what belongs in the data model c) what about geospatial export logic which overlaps a lot with this?

- data model: the GeoJSON schemas/types (today in `@faims3/forms`) (maybe outside this PR)
- data engine: the record-walking spatial extraction, shared between this hook and the api's export
- app: only the React Query hook

app/src/gui/components/notebook/recordFeatures.ts
PeterBaker0: Yes - please spin off separate ticket. Well worth unifying into a spatial component of the data engine - e.g. dataEngine.spatial.getSpatialFields(), getSpatialDataForField(), getAllSpatialData() etc.

PeterBaker0 (approved):
Seems to work well and comments have been addressed. As noted for spin off is the need for a more unified spatial processing pipeline.

PR #2215, opened 2026-07-27
fix: distinguish an unloaded or incomplete record list from an empty one

### LINKED ISSUE #2224: Record list cannot distinguish an unloaded or incomplete list from an empty one
`useRecordList` reports no difference between "this project has no records", "the list has not loaded" (including a failed initial fetch, which TanStack presents as a loaded empty list), and "a pull is still delivering records". Any UI that reads a record's absence as meaningful is silently wrong: an errored load renders as an empty notebook, and a mid-download project looks complete.

🤖 Written with [Claude Code](https://claude.com/claude-code)

jayenashar:
# fix: distinguish an unloaded or incomplete record list from an empty one

## Ticket

Closes #2224.

## Description

`useRecordList` could not tell "this project has no records" from "the list has not
loaded" or "records are still arriving", so any UI reading a record's absence as
meaningful was silently wrong. Three additive signals separate them:

- **`isLoading`** — true only while the list has never loaded (`data === undefined`).
  Consumers used `initialQuery.isLoading`, which TanStack sets false as soon as the
  initial fetch *fails*, presenting the empty fallback as a loaded, empty notebook.
- **`canReadAllRecords`** — the query drops records the user may not read, and
  consumers could not tell. Computed from the same token the filter reads.
- **`useIsRecordDownloadUnderway`** — true while a pull may still be delivering
  records, backed by a new `SyncState.isPullCaughtUp`. False when the project never
  pulls, and on error or denial, so an offline device keeps working from local data.

That marker needed two fixes to be trustworthy:

- `change.pending` was typed as required, but PouchDB only populates it when the
  source reports it, and a missing value collapsed to `0`, letting the first batch of
  a bulk download declare the pull caught up. Now optional; `pendingRecords` still
  displays `0`.
- A clean `paused` was treated as a catch-up. `PouchDB.sync` re-emits its children's
  pause with the error stripped, so a device that went offline mid-download looked
  finished. A new `pullPaused`, on the pull side's own emitter, is the only thing that
  raises the marker; the aggregate `paused` still drives the visible sync status.
  Per-direction state belongs on `sync.pull` by design
  ([apache/pouchdb#4592](https://github.com/apache/pouchdb/issues/4592)).

## Proposed Changes

- `utils/customHooks.tsx`: `useRecordList` returns `isLoading` and
  `canReadAllRecords`; new `useIsRecordDownloadUnderway`.
- `slices/helpers/syncStateService.ts`: `isPullCaughtUp`, moved only by the pull's own
  signals; new `recordPullPause`.
- `slices/helpers/databaseHelpers.ts`: `pending` optional; `pullPaused` routed to the
  pull side's own emitter.
- `slices/projectSlice.ts`: wire `pullPaused`.
- `notebook/index.tsx`: the record tables gate on `isLoading`.
- Tests: `syncStateService.test.ts`, `customHooks.test.ts`,
  `databaseHelpers.replication.test.ts`.

## How to Test

1. Open a notebook offline so the initial fetch fails: the tables stay loading and
   retry instead of showing an empty list.
2. Load it, then break the network so a background refetch fails: the loaded list
   keeps being served.
3. Activate a notebook with server-side records: `useIsRecordDownloadUnderway` is true
   through the initial pull, false once caught up, and false if the device goes offline
   mid-download.

## Additional Information

The only existing consumer whose behaviour changes is the notebook record tables,
moving from `initialQuery.isLoading` to `isLoading`.

`useIsRecordDownloadUnderway` and `canReadAllRecords` have no callsite here: they are
surface for notebook view components supplied by third-party packages, and are
commented as such so neither is removed as dead code.

`sync.push` / `sync.pull` are not declared in `@types/pouchdb-replication`, so the code
guards with `'pull' in replication` and degrades to change-event-only catch-up.

## Checklist

---


app/src/context/slices/helpers/syncStateService.ts
```diff
@@ -122,22 +156,43 @@ class SyncStateService {
...
       docsRead: number;
       docsWritten: number;
       direction: 'push' | 'pull';
     }
   ): SyncState {
-    return this.updateSyncState(serverId, projectId, {
+    const update: Partial<SyncState> = {
       status: 'active',
-      pendingRecords: info.pending,
+      pendingRecords: info.pending ?? 0,
```
PeterBaker0, line 157: checking you want to pass coerced to zero here?

app/src/gui/components/notebook/index.tsx
```diff
@@ -363,7 +363,7 @@ export default function NotebookComponent({project}: NotebookComponentProps) {
             project={project}
             maxRows={25}
             rows={records.otherRecords}
-            loading={records.initialQuery.isLoading}
+            loading={!records.isLoaded}
```
PeterBaker0: Is there a reason for swapping things here - why not records.isLoading?

app/src/utils/customHooks.tsx
```diff
@@ -626,10 +658,77 @@ export const useRecordList = ({
...
+ * Whether a record download (the initial pull after activation, or a catch-up
+ * pull) is underway for a project, polled from the in-memory
+ * {@link syncStateService}. While true, records that exist on the server may
+ * not yet be in the local database, so a consumer must not treat a record's
+ * local absence as proof that none exists. Always false when the project
+ * never pulls (sync off, or push-only): local data is then all the device
+ * will ever have. A sync error or denial also reads as not-downloading, so
+ * an offline device keeps working from its local data.
+ */
+export const useIsRecordDownloadUnderway = ({
```
PeterBaker0, line 678: Do we use this anywhere?

PeterBaker0 (commented):
I think it looks sensible. Would appreciate a more concise summary of the core issue in the PR body. 

The custom hook introduced, is that used somewhere? Just can't see any callout to it.

Looks like an improvement to the sync status management so seems worthwhile. Would be good to have your sense of why you did this and what you saw/what this fixes.

app/src/context/slices/helpers/syncStateService.ts
jayenashar: Only for `pendingRecords`, which has no "unknown" to display. The marker below reads the raw value, so an unreported `pending` stays distinct from zero; otherwise the first batch of a bulk download declares the pull caught up. Both cases are covered in `syncStateService.test.ts`.

app/src/gui/components/notebook/index.tsx
jayenashar: There wasn't one — `useRecordList` only exposed the raw query. It now returns `isLoading`, so this reads `loading={records.isLoading}`.

Still deliberately not `initialQuery.isLoading`, which goes false the moment the *initial* fetch fails. This one is `data === undefined`.

app/src/utils/customHooks.tsx
jayenashar: Not in this repo. This and `canReadAllRecords` are consumed by third-party notebook view components, which need "no record here" distinguished from "not downloaded yet". Both doc comments now say so, and the hook has tests.

jayenashar:
Body rewritten. What prompted it: a notebook opened without network read as empty rather than loading, and one still downloading read as complete — neither distinguishable from a genuinely empty notebook.

The three comments are answered inline. `e60d9066c` renames the flag to `isLoading`, `7b2a79889` documents why the two hooks have no callsite here, `30c2b8129` tests the download hook.

app/src/utils/customHooks.tsx
PeterBaker0: Not loving these hyper verbose comments - but I get the point now

PeterBaker0: Just 'only currently used in third party plugin, not dead code' would be sufficient :D

PeterBaker0 (approved):
Looks good, thanks. 

In the future I realise that comments are largely being AI generated but I think there verbosity is getting out of hand. I think there's a tipping point where there is important information there, but developers will just ignore it as it's too long. Also subsequent agents are going to get large context bloat reading all these verbose comments. Maybe a sanity check pre-PR to ensure the very long comments are justified by very high complexity or very obscure behaviour.

PR #2217, opened 2026-07-28
Use repository name for the api docker image name

stevecassidy:
# Use repository name for the api docker image name

## Ticket

None

## Description

The docker image name was fixed as 'faims3-api' but this means we can't publish different versions from alternate repositories.  

## Proposed Changes

Make the image name be based on the repository name.  For FAIMS3 this will still be faims3-api but for a repository called FAIMS3-Alternate it would be faims3-alternate-api.  

## How to Test

Trigger the publish-conductor.yml workflow

## Checklist

### DIFF
--- a/.github/workflows/publish-conductor.yml
+++ b/.github/workflows/publish-conductor.yml
@@ -7,12 +7,10 @@ on:
     types: [published]
   workflow_dispatch:
 
-# Defines two custom environment variables for the workflow. These are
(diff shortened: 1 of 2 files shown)


PeterBaker0 (approved):
This is better, but I think an image name override as a variable would be good (to override the auto name if this doesn't suit) - but can be later/doesn't block this improvement.

PR #2219, opened 2026-07-29
build: ts7 upgrade

PeterBaker0:
Upgrades from typescript 5->7. 

The main driving factor is that [TS7 uses the new go port under the hood.](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/) We see about 2-3x reduced TS build times. This is important for rapid dev work, IDE language server integrations etc. Reduced memory usage too.

Most changes are just package bumps, but TS7 deprecates some features that we used for some bits. The most noticeable shim is the validate middleware in the API (`validate.ts`). This is AI developed, explanation from the README: 

```text
 * express-zod-safe is a dual package: ESM exposes a named `setGlobalOptions`,
 * while CJS hangs it on `module.exports` (`export =` in the .d.ts). This package
 * typechecks/emits as CJS (`module: nodenext` without `"type": "module"`), so a
 * named ESM import fails tsc (TS2305). Loading via `createRequire` keeps one CJS
 * instance for both `validate` and `setGlobalOptions` (mixing ESM import with
 * CJS require would configure a different options object). `typeof import(...)`
 * preserves the validate overloads that narrow Express Request.
```
This is an isolated shim that doesn't propagate out anywhere, so consider it worth it.

We also had issues with mocha, so AI rewrote (in a mostly bland find and replace way) mocha -> vitests. We should do a similar thing for `data-model` at some point. We see the introduction of SWC because of data model holding onto mocha and sinon mocks:

```markdown
## Testing

Unit tests use Jest. TypeScript 7 dropped the classic compiler API that `ts-jest`
relied on, so tests are transformed with `@swc/jest` (see `.swcrc` and
`jest.config.cjs`) instead.

`@swc-contrib/mut-cjs-exports` is required because some tests
`jest.spyOn(require(...), 'fn')` on ESM-style exports: SWC’s normal CJS emit
makes those exports immutable, and the plugin keeps them spy-able. This is
Jest-only noise; a later move to Vitest (like `api` / `app` / `web`) could
remove the SWC stack.
```

## Performance

We see a good bump here:

```text
──────────────────────────────────────────────────────────────
package            before      after      delta  %faster
──────────────────────────────────────────────────────────────
data-model         3562ms     1145ms    -2417ms   +67.9%
forms             11997ms     4705ms    -7292ms   +60.8%
api                7381ms     1867ms    -5514ms   +74.7%
app               10516ms     4601ms    -5915ms   +56.2%
web               11742ms     3526ms    -8216ms   +70.0%
e2e                1952ms      806ms    -1146ms   +58.7%
aws-cdk            3986ms      965ms    -3021ms   +75.8%
──────────────────────────────────────────────────────────────
AGGREGATE         51136ms    17615ms   -33521ms   +65.6%
 
baseline TS: 5.9.3  →  current TS: 7.0.2
overall: 33521ms faster (65.6% improvement)
```


stevecassidy (approved):
Looks good, everything works as expected and the code changes are minimal.

PR #2222, opened 2026-07-31
feat: improve photo taking  process, add from gallery added with multi-select ability

ranisa-gupta16:
Updated design on photo taking process and gallery option

Screenshots 

Screenshot 2026-09-09 140344

Screenshot 2026-09-09 140316

Screenshot 2026-09-09 140248

## Ticket

https://jira.csiro.au/browse/BSS-1358

## Description

The camera now opens straight away when you tap Take Photo, and choosing existing images is a separate control. You can select multiple photos from the gallery in one go instead of one at a time.

New "Choose from Gallery" button and gallery tile using Camera.pickImages() for multi-select, with a note so users know they can pick several.

image

image

## Checklist


PeterBaker0 (changes requested):
Looking good, some follow ups

1) can you set a batch add limit, like maybe 10 photos? 
2) can you confirm that the capacitor quality is still applied (limiting the resolution/size for large gallery images)
3) can you ensure that pending preview images are cleaned up if there is a subsequent error - at the moment it seems like you could get images stuck in 'Saving...' as the pending attachment is never cleared if there's an error (and is this error surfaced?) 

Thanks

ranisa-gupta16:
> Looking good, some follow ups
> 
> 1. can you set a batch add limit, like maybe 10 photos?
> 2. can you confirm that the capacitor quality is still applied (limiting the resolution/size for large gallery images)
> 3. can you ensure that pending preview images are cleaned up if there is a subsequent error - at the moment it seems like you could get images stuck in 'Saving...' as the pending attachment is never cleared if there's an error (and is this error surfaced?)
> 
> Thanks

thanks for review

1. Batch limit. Added MAX_GALLERY_BATCH = 10, passed as limit to Camera.pickImages and also enforced client‑side with photos.slice(0, MAX_GALLERY_BATCH)  If more than 10 are somehow returned, the user sees a message that only the first 10 will be added.

2 capacitor quality, confirmed and still applied to gallery picks, camera.pickimages is called with same quality , large gallery images get re-encoded the same way as captures.

3 - stuck saving error is fixed, thanks for the catch.
the pending preview was never cleared fixed it by adding removependingphoto helper..

ranisa-gupta16:
have applied fixes based on yuor comments, ready for review @PeterBaker0

ranisa-gupta16:
The PR has been updated based on discussion
The close button visibility issue on mobile DevTools is likely a dev console limitation — browser viewport emulation doesn't account for native safe area insets (safe-area-inset-top) that the Capacitor runtime provides.Will test it on device once deployed to find exact cause.

PeterBaker0 (approved):
Happy now after some testing and further refinements.

Had to add some new working around permissions and other web specifics. Testing on device and works nicely.

PR #2233, opened 2026-08-06
Feature/2232 parent field display

### LINKED ISSUE #2232: ParentFieldDisplay: show parent record values in child records
Adds a field type for displaying values from a parent record within a child record, replacing the old `displayParent` field property (which has been dead since the form refactor).

The design was worked out in the GroundSight fork (FAIMS/GroundSight-FAIMS#30 and FAIMS/GroundSight-FAIMS#31, private): rather than a flag on the parent field, the child form declares what parent context it wants via a read-only ParentFieldDisplay field, configured in the designer with the parent field to show. A parent form can have several kinds of child record each wanting different context, so the child is the right place for the decision.

The field stores nothing — the value is resolved live from the parent's current data whenever the child is viewed or edited, so parent edits are always reflected. `displayParent` is removed end to end (schema, designer checkbox, docs) with a notebook schema migration stripping it from existing notebooks.

Implemented and verified in the fork; transferring here per @stevecassidy. A follow-up will allow templated and computed fields to reference parent values using the same resolution mechanism (FAIMS/GroundSight-FAIMS#32).

Caedous:
## Ticket
Closes #2232.

## Description
Adds a ParentFieldDisplay field type: a read-only field placed in a child form that shows a value from the record's parent — the grid square's name while filling out a layer. This is the inversion from the design discussion (FAIMS/GroundSight-FAIMS#31): rather than re-implementing the `displayParent` flag on the parent, the child form declares what parent context it wants, since a parent can have several kinds of child record each wanting different context.

The field stores nothing. It resolves the value live from the parent's current data on every render, so a parent edit is reflected the next time the child is opened, with no recompute machinery. That question only genuinely arises for the follow-up (parent references in templated/computed fields), where values are stored; the resolver here is the mechanism it can reuse. With the replacement in place, `displayParent` is removed end to end, with a notebook schema 7.0 migration stripping it from existing notebooks.

## Proposed Changes
The forms library gains the field under `fieldRegistry/fields/ParentFieldDisplay`: a resolver that reads the parent relationship from the record's own revision and loads form data from the first parent whose form holds the configured field (a record may have parents of more than one form), an edit component, and a view renderer following the RelatedRecord pattern, so the value shows in both modes. The relationship is written at child creation, not first save, so a brand-new child displays the value immediately. Because nothing is stored the spec declares `returns: null`, the value schema never blocks validation — a collector cannot satisfy a requirement on a value they don't control — and the field is excluded from completion progress. Values are formatted locally rather than via registry `templateFunction`s, since a field importing the registry that imports it back is circular. Where a record has multiple parents, the value comes from the first parent whose form contains the configured field — a deliberate, non-obvious assumption, documented in the user docs and surfaced in the designer editor.

The designer offers the field in the Display tab. Its editor is a fuzzy-searchable field picker (the shared FieldSearchAutocomplete) over fields from every form that can parent this form — those holding a Child-relation Related Records field targeting it — each option showing its form and section, with a notice when no parent form exists and a warning rather than a silent clear when a chosen field has since been deleted. Per the RichText precedent it passes `showExtraConfig={false}`: Required, Conditions, and the advanced controls take input the field doesn't have. Conditional parent displays later would be a one-prop change.

The `displayParent` removal — schema property, designer checkbox, fixtures, user and developer doc references, sample notebooks — has the migrateV6-pattern migration and sits in its own commits.

## How to Test
1. `pnpm --filter @faims3/forms test` — resolver cases (value found, no parent, field on no parent's form, multiple parents of mixed forms, empty value)
2. `pnpm --filter @faims3/data-model test` — v7 migration and chain
3. In the designer: two forms, one holding a Child-relation Related Records field targeting the other. Add Parent Field Value to the child form from the Display tab; the select lists the parent form's fields under its name. On a form nothing parents, the editor shows the notice instead
4. In the app: create a parent, fill the field, create a child from it — the value shows before first save and persists in edit and view modes
5. Edit the parent's value, reopen the child — the display updates
6. A child created directly from the record list shows an em dash and saves normally; an empty parent value shows the same

**Designer — chooser and configuration:**
chooser
parent drop down

**App — new unsaved child; live update after editing the parent; no parent:**
parent test
changed value
no parent feature

## Additional Information
Developed and verified end to end in the GroundSight fork (FAIMS/GroundSight-FAIMS#31) discussion, superseded PR there closed in favour of this one); commits cherry-picked onto main with the sample-notebook fix redone against this repo's fixtures. User docs gain a Parent Field Value page under display fields; the "Display in child records" shared option (documented as a planned feature) is removed from ~30 pages, and stale `displayParent` references in the developer forms docs are removed, with a dedicated developer page deferred to a follow-up.

## Checklist


library/forms/lib/fieldRegistry/fields/ParentFieldDisplay/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+        parentFieldId,
+      }),
+    // Always re-resolve so edits to the parent are reflected.
+    staleTime: 0,
+    refetchOnMount: 'always',
+    networkMode: 'always',
+  });
+
+  let content: React.ReactNode;
+  if (!fullConfig) {
```
PeterBaker0: would be better to explicitly distinguish between the preview state (mode = preview) vs the config not being defined here - the latter is an error condition we should be aware of (i.e. it somehow got a 'full' mode but there was no config!)

library/forms/lib/fieldRegistry/fields/ParentFieldDisplay/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+      <MuiTextField
+        value=""
+        placeholder="Parent record value (shown when editing a record)"
+        fullWidth
+        disabled
+        variant="outlined"
+      />
+    );
+  } else if (query.isPending) {
+    content = <Skeleton variant="rounded" height={56} />;
```
PeterBaker0: magic number? Where did this come from - shared constant somewhere?

PeterBaker0: or put in theme or something?

library/forms/lib/rendering/fields/view/specialised/ParentFieldDisplay.tsx
```diff
@@ -0,0 +1,55 @@
...
+    staleTime: 0,
+    refetchOnMount: 'always',
+    networkMode: 'always',
+  });
+
+  if (typeof parentFieldId !== 'string') {
+    return <TextWrapper content="Field is not configured." />;
+  }
+  if (query.isPending) {
+    return <Skeleton variant="text" width="60%" />;
```
PeterBaker0: 60% magic number

web/src/designer/components/Fields/ParentFieldDisplayEditor.tsx
```diff
@@ -0,0 +1,157 @@
...
+  viewId: string;
+  viewsetId: string;
+};
+
+// Fields with no meaningful single value to display.
+const EXCLUDED_COMPONENT_NAMES = [
+  'RelatedRecordSelector',
+  'ParentFieldDisplay',
+  'RichText',
+];
```
PeterBaker0: I think we shouldn't localise these kind of concerns into distributed spots - add optional field to the registry spec for a field 'excludeFromParentView' or something - default is false (i.e. it should be included), then add this flag for offending fields.

PeterBaker0: Unless this is difficult to do for some reason

web/src/designer/components/Fields/ParentFieldDisplayEditor.tsx
PeterBaker0, line 1: Can you expand the field search pattern to instead of building another select/autocomplete dropdown - lots of work to make the fuzzy search selector nice!

PeterBaker0 (dismissed):
Looking good, but some changes requested. Approach makes sense. 

Worth highlighting more concisely in the PR the main 'assumption' is that for children with multiple parents you take the 'first' value - this is a non trivial, non-obvious assumption. Also ensure it's clearly documented for users in relevant places.

Caedous:
> Looking good, but some changes requested. Approach makes sense.
> 
> Worth highlighting more concisely in the PR the main 'assumption' is that for children with multiple parents you take the 'first' value - this is a non trivial, non-obvious assumption. Also ensure it's clearly documented for users in relevant places.

Agreed it deserved prominence, now stated in three places: highlighted in the PR description's Proposed Changes, a dedicated bullet in the user doc's Behaviour section, and as a caption directly under the field picker in the designer so notebook designers meet the rule where they configure the field (screenshot below).

library/forms/lib/fieldRegistry/fields/ParentFieldDisplay/index.tsx
Caedous: Split as suggested, preview mode gets the placeholder via an explicit `config.mode === 'preview'` branch, while full mode with a missing data engine is treated as a wiring error and is logged via the forms logger and rendered as an error state rather than masquerading as preview.

library/forms/lib/fieldRegistry/fields/ParentFieldDisplay/index.tsx
Caedous: Checked for a shared constant, none exists (the web components use the shadcn/Tailwind skeleton, and RelatedRecord inlines the same values), so rather than mint a local constant that just names the number, I've switched both skeletons to MUI's size inference: the Skeleton wraps a hidden instance of the element it stands in for and takes its exact dimensions. No number at all, and it tracks MUI's sizing automatically.

library/forms/lib/rendering/fields/view/specialised/ParentFieldDisplay.tsx
Caedous: Same fix as the height, size inference from a hidden placeholder child, no number.

web/src/designer/components/Fields/ParentFieldDisplayEditor.tsx
Caedous: Moved to the registry as suggested: optional `excludeFromParentDisplay?: boolean` on `FieldInfo` (default false), set on the RelatedRecord, RichText, and ParentFieldDisplay specs, with the editor consulting `getFieldInfo` instead of a local list. Wasn't difficult, `getFieldInfo` was already exported from the package.

web/src/designer/components/Fields/ParentFieldDisplayEditor.tsx
Caedous: Swapped to `FieldSearchAutocomplete`. I keep forgetting this a thing and don't have to build it from scratch anymore!

stevecassidy (commented):
As discussed, let's persist the value of the parent field in the form so that it appears in exported data.   The value is copied over only when the child record is opened so may be stale if the parent is edited later.   We can perhaps resolve that when we do the more general case of using values from other records in computed fields.

Caedous:
> As discussed, let's persist the value of the parent field in the form so that it appears in exported data. The value is copied over only when the child record is opened so may be stale if the parent is edited later. We can perhaps resolve that when we do the more general case of using values from other records in computed fields.

Added following the review discussion: the resolved parent value is now written into form state on resolution, so it persists on save and appears in exports (typed as String, matching the designer's declaration). Display still resolves live; an unchanged value doesn't dirty the form, so reopening a child without edits creates no new revision. Verified locally end to end, the value lands in the record's AVP data and updates on the child's next save after a parent edit.

stevecassidy (approved):
Happy with this now but we've introduced an issue that we'll have to deal with (stale values).   I've made a new ticket for a short term fix but long term I think we need to build deeper support for dependencies between records.

PR #2234, opened 2026-08-07
feat: XLSForm-to-Fieldmark converter 

anesuc21:
# feat: XLSForm-to-Fieldmark converter

## Ticket
#NNNN 

## Description
Adds the ability to create and replace Fieldmark templates from uploaded
XLSForm (.xlsx) files, converting them into Fieldmark's NotebookDefinition
JSON format. The conversion happens server-side where the browser uploads the raw
file (base64-encoded) to a dedicated endpoint, which parses and converts it,
then reuses the existing template create/update logic and permission model.

**Converter**
- Added `convertXlsformToNotebookDefinition()` in
  `library/data-model/src/uiSpecification/xlsformConverter.ts`, covering
  v1 field types: `text`, `integer`, `decimal`, `select_one`,
  `select_multiple`, `date`, `dateTime`, `geopoint`, `image`, `file`.
  Unsupported types are collected into a `skipped` list rather than
  failing the whole conversion.
- Field names are derived deterministically from each XLSForm question's
  own `name` column, so re-converting the same file produces the same
  field names every time so that the replace-template flow doesn't
  silently orphan existing project data on re-import.
- 27 unit tests in `library/data-model/tests/xlsformConverter.test.ts`,
  covering all field types plus edge cases (missing labels/hints,
  `required` string-to-boolean handling, unsupported-type skipping, HRID
  selection, and determinism).

**Backend**
- Added `read-excel-file` as a dependency for parsing uploaded `.xlsx`
  files ([npm](https://www.npmjs.com/package/read-excel-file)). 
- Added `parseXlsformBuffer()` in `api/src/utils/xlsformParsing.ts`,
  reading an uploaded file buffer into survey/choices/settings row
  objects, with error messages that cite the specific sheet, row, and
  column when something is missing or invalid.
- Added `POST /api/templates/from-xlsform` which creates a new template from
  an uploaded XLSForm file, reusing the existing `createTemplate()`
  function and permission model (`CREATE_TEMPLATE` /
  `CREATE_TEMPLATE_IN_TEAM` / `CREATE_PUBLIC_TEMPLATE`).
- Added `PUT /api/templates/:id/uiSpecification/from-xlsform` which replaces
  an existing template's design from an uploaded XLSForm file, mirroring
  the existing JSON-based `PUT /:id/uiSpecification` route.
- Both endpoints return the created/updated template alongside the
  `skipped` list, so the caller can report which question types didn't
  convert.
- 12 integration tests in `api/test/xlsformTemplate.test.ts` (in-memory
  PouchDB, no external services required), covering successful
  conversion, global vs. team-owned templates, input validation,
  structured error messages, permission denial, the skipped behaviour,
  and the bare authentication check.

**Frontend**
- Widened `designFileSchema()` (`web/src/lib/input-limits.ts`) to accept
  `.xlsx` alongside `.json`.
- Added `fileToBase64()` helper for encoding uploaded files.
- Branched `create-template-form.tsx` and `update-template-form.tsx` to
  route `.xlsx` uploads to the new server-side endpoints; the existing
  JSON upload path is unchanged.
- Added `updateTemplateUiSpecificationFromXlsformRequest()` to
  `template-hooks.ts`, mirroring the existing JSON update request helper.
- Updated label/description text in `create-template-dialog.tsx`,
  `edit-template.tsx`, and `templates/actions.tsx` to mention XLSForm
  support.

## How to Test

1. `pnpm install` to pick up the new dependencies.
2. Start the local stack (`./localdev.sh --all` or your usual dev
   workflow).
3. **Create flow:** Templates → Create Template → upload a `.xlsx`
   XLSForm file (survey/choices/settings sheets). A template should be
   created successfully.
4. **Replace flow:** open an existing template → Actions → Replace
   Template JSON File or XLSForm → upload a `.xlsx` file. The template's
   design should update in place.
5. Automated tests:
   - `cd library/data-model && pnpm test -- xlsformConverter.test.ts`
   - `cd api && pnpm test -- xlsformTemplate.test.ts`

**Not yet done:** manual/live verification of the replace-template
(`PUT .../from-xlsform`)

## Checklist


api/src/api/templates.ts
```diff
@@ -60,17 +60,119 @@ import {isAllowedToMiddleware, requireAuthenticationAPI} from '../middleware';
 import {saveExpressUser} from '../couchdb/users';
 import patch from '../utils/patchExpressAsync';
 
+import {parseXlsformBuffer} from '../utils/xlsformParsing';
+import {
+  convertXlsformToNotebookDefinition,
+  CURRENT_NOTEBOOK_UI_SCHEMA_VERSION,
+} from '@faims3/data-model';
+
+const PostCreateTemplateFromXlsformInputSchema = z.object({
```
stevecassidy: These should go in library/data-model with all of the other request types, then we can access them from the clients as well.

api/src/api/templates.ts
```diff
@@ -60,17 +60,119 @@ import {isAllowedToMiddleware, requireAuthenticationAPI} from '../middleware';
...
+type PostCreateTemplateFromXlsformInput = z.infer<
+  typeof PostCreateTemplateFromXlsformInputSchema
+>;
 // This must occur before express api is used
 patch();
 
 export const api: express.Router = express.Router();
 
+/**
+ * POST create new template from an uploaded XLSForm (.xlsx) file.
```
stevecassidy: I think a better approach would be to have a separate endpoint to convert the XLSform to a uiSpec, we could then use the regular endpoint to create either a template or a notebook from the uiSpec - noting also that this only gives us the uiSpec, not the other elements of a template like the metadata etc.  So that endpoint would not go here, maybe in utilities.ts with an endpoint of /convert-xlsform.  It should just return the uiSpec JSON.

api/src/utils/xlsformParsing.ts
```diff
@@ -0,0 +1,71 @@
+import readXlsxFile, {parseSheetData} from 'read-excel-file/node';
+import * as Exceptions from '../exceptions';
+import type {XlsformSheets} from '@faims3/data-model';
+
+const surveySchema = {
```
stevecassidy, line 5: Some documentation here would be useful, what are these schema doing?

api/src/utils/xlsformParsing.ts
```diff
@@ -0,0 +1,71 @@
...
+  name: {column: 'name', type: String, required: true},
+  label: {column: 'label', type: String, required: true},
+};
+
+const settingsSchema = {
+  form_title: {column: 'form_title', type: String, required: false},
+  form_id: {column: 'form_id', type: String, required: false},
+};
+
+export async function parseXlsformBuffer(
```
stevecassidy, line 27: Add documentation comments before your major functions to help future developers understand what's going on.

api/src/api/templates.ts
```diff
@@ -60,17 +60,119 @@ import {isAllowedToMiddleware, requireAuthenticationAPI} from '../middleware';
...
+      })
+    ) {
+      throw new Exceptions.UnauthorizedException(
+        'You are not authorized to create a public template.'
+      );
+    }
+
+    // Decode and parse the uploaded spreadsheet
+    const fileBuffer = Buffer.from(body.fileBase64, 'base64');
+    const sheets = await parseXlsformBuffer(fileBuffer);
```
stevecassidy: Why do you do the parsing here rather than in the convert function? They would seem to belong together.

stevecassidy: Ok, I can see that this keeps the XLS parser dependency in API so it doesn't end up in the front end clients - maybe note that in the xlsformParsing module.

library/data-model/src/uiSpecification/xlsformConverter.ts
```diff
@@ -0,0 +1,305 @@
...
+ * column, so re-converting the same file produces the same field names
+ * every time -- required for the Replace Template JSON flow to not
+ * silently orphan existing project data on re-import.
+ */
+
+// ---------------------------------------------------------------------------
+// Input shapes -- what we expect after read-excel-file + parseSheetData
+// ---------------------------------------------------------------------------
+
+export interface SurveyRow {
```
stevecassidy, line 22: These are repeated from the conversion module - put shared types here in data-model and import them into the converter. We want to avoid repetition like this.

library/data-model/src/uiSpecification/xlsformConverter.ts
```diff
@@ -0,0 +1,305 @@
...
+function convertChoices(choices: ChoiceRow[], listName: string) {
+  return choices
+    .filter(c => c.listName === listName)
+    .map(c => ({label: c.label, value: c.name}));
+}
+
+function convertField(row: SurveyRow, choices: ChoiceRow[]): FieldDefinition {
+  const baseType = row.type.split(' ')[0];
+
+  switch (baseType) {
```
stevecassidy, line 126: Ideally we would type-check these return values as valid field definitions, unfortunately we can't access the field schema from the forms module here because that has a react dependecy.   Maybe just put a comment here on the need to keep these results in sync with form schema.  In future we might find a way to do this properly.

web/xlsformConverter.ts
```diff
@@ -0,0 +1,305 @@
+/**
```
stevecassidy, line 1: I think this should be removed.

stevecassidy (changes requested):
Generally good, some notes on the location of a few things and suggesting a change in the way you wire this into the API. Nice work overall, thanks!

PR #2238, opened 2026-08-09
Recursive record status report

### LINKED ISSUE #2237: Recursive completion status for a record and its child records
We show percent complete for a single record (required-field completion in @faims3/forms) but there is no way to get the completion of a record including its child records, where a parent form's children each have their own forms and children.

Add a status report for a record: HRID, percent complete, summary field values, and the same recursively for children (faims-core::Child relations only). Each record counts as one unit so a large child form can't dwarf the parent: percent = (own + sum of child percents) / (1 + child count), with a required child field that has no records yet counting as one expected child.

This needs completion() moved into @faims3/data-model so the api can use it too, plus a route like GET /api/notebooks/:id/records/:recordId/status for dashboards.

jayenashar:
## Ticket

#2237

## Description

Adds a recursive status report for a record and its child records: per node the HRID, required-field completion, summary field values, and the same for children. Closes #2237.

## Proposed Changes

- `completion()` moves from @faims3/forms to @faims3/data-model so the status report can compute it; the forms package keeps the same public API and injects its field-registry overrides. One behavior fix while moving: a required field listed in two visible sections now counts once instead of once per section.
- `computeRecordStatusReport` walks faims-core::Child relations from a record's RelatedRecordSelector values and rolls completion up as (own progress + sum of child progress) / (1 + live children). Each record is one unit so a large child form cannot dwarf its parent; an empty required child field counts once, through own progress.
- A required child field is complete in own progress only while it has a live child, so a link to a deleted record scores no better than an empty field. The edit form's progress bar scores synchronously without this liveness check, an accepted divergence after a child delete.
- Deleted and unreadable children drop out of both sides of the ratio, so percentages are per-viewer on mixed-author trees.
- Child graphs are assumed acyclic (the app only creates parent-to-child links; guarding creation against cycles is #2244). A cycle in corrupt data is cut where it closes so the walk still terminates.
- A record whose form was removed from the notebook fails fast with `UnknownFormTypeError` rather than reporting an unmeasurable 100%.
- No HTTP route (in-app use only for now): the report is exposed from @faims3/data-model. The `recordFilter` hook was dropped as unused; a future API caller should add filtering shaped to its real permission model, and map a filtered-out root and not-found to one response so record ids cannot be probed.

Video: https://github.com/user-attachments/assets/f3034d62-c3cb-441c-9fa2-b9660d4abf75

## Refactors along the way

- Visibility evaluation (`currentlyVisibleMap`, `getFieldsMatchingCondition`, `getViewsMatchingCondition`) moves from @faims3/forms to @faims3/data-model's uiSpecification module: the walk evaluates per-record visibility with no form mounted, and data-model cannot import forms. Forms re-exports the moved types (`FieldVisibilityMap`, `CompletionResult`) so existing import sites keep working; pointing them at data-model directly is a mechanical follow-up.
- While moving, `getFieldsMatchingCondition`/`getViewsMatchingCondition` drop two dead parameters (`fieldNames`/`views` and `touched`, which every caller passed empty), and `conditional_sources` goes with them since nothing reads it any more. The exported signatures change in place; all in-repo callers are updated.
- `formDataExtractor` in forms is now an alias of data-model's `formDataToValues` (extracted so `completion()` can run in data-model); collapsing the two names is a mechanical follow-up.
- The record list's summary columns now compute through the shared `getSummaryValues` in the hydration select, so the grid and the Status tab read summary values from one implementation. Column headers switch from `prettifyFieldName` (text derived from the field id, now deleted) to the field's designer label via `getFieldLabel`.

## Open Design Questions

- Resolved: the roll-up no longer charges an empty required Child field twice; it counts once, through own progress, so the bar matches the "n/m required fields" caption. Consequence to be aware of: an entirely missing required subtree weighs one required field (a Site at 1 of 2 own fields with no Features reports 50%), with the gap still flagged in warning colour on the tab.
- `related_type` is validated only by the designer, so a raw-JSON notebook upload can carry an empty or dangling id, which renders as a required child row with a blank form label. A cheap guard skips Child fields whose `related_type` is not a viewset key. Worth doing here, or leave to notebook validation?

## How to Test

`pnpm --filter @faims3/data-model test` covers the roll-up math, exclusions, conflict-head parity and cycle cutting (tests/statusReport.test.ts). For a live check, open a record with children in the app: the record view's new Status tab renders this report as the completion tree.

## Additional Information

The walk descends depth-first with no cache, fanning out concurrently over each record's distinct children, so a record linked from several parents is fetched once per path; app-created data is strictly tree-shaped, so in practice that is once per record. Batched fetching, a concurrency cap, and a total-node budget for cross-linked data are possible follow-ups that need no API change.

Conflict policy: the status walk hydrates with the page default (pickFirst), matching the record view, edit form and API reads, and a test pins the walk to whatever head `getExistingFormData` resolves. The RelatedRecord widgets and the export stripper still pass pickLast, so a conflicted child can render differently inside a RelatedRecord widget than on its own record page; aligning those call sites is a possible follow-up.

`getDependantFields` has no production caller now that `conditional_sources` is gone; it stays as tested public API rather than shrinking the package surface in this PR.

Parent-form inference, the status report and the designer's ParentFieldDisplayEditor now share one validated child-relation predicate (`getChildRelationParams`) and one viewset field enumeration (`fieldIdsForViewset`). The form editor's `preventFormDelete` check still uses a looser name-only scan; switching it to the shared predicate is a possible follow-up.

Known display gap, inherited from the record table: `getDisplayDataFromRecordMetadata` joins array values without stringifying objects, so an object-array summary field (e.g. photos or related records) renders as `[object Object]` on the Status tab, as it already does in the record list. A shared fix in formUtilities is a possible follow-up.

## Checklist


library/data-model/src/databaseEngine/statusReport.ts
```diff
@@ -0,0 +1,404 @@
...
+// Backstop against corrupt/pathological deep data (real trees are ~4 levels);
+// it also bounds the response, which repeats a shared subtree once per path
+export const STATUS_REPORT_MAX_DEPTH = 10;
+
+// Only the id and project tag matter here, so legacy vocab-pair drift in a
+// stored link cannot invalidate a live child
+const storedLinkEntrySchema = z.object({
+  record_id: z.string().min(1),
+  project_id: z.string().optional(),
+});
```
PeterBaker0, line 37: This is duplication of data model elements I believe - can you double check these relation fields aren't already defined elsewhere? I think there is even logic existing in the data engine which does the array vs singleton unwrap logic you have below.

library/data-model/src/databaseEngine/statusReport.ts
```diff
@@ -0,0 +1,404 @@
...
+      : raw === null || raw === undefined
+        ? []
+        : [raw];
+    const childIds: string[] = [];
+    for (const rawEntry of rawEntries) {
+      const entry = storedLinkEntrySchema.safeParse(rawEntry);
+      if (!entry.success) {
+        continue;
+      }
+      const {record_id: childId, project_id: linkProjectId} = entry.data;
```
PeterBaker0: Pretty sure this is duplicated from elsewhere in data engine - try not to redefine these models as they could shift through migrations.

library/data-model/src/databaseEngine/statusReport.ts
```diff
@@ -0,0 +1,404 @@
...
+      // Legacy app versions stored the system-wide id (listing||notebook);
+      // only the notebook part is comparable to the API's bare project id.
+      // An empty-string tag means untagged, like an absent one
+      const linkNotebookId = linkProjectId
+        ? linkProjectId.split(ENCODING_SEPARATOR).pop()
+        : undefined;
```
PeterBaker0: At a certain point we should drop these legacy concerns. @stevecassidy  what do you think?

library/data-model/tests/statusReportUiSpec.json
PeterBaker0, line 1: Just checking @jayenashar  that this sample notebook is not a leakage of any IP in your proprietary work? Not sure just wortha. check

PeterBaker0 (changes requested):
Please discuss with Steve re: usage in FAIMS3 (propose as status panel on view record page + removal of API route). Also some concerns around duplication of the record relation models.

jayenashar:
steve's comments during standup:
- no use for api right now.  just used within app for now.
- may make a tab in the record view to the status.  show the full tree.
- create an issue to prevent parent-child cycles.

library/data-model/src/databaseEngine/statusReport.ts
jayenashar: deduped in 7eb8d8d0 / 72e45be0

library/data-model/src/databaseEngine/statusReport.ts
jayenashar: deduped in 7eb8d8d0 / 72e45be0

library/data-model/tests/statusReportUiSpec.json
jayenashar: checked.  generic Site/Feature/Sample/Photo forms.

library/data-model/src/databaseEngine/statusReport.ts
jayenashar: the existing unwrap (`normalizeRelationshipInstances`) is for revision.relationship entries with field_id, and whole-value parsing via `relatedRecordFieldAvpValueSchema` would drop a field's healthy links when one entry is corrupt

library/data-model/src/databaseEngine/statusReport.ts
jayenashar: dropped in c935e27be.  hope that's ok.

library/data-model/src/databaseEngine/completion.ts
```diff
@@ -0,0 +1,117 @@
...
+/** Pulls raw values out of form data, e.g. for visibility evaluation. */
+export function formDataToValues(
+  data: FormUpdateData | undefined
+): Record<string, unknown> {
+  const out: Record<string, unknown> = {};
+  for (const [k, v] of Object.entries(data ?? {})) {
+    out[k] = v.data;
+  }
+  return out;
+}
```
PeterBaker0, line 117: I'm 90% sure this is part of the data model somewhere - definitely used util funciton like this elsewhere.

PeterBaker0: nitpick though

library/data-model/src/databaseEngine/types.ts
```diff
@@ -69,13 +69,6 @@ export const relationshipInstanceSchema = z.object({
 });
 export type RelationshipInstance = z.infer<typeof relationshipInstanceSchema>;
 
-/** Relation kind for RelatedRecordSelector `component-parameters.relation_type`. */
-export const relatedTypeSchema = z.enum([
-  'faims-core::Child',
-  'faims-core::Linked',
-]);
-export type RelatedType = z.infer<typeof relatedTypeSchema>;
```
PeterBaker0, line 77: Where did this go?

library/data-model/src/databaseEngine/types.ts
```diff
@@ -101,18 +94,12 @@ export type RelatedRecordFieldAvpValue = z.infer<
...
  */
-export const relatedRecordSelectorComponentParamsSchema = z
-  .object({
-    related_type: z.string(),
-    relation_type: relatedTypeSchema,
-    multiple: z.boolean().optional().default(false),
-    allowLinkToExisting: z.boolean().optional().default(false),
-    hideCreateAnotherButton: z.boolean().optional().default(false),
-  })
-  .passthrough();
```
PeterBaker0, line 115: And this one?

library/data-model/src/uiSpecification/types.ts
```diff
@@ -170,6 +170,33 @@ export const CompiledUiSpecFieldsSchema = z.record(
 );
 export type CompiledUiSpecFields = z.infer<typeof CompiledUiSpecFieldsSchema>;
 
+/** Relation kind for RelatedRecordSelector `component-parameters.relation_type`. */
+export const relatedTypeSchema = z.enum([
+  'faims-core::Child',
+  'faims-core::Linked',
+]);
+export type RelatedType = z.infer<typeof relatedTypeSchema>;
```
PeterBaker0, line 178: ah here... okay I see.

PeterBaker0 (approved):
Looks good, I like the approach and think the status tab will be helpful - just needs some more thinking about how best to present the information visually, and what subset of information might be useful to users. 

Appreciate accommodating the open source in the meantime by adding the feature flag.

library/data-model/src/databaseEngine/completion.ts
jayenashar: the only prior copy was the inline loop in forms' `formDataExtractor`, which this PR replaced with a call to this function. the closest existing data-model util is `dataMap` in engine.ts, but it's typed to `HydratedDataField`, not `FormUpdateData`.

PR #2240, opened 2026-08-10
build: adds new EBS recovery documentation and improves docs

PeterBaker0:
Mostly just a new guide, but also exercises it and fixes a couple of issues. 

This included some missing lambda memory size limit bumps (causing very slow deploys), unnecessary `--` pass through flag in pnpm scripts, and other small touch ups.


stevecassidy (approved):
Seems harmless!

Issue #2245, opened 2026-08-11
Refresh parent-dependent values before export


Caedous:
Follow-up to #2232 and companion to #2235. ParentFieldDisplay persists the resolved parent value in the child record, and #2235 will do the same for templated/computed values referencing parent fields. These stored values are accurate as of the child's last save — if the parent changes and the child isn't subsequently viewed, an export contains the stale value.

Originates from FAIMS/GroundSight-FAIMS#109 (private). Implement a refresh pass in the API that runs before export: examine the uiSpec for forms with parent-dependent fields; where found, load records of those forms, re-derive the dependent values (same evaluation the form runs on view), and where a value differs, write a new revision credited to the admin user. Records whose derived values are unchanged are untouched, so the pass creates no revision churn.

Note the accepted trade-off: server-written revisions can land beside unsynced device edits of the same record; this resolves through the existing head/conflict machinery.

Caedous:
Addition from discussion: alongside the pre-export refresh, expose the same operation as an API endpoint with a manual "recalculate values" button (likely in the web manager near the export controls). The endpoint returns a summary (records examined/updated) for the UI to display.

Caedous:
This has become a simple movement of the derivation into the data-model. Refresh of data to be done at a later date.

PR #2246, opened 2026-08-11
feat: Reference parent record values in templated and computed fields

### LINKED ISSUE #2235: Allow templated and computed fields to reference parent record values
Follow-up to #2232 (ParentFieldDisplay). Allow TemplatedStringField and computed fields (ComputedNumber/ComputedText) in a child form to reference values from the parent record, e.g. generate a layer name from the grid square's name.

Originates from FAIMS/GroundSight-FAIMS#32 (private); transferred here as the work belongs in the open source codebase. Uses the parent resolution mechanism introduced in #2232.

Design questions to settle before implementation: reference syntax in templates and expressions, how the expression type-checker validates a parent reference when a child form has multiple possible parent forms, and recompute semantics given these values are stored (unlike ParentFieldDisplay, which resolves live).

Caedous:
## Ticket
Closes #2235.

## Description
Lets TemplatedStringField and computed fields (ComputedNumber/ComputedText) in a child form reference values from the record's parent — the GroundSight use case being a layer's name generated from its grid square's name. Templates use `{{parent.Field-ID}}` alongside the existing tokens; expressions use `{parent.Field-ID}` alongside existing braced references. Design agreed in FAIMS/GroundSight-FAIMS#32 (private) and the follow-up discussion on #2232.

Two simplifications emerged during implementation and are worth calling out. First, no new parsing was needed on either side: Mustache resolves dotted names natively, and the expression compiler already treats braced references as opaque strings typed via a caller-supplied map — parent references are just additional entries. Second, the ambiguity concern from the design discussion ("what type is `parent.X` when a form has multiple possible parent forms?") largely dissolves: field IDs are globally unique in the uiSpec, so an ID cannot carry two types. The ambiguity handling remains as a defensive branch, and a reference to a field on no parent form gets a targeted compile error rather than a generic one.

Unlike same-form references, parent references may point at the parent's own templated and computed fields: their stored values are read, which is exactly what #2232's persist-on-save guarantees exist, and it means a child's HRID can extend its parent's HRID.

## Proposed Changes
data-model gains `parentForms.ts`: `getParentFormsForForm` (which forms hold a Child-relation Related Records field targeting a given form — the scan the ParentFieldDisplay editor performs, moved down so all layers share it), `buildParentFieldTypes` (types the available `parent.` references), and `compileComputedExpressionForForm` (per-form compile wrapper with targeted errors for bad parent references, including a reserved-prefix guard). The notebook-load compile pass uses the wrapper.

The forms library gains `resolveParentValues` beside the #2232 resolver — same resolution rule (first parent whose form can parent this form), returning the parent's raw values. `RecordContext` carries them; templates receive a formatted `parent` object (parent fields go through the same templateFunction formatting as local ones) and the computed scope resolver types `parent.` references from the parent form's declared types. `EditableFormManager` resolves once at mount and fires one recompute when values arrive, explicitly scheduling a save when the recompute changed anything — without that, values displayed but never persisted. The save path passes the same context, so values re-derive on open and on save: stored values are accurate as of the child's last save, with no write-time propagation from parent edits (unworkable offline; the export-time staleness question is #2245).

The designer's expression editors validate with the per-form wrapper (so parent references no longer flag as unknown, and bad ones get specific messages), gain an Insert parent field picker beside the existing field search, and the template Visual Builder lists parent fields as "Parent:" variables.

A record with no parent contributes empty values: templates render the reference as empty text, computed results stay blank, and the record saves normally — matching ParentFieldDisplay's behaviour.

## How to Test
1. `pnpm --filter @faims3/data-model test` — parent form scan, reference typing, per-form compile errors (12 new tests)
2. `pnpm --filter @faims3/forms test` — resolver walk, template injection, computed scope resolution including parent derived fields (12 new tests)
3. In the designer: two forms linked by a Child-relation Related Records field. On the child, a Templated String with `{{parent.<field>}}` and a ComputedNumber using the Insert parent field picker — both validate clean; `{parent.nonexistent}` reports the targeted error; a form with no parent shows no picker and its own targeted error
4. In the app: open a child of a parent with values — template and computed fill on open and persist on save (visible in export/record data); reopening without edits creates no new revision; editing the parent then reopening the child re-derives on open with one new revision; a child created with no parent renders empty parts and saves normally

## Additional Information
User docs: templated string and both computed field pages gain a "Referencing Parent Record Values" section and lose the "all referenced fields must be in the same form" restriction; the Parent Field Value page is corrected for #2232's persist-on-save behaviour (its text predated that review change) and cross-references the new sections. Conditions remain same-form (that would be FAIMS/GroundSight-FAIMS#70). Export-time refresh of stale stored values is #2245, which builds on the pure recompute functions and the resolver this PR leaves engine-agnostic.

## Checklist


library/data-model/src/uiSpecification/parentForms.ts
```diff
@@ -0,0 +1,206 @@
...
+}: {
+  source: string;
+  uiSpecification: UiSpecModel;
+  formId: string;
+  requiredType?: ExprType;
+}): CompiledExpression => {
+  const fieldTypes = new Map<string, ExprType>();
+  for (const [id, f] of Object.entries(uiSpecification.fields)) {
+    if (id.startsWith(PARENT_REFERENCE_PREFIX)) {
+      // Reserved prefix - a real field with this ID would be shadowed.
```
stevecassidy, line 163: Hmm, maybe we use _PARENT as the prefix then to make this less likely.  Similar to the way we have _CREATED_TIME in templated strings.

stevecassidy (approved):
Very happy with this. Nicely done.  One suggestion about the prefix we use for parents but otherwise ready to merge.

library/data-model/src/uiSpecification/parentForms.ts
Caedous: Done, changed to `_PARENT` throughout, consistent with the _CREATED_TIME convention.

PR #2247, opened 2026-08-11
feat: improved survey cleanup safety with tombstone DB

PeterBaker0:
A pretty simple and nice change which adds an 'api only' couch DB called 'tombstone'. It just leaves a breadcrumb for surveys which were explicitly deleted. It is exposed through a single API endpoint for getting a tombstone by ID. 

This means the app, when a survey is missing upon initialization, can check explicitly if the survey was deleted. This protects against device deletion of local data in the case where a misconfigured/corrupted server responds with an empty list of surveys. 

Adds some e2e tests to check this logic working too.


stevecassidy (approved):
Nice solution, Thanks.

PR #2248, opened 2026-08-12
feat: ttl cleanup script for expiring DB records + CDK wiring

PeterBaker0:
Implements an API ts script which uses the existing types/infra of the TS API context to write a safe DB TTL cleanup job. It targets expiring ephemeral DB artifacts, especially refresh tokens. These currently hang around forever and in a busy system would eventually exhaust the DB.

Alternatively, we could have done a cron bash job, but this would mean either we have a bash script which would be tricky to maintain given DB types migrate through time, or we need to have the full TS runtime/repo context in the VM which is best avoided. It also leaks a fair bit of FAIMS specific stuff in the DB layer which is not ideal as it makes transferring to other couch hosts harder. 

In the AWS world, running this as a scheduled job on the API node is not ideal as they intentionally should be able to run in a cluster, so you'd need some master. The ideal approach as implemented is to run an ECS 'task' which is ephemeral (exits as a typical task with 0/1 exit code) scheduled on a cron job. It can reuse the whole TaskDfn from the existing API task, and just override the run command. Also implemented alerting against the existing SNS topic. 

Adds CDK config to enable and configure the script including scheduling and options.

### DIFF
--- a/api/src/couchdb/ttlCleanup.ts
+++ b/api/src/couchdb/ttlCleanup.ts
@@ -255,7 +255,10 @@ async function* paginateAuthView<T extends {_id?: string}>(
     const result = await authDB.query<T>(viewName, opts);
     let rows = result.rows.filter(r => !!r.doc).map(r => r.doc as T);
 
-    if (skipFirst && rows.length > 0) {
+    // Only drop the startkey row when it is still present. Deletes between
+    // pages remove the previous page's last id from the view, so CouchDB's
+    // next page already starts at the next live key — blind skip would drop it.
+    if (skipFirst && rows.length > 0 && rows[0]._id === startkey) {
       rows = rows.slice(1);
     }
     if (rows.length === 0) {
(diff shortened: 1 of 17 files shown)


api/src/couchdb/ttlCleanup.ts
```diff
@@ -0,0 +1,563 @@
...
+  const auditAnchor = revoked
+    ? (doc.updatedTimestampMs ?? doc.createdTimestampMs)
+    : (doc.expiryTimestampMs ??
+      doc.updatedTimestampMs ??
+      doc.createdTimestampMs);
+
+  return auditAnchor < nowMs - auditRetentionMs;
+};
+
+async function* paginateAuthView<T extends {_id?: string}>(
```
stevecassidy: Documentation of these would be useful

stevecassidy (approved):
Looks ok, just one request for some docs

PR #2251, opened 2026-08-13
feat: strengthening invite code generation

PeterBaker0:
Deprecates the 'short code' concept from FAIMS. This is for security reasons. More information available upon request. 

Changes the UI/API to 

- generate longer codes
- strengthen audit logs
- improve UX for longer codes including modal dialog for copying the code if necessary
- stack layout on mobile devices so that code input is still possible

Also strengthens input validation for input codes across teams/surveys/users.


stevecassidy (approved):
Looks good, nice tidy up at the same time. Thanks.

PR #2252, opened 2026-08-14
refactor: move derived-value evaluation into data-model and modularise it (#2245)

Caedous:
## Ticket
Relates to #2245 (refresh work parked — see below).

## Description
Moves the pure derived-value evaluation code from `@faims3/forms` into `@faims3/data-model` so it can run server-side as well as in the app. This is groundwork: the app and any server-side consumer must evaluate templated strings, computed fields and parent-field lookups identically, and one shared implementation is how that's guaranteed. The refresh-before-export feature originally built on this has been parked in favour of a resolve-on-read approach (per discussion with @stevecassidy and @PeterBaker0); that work is preserved on `park/2245-refresh-pass`.

## Proposed Changes
`resolveParentValues`, `resolveParentField`, the templated-string and computed-field recompute, record context, and `addressTypes` move to `data-model/derivedFields`. Forms imports point directly at data-model, no re-export shims and the tanstack bindings (`onChangeTemplatedFields`/`onChangeComputedFields`) plus the registry-bound `recomputeDerivedFields` wrapper stay in forms. Template functions (per-field-type value-to-text converters used by templates) move to a builtin map in data-model, with the lookup injected so the field registry entry references the same function; the file explains why the address function lives there rather than on the field. Ticket references removed from code comments.

## How to Test
1. `pnpm --filter @faims3/forms test` — full suite passes against the moved code imported directly from data-model, evidencing behaviour preservation
2. `pnpm --filter @faims3/data-model test` — passes
3. App: templated strings, computed fields and Parent Field Value fields behave as before on open and save

## Additional Information
Testing the original refresh work surfaced a pre-existing bug in the export AVP loader, fixed in the second commit: `loadAttributeValuePair` treated a present-but-empty `faims_attachments` list as attachment-bearing under `includeAttachments: false` (the CSV/streaming export mode) and overwrote the AVP's data with the empty list, so every field exported as `[]`. Dormant in production because client-written AVPs omit the key entirely; the engine stamps `faims_attachments: []` when a payload supplies an empty array, so the first server-side writer would have tripped it. The guard now treats present-but-empty as no attachments. Regression tests cover empty, absent, real-attachment and mixed-batch cases. No migration needed: the fix is read-side, so any already-stamped records export correctly without rewriting.

## Checklist


web/src/components/forms/recalculate-values-form.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+type RunState =
+  | {phase: 'idle'}
+  | {phase: 'running'}
+  | {phase: 'done'; summary: RefreshDerivedValuesSummary}
+  | {phase: 'error'};
+
+/**
+ * RecalculateValuesForm triggers the server-side refresh of values derived
+ * from parent records (#2245) and displays the resulting summary. The same
```
stevecassidy: Just noting this here but it appears elsewhere too - don't put ticket references in code, they won't mean anything to future readers and we have Github history if we want to trace changes.

web/src/components/forms/recalculate-values-form.tsx
```diff
@@ -0,0 +1,93 @@
...
+  };
+
+  return (
+    <div className="flex flex-col gap-4">
+      <p className="text-sm text-muted-foreground">
+        Some fields show values copied or calculated from a record's parent.
+        These update when a record is next opened, so records not opened since
+        their parent changed can hold outdated values. Recalculating brings
+        every record up to date now. Note that records edited on a device but
+        not yet synced are updated once they sync, through the usual conflict
```
stevecassidy: wording...'Note that if a mobile device has unsynced records, they may overwrite these changes when that data is synced.' (we don't actually have a conflict resolution mechanism at the moment).

web/src/components/tabs/project/export.tsx
```diff
@@ -47,6 +48,17 @@ const ProjectExport = (): JSX.Element => (
       </ListItem>
       <PhotoExportDialog />
     </Card>
+    <Card className="flex flex-col gap-4 flex-1 justify-between">
```
stevecassidy: This should go on the Actions tab rather than here in export (it will happen on export anyway).

stevecassidy: Also, can we hide this if there are no fields in the notebook that would need recalculation?

library/data-model/src/derivedFields/templateFunctions.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export type TemplateFunctionLookup = (args: {
+  namespace: string;
+  name: string;
+}) => TemplateFunction | undefined;
+
+/** Nullable schema for field value (empty when no address entered). */
+const AddressValueNullableSchema = AddressValueSchema.nullable();
+
+/** Template expansion: parse with AddressValueNullableSchema, then return `display_name`. */
```
stevecassidy, line 43: Need some explanation here as to why we have an address field function outside of the field itself.

library/data-model/src/derivedFields/resolveParentValues.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (parents.length === 0) {
+      return null;
+    }
+
+    const parentForms = new Set(
+      getParentFormsForForm({uiSpecification: engine.uiSpec, formId})
+    );
+
+    for (const rel of parents) {
+      const parent = await engine.form.getExistingFormData({
```
stevecassidy, line 59: Here you are getting all data for the parent when you only really need one (or a few) values for those fields mentioned in the expression.  Can we avoid a costly record hydration by adding something to the engine operations to get just some field values for a record? This might address some of the potential complexity of this operation.

library/data-model/src/derivedFields/templatedFields.ts
```diff
@@ -0,0 +1,229 @@
...
+      const template = fieldDetails['component-parameters']?.template;
+      if (!template) {
+        logWarn('TemplatedStringField missing template prop - cannot render.');
+        continue;
+      }
+      if (typeof template !== 'string') {
+        logWarn('TemplatedStringField template prop is not a string.');
+        continue;
+      }
+      fieldsToBeUpdated.push({fieldName, template});
```
stevecassidy, line 201: It's a templated string field but does it have references outside the record? If not we can leave it alone.

library/data-model/src/derivedFields/refreshDerivedValues.ts
```diff
@@ -0,0 +1,282 @@
...
+    recordsFailed: 0,
+  };
+  if (dependentForms.size === 0) return summary;
+
+  for (const [formId, {fields}] of dependentForms) {
+    // Field IDs writable by this pass, per dependency kind.
+    const templateFields = fieldIdsOfKind(fields, 'template');
+    const expressionFields = fieldIdsOfKind(fields, 'expression');
+    const displayFields = fieldIdsOfKind(fields, 'parent-field-display');
+
```
stevecassidy: Some analysis here to find out if any of these fields reference parent data. displayFields will of course but the other two might not.  If they don't you can return early here.  If they do, remember which parent fields they need and get only those values when you pull data from the parent for the context.

stevecassidy (changes requested):
I've left a few comments on potential efficiency improvements for this.  There is one potential issue that you might have two or more levels of reference (GridCell to Layer to Test) which we wouldn't pick up in one pass if you did the records at random.   If you can do some introspection on the uiSpec to work out what the actual dependencies are maybe you can choose an order for recalculation that ensures we work downwards in the tree. 

Also note that we can't test this on lots of data since our random record generator doesn't generate parent/child relations.   Can you extend it to do that - add in a randomly generated child record.  Then try a test with a large number of records for this update and check for significant lag.

library/data-model/src/derivedFields/logging.ts
```diff
@@ -0,0 +1,24 @@
...
+ * See, the License, for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Filename: logging.ts
+ * Description:
+ *   Warn logging for the derived-fields module. Swap the implementation here
+ *   if data-model adopts a shared logging utility.
+ */
+
+export const logWarn = (...args: unknown[]): void => {
```
stevecassidy: We already have a logging.ts module in data-model so make use of that.

library/data-model/src/derivedFields/recordContext.ts
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * @param timestamp - Unix timestamp in milliseconds (e.g., from Date.now())
+ * @param timezone - Optional IANA timezone; defaults to the system timezone.
+ *   Note the default makes output environment-dependent: two devices (or a
+ *   server) in different timezones render different strings.
+ * @returns Formatted date-time string or empty string if input is invalid
+ *
+ * @throws Never - Returns empty string for all error cases
+ */
+export function formatTimestamp(
```
stevecassidy: We have a similar function in the app (same name) - it would be good to remove the duplication and use a single implementation in data-model somewhere - eg. in utils.ts

stevecassidy (changes requested):
Can you refactor the tests to put the ones relating to computing values in the data-model module and make sure that those tests are reasonably complete.   A couple of minor notes but this should be good to go with those changes.

library/data-model/src/derivedFields/logging.ts
Caedous: Done, logwarn added to the existing logging.ts module

library/data-model/src/derivedFields/recordContext.ts
Caedous: Done, single implementation in data-model utils.ts, it is now imported and its timestamp test suite moved across to data-model

Caedous:
Evaluation tests moved to data-model/tests under jest

library/data-model/src/derivedFields/computedFields.ts
```diff
@@ -0,0 +1,225 @@
...
+ * @returns Whether anything changed, and the new values keyed by field name
+ */
+export function recomputeComputedFields({
+  values,
+  uiSpecification,
+  formId,
+  context,
+}: {
+  values: ValuesObject;
+  uiSpecification: UiSpecModel;
```
PeterBaker0: There is a type for compiled ui spec model - risky not asserting that type here than casting it later. The compiled version will have those evaluation functions to avoid line 106 cast

library/data-model/src/derivedFields/computedFields.ts
```diff
@@ -0,0 +1,225 @@
...
+      // off the compiled field definition.
+      const compiledField = fieldDetails as CompiledFieldDefinition;
+      const expressionFn = compiledField.expressionFn;
+      const references = compiledField.expressionRefs;
+      if (!expressionFn || !references) {
+        logWarn(
+          `${componentName} has no compiled expression - cannot evaluate. ` +
+            'Was the UI spec compiled (compileUiSpecConditionals)?'
+        );
+        continue;
```
PeterBaker0, line 114: Extra reason you should force the type rather than cast than throw error. Still good to check but not ideal.

library/data-model/src/derivedFields/computedFields.ts
```diff
@@ -0,0 +1,225 @@
...
+      case 'number': {
+        const n = typeof raw === 'number' ? raw : Number(raw);
+        return Number.isNaN(n) ? null : n;
+      }
+      case 'string':
+        return typeof raw === 'string' ? raw : null;
+      case 'boolean':
+        return typeof raw === 'boolean' ? raw : null;
+    }
+  };
```
PeterBaker0: This feels like a critical function - can you elevate it outside this function scope as a util function and add specific unit tests for it- carefully checking edge cases.

library/data-model/src/derivedFields/computedFields.ts
```diff
@@ -0,0 +1,225 @@
...
+    );
+  };
+
+  // Resolves a _PARENT.<Field-ID> reference from the parent's stored values.
+  // Parent derived fields ARE usable: their stored values carry no evaluation
+  // ordering problem, and persist-on-save is what guarantees they exist.
+  // Field IDs are globally unique in the uiSpec, so the field's type can be
+  // looked up directly.
+  const resolveParentRef = (ref: string): ExprValue | null => {
+    const parentFieldId = ref.slice(PARENT_REFERENCE_PREFIX.length);
```
PeterBaker0: Can you separate out the logic which encodes and decodes the format from the data. e.g. there should be a tested function

a) encodeParentRef(fieldId)=> string which gives you the _PARENT.fieldId (or similar - not verbatim)
b) decodeParentRef(ref: string) => string (fieldId). 

Then unit test thoroughly in both directions.

library/data-model/src/derivedFields/computedFields.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  for (const {fieldName, expressionFn, references} of computedFields) {
+    // Build the scope from referenced symbols: parent references resolve from
+    // the parent record's values, local references from fields in this form.
+    // A reference with no usable value leaves the result blank; any other
+    // symbol is treated as unknown.
+    const scope = new Map<string, ExprValue>();
+    let incomplete = false;
+    for (const ref of references) {
+      if (ref.startsWith(PARENT_REFERENCE_PREFIX)) {
```
PeterBaker0: Separate function for checking this condition - isParentRef() or similar -> this is important logic and isn't really about resolution so much as encoding/decoding. Should be unit tested.

PeterBaker0: I would even recommend enumerating the ref types e.g. REF_TYPES = SYSTEM | FIELD | PARENT_FIELD or similar (whatever makes sense) and having a function (tested) which takes a ref and then resolves the type. Later if we have many we could have a dispatcher a) ref b) ref type c) ref decoder d) ref evaluation

library/data-model/src/derivedFields/computedFields.ts
```diff
@@ -0,0 +1,225 @@
...
+    const previous = values[fieldName];
+    const prev =
+      previous === undefined || previous === null || previous === ''
+        ? null
+        : (previous as ExprValue);
+
+    if (!Object.is(prev, result)) {
+      updates[fieldName] = result;
+      changes = true;
+    }
```
PeterBaker0: Critical function - please split out and unit test and document.

library/data-model/src/derivedFields/computedFields.ts
```diff
@@ -0,0 +1,225 @@
...
+ * parent value that is missing - including when the record has no parent -
+ * leaves the result blank.
+ *
+ * @param values Current form data values
+ * @param uiSpecification The decoded UI spec (with compiled expressions attached)
+ * @param formId The target form ID to update
+ * @param context Record context, carrying parent values if resolved
+ * @returns Whether anything changed, and the new values keyed by field name
+ */
+export function recomputeComputedFields({
```
PeterBaker0, line 67: This function is monolithic and needs more modularised parts with tests. It's mission critical.

library/data-model/src/derivedFields/resolveParentValues.ts
PeterBaker0: Agree - there are a lot of AVP lookups. Would be very easy to pass optional argument with a list of field IDs into the form engine call - e.g. {select: {...}} which hydrates only the matching AVPs - the return type is already null tolerant as not all data is expected to be available. Would recommend doing that in this PR as this will help a fair bit.

library/data-model/src/derivedFields/templatedFields.ts
PeterBaker0: But isn't this also the call path for the normal template update function so is called for normal within record updates too?

library/data-model/src/derivedFields/templateFunctions.ts
PeterBaker0: This explanation makes sense. We may need a 'react part' and 'non react part' of the field registry. But for now please just document this exception in the field authoring guide as this is important change.

PeterBaker0 (changes requested):
While I recognise that this is mostly moving stuff, while here, some changes suggested. 

Agree its worth doing the pretty simple addition to the data engine for filtering which AVPs you hydrate. 

I also think the main resolve function is too monolithic and is tested as a black box. There are particular parts we need to decouple e.g. ref encoding/decoding logic, equality evaluation.

library/data-model/src/derivedFields/computedFields.ts
Caedous: Done, `recomputeComputedFields` takes `CompiledUiSpecModel` and the casts are gone. The guard remains since the compiled schema marks the evaluator optional

library/data-model/src/derivedFields/computedFields.ts
Caedous: Done

library/data-model/src/derivedFields/computedFields.ts
Caedous: Extracted to `derivedFields/valueUtils.ts` as `coerceToExprType`, with unit tests covering numeric strings, mistyped values, blanks, unknown types and the inherited `Number('  ') === 0`.

library/data-model/src/derivedFields/computedFields.ts
Caedous: `uiSpecification/parentReferences.ts` has `encodeParentRef/decodeParentRef`, round-trip tested including dots-in-IDs and bare-prefix cases.

library/data-model/src/derivedFields/computedFields.ts
Caedous: Both added. `isParentRef`, the `RefType` enum (`SYSTEM | FIELD | PARENT_FIELD`) and a tested `resolveRefType`. The dispatcher shape you describe would slot in naturally on top if the ref vocabulary grows.

library/data-model/src/derivedFields/computedFields.ts
Caedous: Extracted as `normaliseStoredValue` + `hasComputedValueChanged` with tests pinning the Object.is semantics (NaN stability, ±0) and blank equivalence.

library/data-model/src/derivedFields/computedFields.ts
Caedous: Addressed via the extractions above. `recomputeComputedFields` is now orchestration over the tested helpers.

library/data-model/src/derivedFields/resolveParentValues.ts
Caedous: Done via Jayen's `engine.hydrated.getFieldValues` (#2263). A scan collects the parent fields the child's templates, expressions and ParentFieldDisplay reference, and only those are fetched from the parent's head. Untrusted scan or conflicting heads fall back to the full fetch, preserving existing semantics.

library/data-model/src/derivedFields/templatedFields.ts
Caedous: Yes this is the shared path, so in-form updates hit it too. The warn only fires for a malformed template prop (non-string), which is a spec-authoring error worth surfacing in both contexts.

library/data-model/src/derivedFields/templateFunctions.ts
Caedous: Documented in the field authoring guide (`05-field-system.md` → Template functions). Where they live, why, and how to add one for a new field.

library/data-model/src/derivedFields/computedFields.ts
```diff
@@ -0,0 +1,194 @@
...
+  uiSpecification,
+  formId,
+  context,
+}: {
+  values: ValuesObject;
+  uiSpecification: CompiledUiSpecModel;
+  formId: string;
+  context?: RecordContext;
+}): {changes: boolean; updates: Record<string, ExprValue | null>} {
+  const fieldMap = getFieldToIdsMap(uiSpecification);
```
stevecassidy: getFieldToIdsMap does a lot of work on the uiSpec but you only want a list of fields in this form - that's explicitly in the uiSpec as views[formId].fields.

library/data-model/src/derivedFields/parentFieldScan.ts
```diff
@@ -0,0 +1,80 @@
...
+ * Description:
+ *   Scans a form's fields for the parent-record fields its templates and
+ *   computed expressions reference, so parent resolution can fetch only
+ *   those fields.
+ */
+
+import {
+  CompiledUiSpecModel,
+  decodeParentRef,
+  encodeParentRef,
```
stevecassidy: unused import

library/data-model/src/derivedFields/parentFieldScan.ts
```diff
@@ -0,0 +1,80 @@
...
+ * complete, so callers fall back to fetching all parent fields.
+ */
+export function referencedParentFields({
+  uiSpecification,
+  formId,
+}: {
+  uiSpecification: CompiledUiSpecModel;
+  formId: string;
+}): string[] | null {
+  const fieldMap = getFieldToIdsMap(uiSpecification);
```
stevecassidy: Again, just uiSpecification.views[formId].fields

stevecassidy (approved):
Just a couple of comments re. finding fields in a form but generally happy with this now, thanks for the work on tidying it up.

library/data-model/src/derivedFields/computedFields.ts
Caedous: Done, reading straight off `viewsets[formId].views`.

library/data-model/src/derivedFields/parentFieldScan.ts
Caedous: Done, reading straight off `viewsets[formId].views`.

library/data-model/src/derivedFields/parentFieldScan.ts
Caedous: Removed

PeterBaker0 (approved):
comments appear resolved and happy with changes overall. Thanks!

Issue #2259, opened 2026-08-20
Improve configuration management for app build workflows


stevecassidy:
The current requirement to use GitHub settings and secrets to configure the application is hard to manage and debug.   It would be preferable to use a similar mechanism to the AWS-CDK deployment where configuration is stored in a separate repository that is pulled by the workflows to generate the configuration environment. 

Ideally we integrate this with the AWS-CDK deployment as well.

PR #2260, opened 2026-08-20
build/Use json config in app workflows

stevecassidy:
# build/Use json config in app workflows

## Ticket

#2259

## Description

The current requirement to use GitHub settings and secrets to configure the application is hard to manage and debug. It would be preferable to use a similar mechanism to the AWS-CDK deployment where configuration is stored in a separate repository that is pulled by the workflows to generate the configuration environment.

Ideally we integrate this with the AWS-CDK deployment as well.

## Proposed Changes

Adds a new library module that implements a .env generation process from a JSON configuration file.    Includes a validation script that checks that all ENV settings required by app/ and web/ and the app build workflows are covered by the script. This can be used in CI to ensure we don't drift when new configuration options are added.

## How to Test

TBD

## Checklist


app/ios/App/fastlane/Fastfile
```diff
@@ -278,7 +223,6 @@ platform :ios do
     )
 
     pilot(
-      apple_id: "#{DEVELOPER_APP_ID}",
```
PeterBaker0, line 281: Is there an explanation for the removal of this parameter?

library/build-config/src/build-config.ts
```diff
@@ -0,0 +1,154 @@
...
+    showStatusTab: z.boolean().default(true).optional(),
+    debugApp: z.boolean().default(false).optional(),
+    debugPouchDb: z.boolean().default(false).optional(),
+    pouchBatchSize: z.number().int().positive().default(10).optional(),
+    pouchBatchesLimit: z.number().int().positive().default(10).optional(),
+    excludedTeamRoles: z.array(z.string()).default([]).optional(),
+    bugsnagApiKey: z.string().optional(),
+    developerMode: z.boolean().default(false).optional(),
+  })
+  .passthrough();
```
PeterBaker0, line 74: Comments on here, or clear reference to authoritative documentation.

library/build-config/src/build-config.ts
```diff
@@ -0,0 +1,154 @@
...
+
+export const BuildConfigSchema = z
+  .object({
+    urls: UrlBuildConfigSchema,
+    app: AppBuildConfigSchema,
+    web: WebBuildConfigSchema,
+    mobile: MobileBuildConfigSchema,
+    secrets: z.record(z.string(), z.unknown()).default({}),
```
PeterBaker0, line 146: Why can't the secrets be fully typed, the structure is retained with SOPS right? So the values can just be strings?

library/build-config/src/generate-build-config.ts
```diff
@@ -0,0 +1,406 @@
...
+function buildEnvMap(
+  config: ReturnType<typeof parseBuildConfig>,
+  platform: SupportedPlatform,
+  options: {includeEmpty?: boolean} = {}
+) {
+  const {includeEmpty = false} = options;
+  const {app, web, mobile, urls} = config;
+  const commitVersion = resolveGitCommitVersion(app.commitVersion);
+
+  const base = {
```
PeterBaker0, line 136: This should be typed

library/build-config/src/generate-build-config.ts
```diff
@@ -0,0 +1,406 @@
...
+      90
+    ),
+    VITE_LONG_LIVED_TOKEN_DURATION_HINTS: coalesce(
+      web.longLivedTokenDurationHints,
+      [1, 5, 10, 30, 90, 365]
+    ),
+    VITE_EXCLUDED_TEAM_ROLES: coalesce(app.excludedTeamRoles, []),
+  };
+
+  const platformSpecific = {
```
PeterBaker0, line 245: So should this - typing

library/build-config/src/generate-build-config.ts
```diff
@@ -0,0 +1,406 @@
...
+  }
+
+  if (platform === 'web' || platform === 'all') {
+    merged.VITE_WEB_URL = base.VITE_WEB_URL;
+    merged.VITE_API_URL = base.VITE_API_URL;
+    merged.VITE_APP_URL = base.VITE_APP_URL;
+    merged.VITE_APP_NAME = base.VITE_APP_NAME;
+    merged.VITE_APP_THEME = base.VITE_APP_THEME;
+    merged.VITE_THEME = base.VITE_THEME;
+  }
```
PeterBaker0, line 308: This is hard to understand what is happening at a glance-  can you clearly document why/what is being merged?

library/build-config/src/generate-build-config.ts
```diff
@@ -0,0 +1,406 @@
...
+
+  return merged;
+}
+
+export function generateEnv({
+  config,
+  platform,
+  includeEmpty = false,
+}: {
```
PeterBaker0, line 325: This file needs doc strings and more explanation of what each step does.

library/build-config/src/validate-generated-env.ts
```diff
@@ -0,0 +1,328 @@
...
+  return parsed;
+}
+
+/**
+ * Read a TypeScript source file and extract the object literal
+ * that defines the environment variable schema.
+ * @param sourceText The text of the TypeScript source file
+ * @returns The text of the object literal that defines the environment variable schema
+ */
+function extractEnvObjectLiteral(sourceText: string): string {
```
PeterBaker0, line 71: This is very hacky! Parsing marker strings from other files to infer required fields!

Build config is not really consumed anywhere, so there isn't any circular dependency issues, right? Can't we just export the config models from the various files (since they are all nice zod models now), and use that? Or even better, expose a function in each app which does it's own validation against it's own zod model.

library/build-config/src/validate-generated-env.ts
PeterBaker0, line 1: Don't like this approach - risky and too much magic - export stuff from the app/api/web which exposes validation hooks or at least zod models-  let's use the validation we've built with zod to parse the env. You might just need something like (pseudocode)

```text
const apiEnv = readEnv('api/.env')
const apiSchema = someSchemaObject # imported from /api/src/buildconfig.ts
const valid = apiSchema.parse(apiEnv)
```

PeterBaker0: This approach also won't check the real semantics of each libraries validation, such as incorrect combinations, incorrect enums, types, etc.

scripts/sync-deployment-config.sh
PeterBaker0, line 1: I like this workflow - it seems sensible

PeterBaker0 (changes requested):
Overall looking pretty good - some missing documentation, and don't like the validation approach.

PR #2261, opened 2026-08-20
feat: Reference related record values in templated and computed fields (#2256)

### LINKED ISSUE #2256: Allow templated and computed fields to reference values from a related record
Follow-up to #2235 (parent record references). Allow TemplatedStringField and computed fields (ComputedNumber/ComputedText) to reference values from a record linked through a Related Records field, when that field allows a single link (`multiple: false`).

Motivating case (FAIMS/GroundSight-FAIMS#70, private): each core cutter test links to one calibration record; the soil mass calculation needs the cutter mass from that calibration.

Proposed syntax, matching the parent work: `{{Core-Calibration.Cutter-Mass-g}}` in templates and `{Core-Calibration.Cutter-Mass-g}` in expressions, where the first segment is the Related Records field's ID and the second a Field ID on the linked form (`related_type`).

- Only single-link Related Records fields are eligible; the designer offers these in the field picker and reports a targeted error for references through multi-link fields
- The linked form is known from `related_type`, so references are type checked against it; the related record's stored templated/computed values are referenceable, as with `_PARENT.`
- Values are resolved on open, on save, and whenever the link is changed while editing; no linked record contributes no value (computed result blank, record saves normally)
- Same staleness contract as parent references (accurate as of the record's last save); handling of stale derived values at export/read time is being worked through separately under #2245 and the resolve-on-read discussion.

Chained references (`Rel._PARENT.X`, `Rel.Rel2.X`) and conditions are out of scope.

Caedous:
## Ticket

Closes #2256.

## Description

Allows TemplatedStringField and computed fields (ComputedNumber/ComputedText) to reference values from a record linked through a Related Records field, when the field uses a **Linked** relation and allows a single link (`multiple: false`). Follow-up to #2246 (parent references): each core cutter test in GroundSight links to one calibration record, and the soil mass calculation needs the cutter mass from it. References join the link field's ID and the target Field ID with a dot — `{{Core-Calibration.Cutter-ID}}` in templates, `{Wet-Mass-g} - {Core-Calibration.Cutter-Mass-g}` in expressions. Child relations remain the `_PARENT.` path and records linking *to* this one are out of scope, per discussion with @stevecassidy.

## Proposed Changes

`relatedForms.ts` joins `parentForms.ts` in data-model: the single-link Linked field scan, the `Rel.Field` type map built from the linked form's declared types, and reference splitting; the shared viewset scan moves to `formScan.ts` (re-exported, so no importer changes). `compileComputedExpressionForForm` merges related reference types into the same compile and adds targeted errors for references through multi-link or Child-relation fields and for fields missing from the linked form — surfaced unchanged by the designer's live validation. In forms, `resolveRelatedValues` (pure, engine-injected, reading links from supplied form values) resolves the linked record's raw values, carried on `RecordContext.relatedValues`; templates receive them under the link field's key (a bare `{{Link-Field}}` keeps its previous rendering via a non-enumerable `toString`) and the computed scope resolver gains the dotted branch. `EditableFormManager` resolves at mount and re-resolves whenever a link changes mid-edit — a store subscription watches a signature of the link-field record IDs so unrelated edits don't trigger it — then recomputes and schedules a save when derived values changed, matching the existing save behaviour per discussion with @stevecassidy. References may target the linked record's stored derived fields, with the same staleness contract as parent references; no linked record means the reference contributes no value, computed results stay blank and the record saves normally. The designer gains an **Insert linked record field** picker on computed fields and "Link field > Field" variables in the Templated String Visual Builder; user docs gain Referencing Linked Record Values sections for all three field types plus a pointer from the Related Records page.

## How to Test

1. `cd library/data-model && npx jest parentForms relatedForms` — 16 new tests (field discovery, typing, splitting, targeted compile errors) plus the existing parent suite
2. `pnpm --filter @faims3/forms test` — 13 new tests (resolver edge cases including wrong-form links and failed loads; template and computed evaluation mirroring the parent suites) plus the existing suites
3. Designer: on a form with a single-link Linked Related Records field, the computed editor validates `{Link-Field.Field-ID}` green, offers the **Insert linked record field** picker, and reports specific errors for multi-link, Child-relation, and unknown-field references; the Visual Builder lists "Link field > Field" variables
4. App: values fill when a record is linked, update when the link is swapped mid-edit, persist on save, don't churn revisions on an untouched reopen, and re-derive on open after the linked record is edited; `_PARENT` and same-form derived fields behave as before

## Additional Information

`resolveRelatedValues` reads links from supplied form values rather than a stored revision, both to support mid-edit re-resolution and so a future resolve-on-read/export refresh (#2245 discussion) can reuse it server-side. If the parked refresh detector is revived, it needs extending to find dotted related references alongside `_PARENT.` ones.

## Checklist


library/data-model/src/uiSpecification/relatedForms.ts
```diff
@@ -0,0 +1,148 @@
...
+    ? {relatedFormId: params.data.related_type, multiple: params.data.multiple}
+    : null;
+};
+
+/**
+ * Every Linked Related Records field on the given form, keyed by field ID. Callers
+ * decide whether multi-link fields are usable; the compile pass reports them
+ * with a targeted error rather than silently dropping them.
+ */
+export const getRelatedRecordFields = ({
```
stevecassidy, line 82: Rather than doing this every time we need to resolve fields, maybe we can incorporate it into the uiSpec compilation process and memoize the result in our internal version of the uiSpec.  Same for parent references: what can you store in the compiled uiSpec that would speed up access to related fields at runtime.   You could have a flag to say whether or not you need to do anything at all (no field references are used).

library/data-model/src/uiSpecification/parentForms.ts
```diff
@@ -41,28 +44,8 @@ import {
...
-  const viewset = uiSpecification.viewsets[viewSetId];
-  if (!viewset) return [];
-  const ids: string[] = [];
-  for (const viewId of viewset.views) {
-    ids.push(...(uiSpecification.views[viewId]?.fields ?? []));
-  }
-  return ids;
-};
+// Scan helpers live in formScan.ts; re-exported so existing importers hold.
+export {fieldIdsForViewset, type ParentScanUiSpec} from './formScan';
```
stevecassidy, line 48: I don't see the value in this, just fix the imports, there aren't too many of them

library/forms/lib/formModule/formManagers/EditableFormManager.tsx
```diff
@@ -552,6 +558,79 @@ export const EditableFormManager: React.FC<
...
+    let latest = 0;
+    let lastSignature: string | null = null;
+
+    const refresh = () => {
+      const values = currentValues();
+      const signature = signatureOf(values);
+      if (signature === lastSignature) return;
+      lastSignature = signature;
+      const token = ++latest;
+      resolveRelatedValues({
```
stevecassidy: This calls getRelatedRecordFields - you could pass linedFields in so it didn't have to

library/data-model/src/uiSpecification/relatedForms.ts
stevecassidy: It gets called here in compiling expressions and by the EditableFormManager when opening a form.  Not too costly unless there are many forms and many fields in the notebook.  I'm not sure whether memoizing is a good idea but worth thinking about and if there is a clean approach, putting it in.

library/forms/lib/formModule/formManagers/EditableFormManager.tsx
stevecassidy, line 594: Right, so we're doing this for every value change in the form, so the cost of getRelatedRecordFields will be felt here, if the uiSpec is very large, it could be significant so compiling into the uiSpec might be useful.

stevecassidy (commented):
Generally looks good but I'm concerned about the potential time cost of this.   Left some notes about possible compilation of link reference info into the uiSpec - can you take a look at that. Happy to discuss. Maybe you can test on a very large notebook - generate one with 500 fields and see if there is any slowdown.

library/data-model/src/uiSpecification/relatedForms.ts
Caedous: Benchmarked on a generated 500-field/10-form notebook and the scans run 0.003-0.006ms/call and the whole compile pass averages 2.7ms, so I left them uncompiled.

library/data-model/src/uiSpecification/parentForms.ts
Caedous: Done, re-export removed, imports point at formScan directly.

library/forms/lib/formModule/formManagers/EditableFormManager.tsx
Caedous: Fixed, the signature now reads just the link fields off the store (0.115ms to 0.0002ms per change at 500 fields), and forms with no single-link fields never mount the subscription.

stevecassidy (approved):
Looks good now. Thanks.

PR #2263, opened 2026-08-20
feat: read a record's field values from the data engine

jayenashar:
## Ticket

None.

## Description

Reading a few field values across many records currently means reaching into revisions and AVPs from application code. This moves that read into the data engine, where the rest of the revision and AVP handling lives.

## Proposed Changes

- `engine.hydrated.getFieldValues({recordId, revisionId, fields})` returns the values it finds for the named fields, without hydrating the whole record. A missing revision or AVP skips that record or field, so one bad record cannot blank a listing; any other read failure is rethrown, so a systemic fault surfaces rather than reading as no data.
- `extractFeaturesFromRecord` reads through it, so the overview map's skip and rethrow behaviour is now covered by the engine's own tests rather than by mocks in the app.

## How to Test

`pnpm --filter @faims3/data-model test` and `pnpm --filter @faims3/app test`. The overview map should behave unchanged: records still plot, a record with a missing revision is skipped rather than failing the map, and an unopenable database surfaces an error instead of an empty map.

## Additional Information

Prompted by review feedback that this read belongs in the database engine rather than in application code.

## Checklist

### DIFF
--- a/app/src/gui/components/notebook/recordFeatures.test.tsx
+++ b/app/src/gui/components/notebook/recordFeatures.test.tsx
@@ -160,16 +160,15 @@ describe('useRecordFeatures', () => {
       form_id: 'Site',
     });
     expect(getFieldValues).toHaveBeenCalledWith({
-      recordId: 'rec-site',
       revisionId: 'frev-site',
       fields: expect.arrayContaining(['site_location', 'orphan_location']),
     });
   });
 
-  it('keeps other records when one has no readable values', async () => {
-    getFieldValues.mockImplementation(({recordId}: {recordId: string}) =>
+  it('still builds the collection when one record contributes nothing', async () => {
+    getFieldValues.mockImplementation(({revisionId}: {revisionId: string}) =>
       Promise.resolve(
-        recordId === 'rec-obs' ? {} : {site_location: POINT_FEATURE}
+        revisionId === 'frev-obs' ? {} : {site_location: POINT_FEATURE}
       )
     );
 
--- a/app/src/gui/components/notebook/recordFeatures.test.tsx
+++ b/app/src/gui/components/notebook/recordFeatures.test.tsx
@@ -159,8 +159,26 @@ describe('useRecordFeatures', () => {
       revision_id: 'frev-site',
       form_id: 'Site',
     });
-    expect(getFieldValues).toHaveBeenCalledWith(
-      expect.objectContaining({recordId: 'rec-site', revisionId: 'frev-site'})
+    expect(getFieldValues).toHaveBeenCalledWith({
+      recordId: 'rec-site',
+      revisionId: 'frev-site',
+      fields: expect.arrayContaining(['site_location', 'orphan_location']),
+    });
+  });
+
+  it('keeps other records when one has no readable values', async () => {
+    getFieldValues.mockImplementation(({recordId}: {recordId: string}) =>
+      Promise.resolve(
+        recordId === 'rec-obs' ? {} : {site_location: POINT_FEATURE}
+      )
+    );
+
+    const result = await renderFeatures([siteRecord, obsRecord]);
+
+    expect(result.current.isError).toBe(false);
+    expect(result.current.data!.features).toHaveLength(1);
+    expect(result.current.data!.features[0].properties?.record_id).toBe(
+      'rec-site'
     );
   });
 
--- a/app/src/gui/components/notebook/recordFeatures.test.tsx
+++ b/app/src/gui/components/notebook/recordFeatures.test.tsx
@@ -9,7 +9,6 @@
  */
 import {
   compileUiSpecConditionals,
-  DocumentNotFoundError,
   MinimalRecordMetadata,
   type CompiledNotebookUiSpec,
   type NotebookUiSpec,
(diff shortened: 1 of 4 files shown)


library/data-model/src/databaseEngine/engine.ts
```diff
@@ -931,6 +931,63 @@ class HydratedOperations {
...
 
+  /**
+   * Read the given fields' stored values from one revision, without hydrating
+   * the whole record. Returns the stored values rather than the hydrated field
+   * shape, since callers of this want the value and not the AVP envelope.
+   *
+   * A missing revision or AVP skips that record or field; any other read
+   * failure is rethrown, so a systemic fault does not read as no data.
+   *
+   * TODO not optimal for efficiency: a read per revision plus one per field
```
jayenashar: new in this branch — a TODO in the JSDoc shows in every tooltip for this method. main had it inline above the read it describes; moving it back there.

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -931,6 +931,63 @@ class HydratedOperations {
...
+   *
+   * @param recordId - The record the revision belongs to, for the skip warnings
```
jayenashar: new in this branch — `recordId` changes no result, it only feeds the two warn strings. dropping it, the revision id traces back to the record.

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -931,6 +931,63 @@ class HydratedOperations {
...
+   * @param revisionId - The revision to read
+   * @param fields - The field names to read
+   * @returns The values found, keyed by field name; absent fields are omitted
```
jayenashar: new in this branch — main dropped every falsy value (`if (!dataRaw) return`), this keeps `''`, `0` and `null` and omits only `undefined`. no effect on the map since the parse rejects them, but this line says otherwise. fixing the wording rather than the behaviour: a general reader should get a stored `0` back.

app/src/gui/components/notebook/recordFeatures.test.tsx
```diff
@@ -165,13 +159,33 @@ describe('useRecordFeatures', () => {
...
     });
-    expect(getRevision).toHaveBeenCalledWith('frev-site');
+    expect(getFieldValues).toHaveBeenCalledWith({
+      recordId: 'rec-site',
+      revisionId: 'frev-site',
+      fields: expect.arrayContaining(['site_location', 'orphan_location']),
+    });
+  });
+
+  it('keeps other records when one has no readable values', async () => {
```
jayenashar: new in this branch — this covers the hook still building a collection when one record contributes nothing, not the values themselves. renaming to say that.

jayenashar (commented):
self-review before marking ready. each thread says whether it is new in this branch or carried across from main; fixes for the new ones are pushed next, the carried ones are flagged only.

body nit, also fixing: "covered by the engine's own tests rather than by mocks in the app" no longer holds, the app keeps a mocked skip test too.

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -931,6 +931,63 @@ class HydratedOperations {
...
+      console.warn(
+        `Skipping record ${recordId}: revision ${revisionId} not found`
+      );
+      return undefined;
+    });
+    if (!revision) return values;
+
+    // Not fetchAvps: it fails the whole read on a missing AVP, which
+    // hydration wants and a per-field read does not.
+    await Promise.all(
```
jayenashar, line 967: carried across from main — a read per revision plus one per field, for every record in the list. unchanged here, and the TODO above it is the same one main carried.

library/data-model/src/databaseEngine/engine.ts
```diff
@@ -931,6 +931,63 @@ class HydratedOperations {
...
+      fields.map(async field => {
+        const avpId = revision.avps[field];
+        if (!avpId) return;
+        try {
+          const avp = await this.core.getAvp(avpId);
+          if (avp?.data !== undefined) values[field] = avp.data;
+        } catch (error) {
+          if (!(error instanceof Exceptions.DocumentNotFoundError)) throw error;
+          console.warn(
```
jayenashar, line 976: carried across from main — both warn strings came over verbatim. one line per skipped record on a broken listing, and a library caller cannot quiet them. not changing it here.

app/src/gui/components/notebook/recordFeatures.ts
```diff
@@ -63,81 +62,37 @@ export const getGISFields = (uiSpec: NotebookUiSpec): string[] => {
...
- * Extract features from a single record for the given GIS fields.
- *
- * A revision or AVP that is simply missing skips its record or field, and a
- * malformed GIS value skips its field, so one bad record cannot blank every
- * record's geometry. Any other read failure is rethrown, so a systemic fault
- * surfaces as an error rather than as an empty map.
+ * Extract features from a single record for the given GIS fields. A malformed
+ * GIS value skips its field, so one bad value cannot blank a record's geometry.
  */
 export const extractFeaturesFromRecord = async (
```
jayenashar, line 68: carried across from main — exported, but only this module calls it (line 179).

app/src/gui/components/notebook/recordFeatures.ts
```diff
@@ -63,81 +62,37 @@ export const getGISFields = (uiSpec: NotebookUiSpec): string[] => {
...
-  );
+  const values = await dataEngine.hydrated.getFieldValues({
+    recordId: record.recordId,
+    revisionId: record.revisionId,
+    fields,
+  });
 
-  return features;
+  const baseProperties: RecordFeatureProps = {
+    // TODO bring back HRID - or maybe only on records we click on?
```
jayenashar, line 79: carried across from main — TODO unchanged.

PeterBaker0 (approved):
Makes sense - would be useful in #2252 as discussed.

PR #2264, opened 2026-08-21
fix: security - prevent header injection from form name

PeterBaker0:
- Stop putting raw form labels into Content-Disposition, so a quote or CR/LF in a form name cannot change the download filename or inject extra headers.

- Strip control characters from designer form/section labels, slugify designer JSON download names, and HTML-escape usernames in email bodies.

### DIFF
--- a/api/src/couchdb/export/utils.ts
+++ b/api/src/couchdb/export/utils.ts
@@ -373,6 +373,41 @@ export const slugifyLabel = (label: string, maxLength = 50): string => {
 };
 
 /**
+ * Filename stem from user-controlled text (form labels, project ids).
+ *
+ * Quotes, path separators, control characters, and other header
+ * metacharacters are stripped so the result is safe to place in
+ * Content-Disposition `filename="..."`.
+ */
+export const sanitizeDownloadFilename = (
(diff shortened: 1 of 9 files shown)


api/src/couchdb/export/utils.ts
```diff
@@ -372,6 +372,41 @@ export const slugifyLabel = (label: string, maxLength = 50): string => {
...
+const CONTENT_DISPOSITION_FILENAME = /[^\w.-]/g;
+
+/**
+ * Build a Content-Disposition attachment header.
+ *
+ * Restricts the filename to `[A-Za-z0-9._-]` so user-controlled names cannot
+ * terminate the quoted-string, inject extra parameters, change the extension,
+ * or smuggle CR/LF into the header.
+ */
+export const contentDispositionAttachment = (filename: string): string => {
```
stevecassidy, line 400: This is always called with the result of sanitizeDownloadFilename + some fixed text so doesn't seem necessary apart from the slice operation.  Maybe these could be combined?

stevecassidy (approved):
Looks good apart from that one comment.

PR #2269, opened 2026-08-21
feat: let notebook settings carry keys the schema does not declare

### LINKED ISSUE #2267: Allow notebook settings to carry keys FAIMS3 does not declare
`NotebookSettingsSchema` in `@faims3/data-model` is not `.passthrough()`, so any key it does not declare is dropped. A downstream build cannot add its own notebook settings without editing the schema here and adding fields to the designer.

Some form of extension point would avoid naming downstream keys in FAIMS3.

jayenashar:
## Ticket

Closes #2267

## Description

`NotebookSettingsSchema` declares a closed set of keys, so a build on top of FAIMS3 cannot own a notebook setting: the key is dropped on parse, and there is no type under which to read it.

## Proposed Changes

- `NotebookSettingsSchema` passes through keys it does not declare, like the field, section and form schemas above it. The owning module namespaces its keys and validates them itself.

## How to Test

`pnpm --filter @faims3/data-model test`

## Additional Information

Such a key already survived at runtime: `NotebookUiSpecSchema` and `CompiledNotebookUiSpecSchema` both intersect a passthrough `UiSpecModelSchema`, which carried the whole `settings` object through. What changes here is the declared contract and the type. Without the passthrough, `NotebookSettings` has no index signature, so reading an undeclared key does not compile. The tests cover the settings schema on its own and both composed schemas a consumer actually parses.

An earlier revision of this PR also added a registry so a module could contribute a section to the designer's Settings panel. That is removed as out of scope: nothing here registers one, and `web/` is not consumed as a library, so a downstream build authors its own designer UI in its own copy.

## Checklist


jayenashar:
Closing: the change no longer has a consumer.

The passthrough was for four `groundsight*` keys a downstream build kept in notebook settings. Review on that build has since moved them onto the plan that uses them, where they belong: they configure one plan's workflow, and a template carrying two such plans would otherwise have had one set to share. Nothing now reads a notebook setting the schema does not declare.

The case for it on its own merits is real but thin: the field, section and form schemas above it are already passthrough, so this would make the set consistent. Worth reopening if a build wants to own a notebook setting, and #2267 stays open to describe the gap.

@stevecassidy this is the one you suggested on 20 August when the settings lived in the notebook design. @PeterBaker0 sorry for the review request on something with no caller.

jayenashar:
Correction to the comment above: #2267 is closed too, not left open. Same reasoning, and it is quick to reopen if the modular direction wants it.

Issue #2265, opened 2026-08-21
Bring NotebookView and the plan view registry back into FAIMS3


jayenashar:
The notebook body is rendered by `NotebookView`, which resolves a component from a plan view registry keyed on the notebook plan's `planType` and falls back to the existing `NotebookComponent`. That seam lives in a downstream build and has not come back here yet. Worth doing before the two diverge further, since the notebook route work below sits on top of it.

jayenashar:
Done in #2271, merged 2026-08-26.

*Drafted with AI assistance, reviewed before posting.*

PR #2271, opened 2026-08-21
Bring NotebookView and the plan view registry into FAIMS3

jayenashar:
## Ticket

#2265

## Description

The notebook body is rendered by `NotebookComponent`, with no way for a notebook to ask for a different view. This adds the plan seam #2265 describes.

## Proposed Changes

- A plan type registry in the data model, with two built in types: Counted and List of Records.
- An optional `plan` on a notebook definition and an optional `planTemplate` on a template definition, both validated against the registry.
- `POST /api/notebooks` instantiates the template's plan template into the new notebook's plan, from a `planConfig` in the body.
- An optional `planReference` on a record, so a plan view can tell claimed plan entries from unclaimed ones.
- `NotebookView` resolves a view by the plan's `planType` and falls back to `NotebookComponent`.
- Template mode in the designer, the only mode that offers the plan template controls.

## Review notes

Not everything here is a straight port. These parts changed and are worth reading rather than skimming:

- `normalize.ts` runs one migrate-and-validate pass for notebooks and templates. The separate template path had lost the input size cap and validated its plan in a different order.
- `safeValidatePlanTemplate` returns `ValidationResult<PlanTemplate>` rather than `ValidationResult<Plan>`.
- The plan views no longer read a record's absence from the list as proof it does not exist. `CountedPlanView` holds off reporting the target as reached while the count may be understated, and `ListOfRecordsPlanView` offers Create Record only to a user who may create.
- The designer carries an instantiated `plan` through untouched, so a designer edit does not strip the plan from the saved notebook.
- Three doc comments were wrong and are corrected: `designerMode` on `DesignerDialogProps` described a filename, `PutUpdateTemplateUiSpecificationInput` said notebook where it is a template, and `notebookAdapters.ts` had a duplicated line.
- `createNotebookFromTemplate`'s doc comment is now three lines, to say that a template carrying a plan template needs a matching `planConfig`. Say if that is longer than you want and I will trim it.

## How to Test

Add `"plan": {"planType": "Counted", "formType": "Main", "numberRequired": 3, "allowExtraRecords": false}` next to `uiSpec` in `api/notebooks/e2e-minimal.json`, load it with `pnpm --filter @faims3/api load-projects`, and open it in the app. It shows a Counted Plan banner and a Planned tab rather than the usual record tabs.

## Additional Information

A notebook with no plan, or with a plan type that has no registered view, renders exactly as it does today.

The dashboard cannot collect a `planConfig` yet, so creating from a plan carrying template has to go through the API. The create form says so and disables submission rather than letting the request fail.

## Checklist

### DIFF
--- a/api/src/couchdb/templates.ts
+++ b/api/src/couchdb/templates.ts
@@ -4,6 +4,8 @@ PouchDB.plugin(PouchDBFind);
 PouchDB.plugin(require('pouchdb-security-helper'));
 
 import type {
+  NotebookDefinition,
+  Plan,
   TemplateApiDocument,
   TemplateApiListItem,
 } from '@faims3/data-model';
@@ -23,7 +25,10 @@ import {
   TEMPLATES_BY_TEAM_ID,
   TEMPLATES_LISTING_BY_TEAM_ID,
   TEMPLATES_LISTING_BY_TEMPLATE_ID,
+  getPlanTypeDefinition,
+  normalizeNotebookTemplateUiSpecification,
   normalizeRootDescriptionForStore,
+  notebookUiSpecificationValidationMessage,
 } from '@faims3/data-model';
 import {getTemplatesDb} from '.';
 import * as Exceptions from '../exceptions';
@@ -31,7 +36,7 @@ import {nowIso} from '../time';
 import {generateRandomString} from '../utils';
 import {
   clearTemplateIdFromProjectsReferencingTemplate,
-  normalizeUiSpecificationOrThrow,
+  createNotebook,
 } from './notebooks';
 import {getTeamById} from './teams';
 import {stripTemplateRolesForTemplateId} from './users';
@@ -239,9 +244,16 @@ export const createTemplate = async ({
 
   // Setup the document with id included
   const now = nowIso();
-  const uiSpecification = normalizeUiSpecificationOrThrow(
-    payload.uiSpecification
-  );
+  let uiSpecification;
+  try {
+    uiSpecification = normalizeNotebookTemplateUiSpecification(
+      payload.uiSpecification
+    );
+  } catch (error) {
+    throw new Exceptions.ValidationException(
+      notebookUiSpecificationValidationMessage(error)
+    );
+  }
   const templateDoc: TemplateDocument = {
     _id: templateId,
     version: 1,
@@ -393,8 +405,15 @@ export const updateTemplateUiSpecification = async (
     | PutUpdateTemplateUiSpecificationInput
     | NotebookUiSpecificationInput
 ): Promise<ExistingTemplateDocument> => {
-  const normalizedUiSpecification =
-    normalizeUiSpecificationOrThrow(uiSpecification);
+  let normalizedUiSpecification;
+  try {
+    normalizedUiSpecification =
+      normalizeNotebookTemplateUiSpecification(uiSpecification);
+  } catch (error) {
+    throw new Exceptions.ValidationException(
+      notebookUiSpecificationValidationMessage(error)
+    );
+  }
   // Now fetch the existing template - this will allow us to get the latest
   // revision etc
   let existingTemplate;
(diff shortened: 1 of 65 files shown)


app/src/gui/components/notebook/notebookView.tsx
```diff
@@ -0,0 +1,320 @@
...
+  // delegate to the plan view component
+  if (PlanComponent) {
+    return <PlanComponent {...props} />;
+  }
+
+  // fallback to the default notebook component
+  // TODO: port this component to use the same interface
+  // as our custom plan view components once we have sorted
+  // out what that interface looks like
+  return <NotebookComponent project={project} />;
```
stevecassidy, line 319: We should refactor NotebookComponent to use the new API

api/src/couchdb/templates.ts
```diff
@@ -558,3 +577,102 @@ export const restoreTemplateFromArchive = async (id: string) => {
...
+  if (template.uiSpecification.planTemplate) {
+    // what kind of plan is it
+    const planTemplate = template.uiSpecification.planTemplate;
+    const planType = planTemplate.planType;
+    // Get the instantiation function for this plan type
+    const planTypeDefinition = getPlanTypeDefinition(planType);
+    if (!planTypeDefinition) {
+      throw new Exceptions.InternalSystemError(
+        `Cannot create a notebook from template ${template._id} because it has a plan template of type ${planType}, which is not recognised.`
+      );
```
PeterBaker0, line 625: I guess 500 here makes sense since prior parsing at the http level should have caught it? Otherwise should it be 400 error?

app/src/gui/components/notebook/notebookView.tsx
PeterBaker0: Agreed - it can be follow up if painful... but is good demonstration that this plugin interface is feature complete against prior view

library/data-model/src/plans/countedPlan.ts
```diff
@@ -0,0 +1,74 @@
...
+}) => {
+  if (!countedPlanTemplateSchema.safeParse(template).success) {
+    throw new Error('Invalid counted plan template');
+  }
+  return {
+    planType: template.planType,
+    formType: template.formType,
+    numberRequired: config.numberRequired,
+    allowExtraRecords: config.allowExtraRecords,
+  } as CountedPlan;
```
PeterBaker0: no need to cast this here - instead put type on the function or use satisfies instead.

library/data-model/src/plans/planTypeMap.ts
PeterBaker0, line 1: some type wizardry - I assume all useful/needed!

app/src/gui/components/notebook/plans/ListOfRecordsPlanView.tsx
PeterBaker0, line 1: Image This doesn't work very well for forms with many fields - can you reuse the field selector with fuzzy search. E.g. search/select the field, pops up underneath then can be removed.

PeterBaker0 (changes requested):
Agree with Steve's comment to integrate notebook component - but that's okay if follow up ticket. 

Some minor requests. 

The only bigger change request is to improve the ListPlan designer form to use the unified field selector for consistency and scalability when we get to a large number of fields, as per screenshot.

library/data-model/src/plans/countedPlan.ts
jayenashar: now in [ab76bba](https://github.com/FAIMS/FAIMS3/pull/2271/commits/ab76bba0a3427546e73ee60e2e7c3ecf0bb98123), same for `instantiateListPlan`

app/src/gui/components/notebook/plans/ListOfRecordsPlanView.tsx
jayenashar: now a field search scoped to the chosen form, with the picked fields as removable chips underneath, in [adc647f](https://github.com/FAIMS/FAIMS3/pull/2271/commits/adc647f40f9fb7555b336798130858cd513f6dc9) and [4c1c70a](https://github.com/FAIMS/FAIMS3/pull/2271/commits/4c1c70ac1fa103e2165b90f4aa6782d739ec603d)

api/src/couchdb/templates.ts
jayenashar: the plan template comes from the stored template, not the request, and it is validated against the registry when the template is saved, so an unrecognised type here means the stored doc does not match this build. the 400s below it are the caller's `planConfig`.

app/src/gui/components/notebook/notebookView.tsx
jayenashar: #2275

PeterBaker0 (approved):
Approved - added feature flag to disable just the designer add button (defaults to visible so you don't need to take any action for your use case). 

For the general case, this feature shouldn't be unusable without API actions, as discussed with Steve there needs to (at least for built in plan types) be an end-to-end user accessible pathway to realising a plan - we are missing the instantiation step (template - (planConfig) -> survey). This will need some workflow/component registry similar to the other parts of the plan.

PR #2273, opened 2026-08-24
fix: keep the record list's identity stable across renders

jayenashar:
## Ticket

None.

## Description

`useRecordList` returned a fresh object every render, so a caller memoising on it rebuilt that memo every render and remounted whatever it built. A component mounted that way loses the state it holds itself, such as a table's page, sort and typed search text.

## Proposed Changes

- The hook memoises its return, so an unchanged result set keeps its identity.
- It exposes `refetch` in place of the query object: a react-query result is not identity stable across renders, while its `refetch` is.
- The empty-list fallback for a query that has not loaded is a shared constant, so the memos below it hold before the first fetch lands too.

## How to Test

`pnpm --filter @faims3/app test`. The notebook's record tabs should behave unchanged: records list, the refresh button still refetches, and tab switching keeps its state.

## Additional Information

Stacked on #2263 and merges it, so review that one first. GitHub will retarget this to `main` when it lands.

## Checklist

### DIFF
--- a/app/src/utils/customHooks.tsx
+++ b/app/src/utils/customHooks.tsx
@@ -667,15 +667,29 @@ export const useRecordList = ({
     );
   }
 
-  // return both curated record lists and the underlying query where necessary
-  return {
-    allRecords: nonDraftRecords,
-    myRecords: myRecords,
-    otherRecords: otherRecords,
-    isLoading,
-    canReadAllRecords,
-    initialQuery: unhydratedRecordQuery,
-  };
+  // Memoized so an unchanged result set keeps its identity and a caller's own
+  // memos hold; a fresh object each render remounts whatever they build. The
+  // query's own result object is not identity stable, so this exposes its
+  // refetch (which is) rather than the query itself.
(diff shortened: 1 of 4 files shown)


app/src/utils/customHooks.tsx
```diff
@@ -586,8 +589,9 @@ export const useRecordList = ({
    */
   const isLoading = unhydratedRecordQuery.data === undefined;
```
jayenashar, line 590: new in this branch — the doc above still says "deliberately not `initialQuery.isLoading`", and `initialQuery` is what this PR removes. updating it to name the query directly.

jayenashar (commented):
self-review before marking ready. both items are new in this branch; fixes pushed next.

body nit, also fixing: "a table mounted that way loses its page, sort and typed search text" implies a table in this repo does that. none does yet, so it should describe the shape rather than name a symptom a reviewer cannot reproduce.

PeterBaker0 (approved):
Seems fine to me - sensible. Thanks. Basic testing doesn't show any issues.

PR #2274, opened 2026-08-25
fix: pin trapeze project build package at 7.1.8 so that we don't get trust downgrade errors

PeterBaker0:
Pin by installing package explicitly with peer dep override then using pnpm exec instead of dlx. 

Alternatively we could exclude from the trust downgrade policy?


PeterBaker0:
https://github.com/ionic-team/trapeze/issues/260

PeterBaker0:
Closing as upstream fixed issue.

PR #2276, opened 2026-08-25
feat: adding updated timestamp to records, and managing rev/avp timestamps better

PeterBaker0:
To progress in the direction of having server side, efficient filtering of records by timestamps (to enable a 'give me records after X timestamp' query, efficiently served on DB side with pagination), we need to be able to emit an index off the record with an updated timestamp as the key. 

Adds the updatedAt timestamp field to records in the data engine types, and catches stragglers not yet on the new engine. 

There was also a severe bug for the data migrations which meant that the data DBs were never migrated (namely a concat rather than push array bug). Adds audit logging and some unit testing around this fix. 

There are some trade offs here as to thrashing the record with many writes which increases odds of conflict vs deferring down to the rev layer for updates. This tries to balance by only writing the update timestamp any time the record would have been touched already (i.e. when creating a new revision). We may consider freshening this with some debounced form hook... but needs more understanding of conflict implications.

@stevecassidy  just noting that the API tests were failing because the backup notebook files were old versions. I experimented with just migrating them instead of changing the jsonl files. This required a small refactor to expose the data DB migration step from the migration framework. What are your thoughts?

a) keep it as in this PR - i.e. we just migrate the jsonl backups to get them to latest version 
b) keep doing as we were before, where any breaking data DB change requires manual updating (possibly with script assistance) to get to latest version


api/src/couchdb/backupRestore.ts
```diff
@@ -161,4 +165,8 @@ export const restoreFromBackup = async ({
     await file.close();
   }
   console.log(`Restore completed. Total records processed: ${processedCount}`);
+
+  // Restored data DBs may predate data v2 (`updatedAt` on records/revisions).
+  // Init writes the DB at the current design docs but does not migrate content.
+  await migrateAllProjectDataDbs();
```
stevecassidy, line 171: I think this makes sense in general- we don't know when a backup was created so migrating it after restore is the right thing to do.

library/data-model/src/databaseEngine/types.ts
```diff
@@ -31,6 +31,8 @@ export const v1RecordDBFieldsSchema = z
     record_format_version: z.number(),
     created: z.string().datetime(),
     created_by: z.string(),
+    /** When the current head revision was last created or replaced. */
+    updatedAt: z.string().datetime(),
```
stevecassidy: showing as deprecated use `z.iso.datetime()` instead

stevecassidy (approved):
Happy with this following our discussion.  Hopefully not a big hit on the database.

library/data-model/src/databaseEngine/types.ts
PeterBaker0: thanks

PR #2277, opened 2026-08-25
feat: offline map download redesign

xinyuhoulucy:
# feat: offline map download redesign

## Ticket

[BSS-1408](https://jira.csiro.au/browse/BSS-1408)

## Description

Redesign the offline map download workflow and improve the shared `MapComponent` behaviour used by the new selection UI.

## Proposed Changes

- Redesign the offline map download flow with separate pages for creating, editing, and managing offline maps.
- Allow offline maps to have user defined display labels (including duplicate names), while using a unique internal id.
- Separate desktop rectangle drawing and a mobile touch friendly fixed selection frame for choosing offline map download regions.
- Changes in `MapComponent`:
    - optionally auto fly to the first successful current location and avoid unexpected re-centering when the current location state changes,
    - animate center and extent changes,
    - preserve the current map view when a center or extent is cleared,
    - add optional navigation locking for pan, zoom, and rotation when the map is used for preview only scenarios. (e.g. the offline map edit and project offline map prompt pages)

## How to Test

Please help test all `MapComponent` related features and stores on both web and mobile, including offline map creation and editing, region selection, map navigation, current-location behaviour, and map preview states.


app/src/constants/routes.tsx
```diff
@@ -8,7 +8,7 @@
  *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing software
- * distributed under the License is distributed on an "AS IS" BASIS
+ * distributed under the License is distributed on an "AS IS" BASIS,
```
PeterBaker0: unnecessary change

library/forms/lib/components/maps/MapComponent.tsx
```diff
@@ -321,6 +320,109 @@ const MapComponentImpl = (props: MapComponentProps) => {
...
+    if (!map) {
+      return;
+    }
+
+    const view = map.getView();
+    const isInitialView = !hasInitialisedViewRef.current;
+
+    if (isInitialView) {
+      hasInitialisedViewRef.current = true;
+    }
```
PeterBaker0: Can you add some more explanation of the various states/transitions here re: first vs later center and how it interacts with config

PeterBaker0 (changes requested):
Code looks great, and the functionality works perfectly. Well done. 

As discussed, lets take the time to build a basic migration framework and initialisation gate around the tile store to avoid the legacy code paths leaking into mainline code. 

Also, would it be possible to add a link to the 'offline map details' page (given its a route should be easy) from the survey settings page: 

Image

Here - we can have something, e.g. a link 'manage download' or similar which takes them to the download list entry for that layer.

library/forms/lib/components/maps/migrations/versions/migrateV2.test.ts
```diff
@@ -0,0 +1,311 @@
...
+
+    transaction.onabort = () =>
+      reject(transaction.error ?? new Error('IndexedDB transaction aborted'));
+
+    transaction.onerror = () => {
+      // The transaction will abort after an unhandled request error.
+      console.error('IndexedDB transaction error', transaction.error);
+    };
+  });
+}
```
PeterBaker0: Can you move this + other test setup + mocking etc machinery into a test-utils.ts file in this folder.

library/forms/lib/components/maps/migrations/versions/migrateV2.test.ts
```diff
@@ -0,0 +1,311 @@
...
+    setName: '@project/project-a',
+    extent: [0, 0, 1, 1],
+    minZoom: 2,
+    maxZoom: 14,
+    size: 10,
+    expectedTileCount: 1,
+    created: new Date('2026-01-01T00:00:00Z'),
+    tileKeys: [['tile-a']],
+    projectId: 'project-a',
+  });
```
PeterBaker0: this should be typed - your migration and migration tests should clearly have references to Tile types for V1/V2 etc.  So would expect tileSets.put({...} satisfies TileSetV1Schema) or similar - to typecheck that you are inputting old data.

library/forms/lib/components/maps/migrations/versions/migrateV2.test.ts
```diff
@@ -0,0 +1,311 @@
...
+  const transaction = db.transaction(storeName, 'readonly');
+
+  const result = await requestAsPromise(
+    transaction.objectStore(storeName).getAll()
+  );
+
+  await transactionAsPromise(transaction);
+
+  return result as T[];
+}
```
PeterBaker0: More machinery

library/forms/lib/components/maps/migrations/versions/migrateV2.test.ts
```diff
@@ -0,0 +1,311 @@
...
+ */
+async function runMigrationForTest(db: IDBDatabase): Promise<void> {
+  const transaction = db.transaction(['tiles', 'tileSets'], 'readwrite');
+  const completion = transactionAsPromise(transaction);
+
+  await migrateV1ToV2({db, transaction});
+  await validateV2(transaction);
+  // Wait for all validation requests in the transaction to finish.
+  await completion;
+}
```
PeterBaker0: machinery

library/forms/lib/components/maps/migrations/versions/migrateV2.ts
```diff
@@ -0,0 +1,214 @@
...
+  minZoom: z.number(),
+  maxZoom: z.number(),
+  size: z.number(),
+  expectedTileCount: z.number(),
+  created: z.date(),
+  tileKeys: z.array(z.unknown()),
+  projectId: z.string().optional(),
+  label: z.string().optional(),
+  offlineMapRegion: z.unknown().optional(),
+});
```
PeterBaker0: These should be named and exported as V1 types - so you can refer to them in migrateV2.test.ts as well.

library/forms/lib/components/maps/migrations/versions/migrateV2.ts
```diff
@@ -0,0 +1,214 @@
...
+
+      try {
+        visit(cursor);
+        cursor.continue();
+      } catch (error) {
+        reject(error);
+      }
+    };
+  });
+}
```
PeterBaker0: This is reusable machinery - please move into migration framework (parent folder) under utils or similar

library/forms/lib/components/maps/migrations/migrations.ts
```diff
@@ -0,0 +1,47 @@
...
+
+// Keep the IndexedDB and migration versions in step whenever a new migration
+// is added so onupgradeneeded runs and the data migration can execute.
+export const TILE_DB_TARGET_VERSIONS: TileDbTargetVersions = {
+  // IndexedDB database version used to trigger onupgradeneeded.
+  databaseVersion: 2,
+
+  // Migration/data version expected by the current application.
+  targetMigrationVersion: 2,
+};
```
PeterBaker0: This needs more documentation/explanation - why do we have two identical version numbers we always need to increment. One presumably refers to indexed db's own version mechanism, vs ours? Do we need both?

library/forms/lib/components/maps/migrations/runner.ts
```diff
@@ -0,0 +1,207 @@
...
+
+    return migrations;
+  }
+
+  // Run validation for the data format introduced by the given version.
+  private async validateVersion(toVersion: number): Promise<void> {
+    if (toVersion === 2) {
+      await validateV2(this.transaction);
+    }
+  }
```
PeterBaker0: Don't like this - could very easily miss it - embed the validation functions/invariant checks as exported interface for a registered migration entry (TILE_DB_MIGRATIONS), then check against that rather than hardcoding individual validation functions.

library/forms/lib/components/maps/TileStore.ts
```diff
@@ -182,12 +173,40 @@ class MapTileDatabase {
...
+      error => {
+        console.error('[tiles_db] Database upgrade failed', error);
+        // Abort the whole upgrade so partial migration changes are rolled back.
+        try {
+          transaction.abort();
+        } catch {
+          // The transaction may already be aborting.
+        }
+      }
+    );
```
PeterBaker0: Have a fallback where you just flush and wipe it? If the migration fails... we need a path to normal behaviour. Show an alert "An error occurred while upgrading your offline maps. You may need to re-download your offline map regions." then flush/wipe the store? Need a fallback as this migration is still really a 'nice to have'.

PeterBaker0 (changes requested):
looking great - need some refactors to the migration framework but a fantastic start. More clarity on the two versions being managed and why we want both. And also clearer docs in the migration framework about lifecycle hooks into the indexed db open command (though I like this approach).

library/forms/lib/components/maps/tileStoreUtils.ts
```diff
@@ -0,0 +1,68 @@
...
+  expectedTileCount: number;
+  created: Date;
+  tileKeys: IDBValidKey[];
+  /** When set, this tile set is removed when the project is deactivated. */
+  projectId?: string;
+  /** Optional display label (defaults to setName in UI). */
+  label?: string;
+  /** Source offline map region this tile set was downloaded for. */
+  offlineMapRegion?: import('@faims3/data-model').OfflineMapRegion;
+}
```
PeterBaker0: So this should be the V2 types- right? The idea is that we have something like

tiles/types.ts

```text
<v0 type>
<v1 type extends /replaces v0 type>
<v2 type extends / replaces v1 type>
<export TileType = V2Type> 
```

Then in all the code you simply refer to TileType -> this means that you only have to change one single export line in the types, and all code refers/typechecks the new types.

PeterBaker0 (changes requested):
The type management is just a bit weird - but needs minimal changes - can re-review once those types are clean.

stevecassidy (approved):
Very nice work! Great to see some attention paid here and nicely dealing with db migration which is well overdue.  Thanks

PeterBaker0 (approved):
Great work - all good now 🚀

Issue #2278, opened 2026-08-28
Allow field and section conditions to reference parent and related record values


Caedous:
Follow-up to #2235 (parent record references) and #2256 (related record references). Those changes let TemplatedStringField and computed fields reference values from a parent record (`_PARENT.Field`) and from a record linked through a single-link Related Records field (`Link-Field.Field`). This issue extends the same references to conditional logic: the conditions controlling field and section visibility.

Motivating case (FAIMS/GroundSight-FAIMS#117): within a child record, show or hide fields/sections based on a value recorded on the parent (e.g. the grid square) or on the linked calibration record.

Proposed behaviour, consistent with the earlier work:

- A condition's field reference may be `_PARENT.Field` or `Link-Field.Field` in addition to a local field name, with the same eligibility rules (single-link Related Records fields only; parent references follow the existing ambiguity rule).
- Values resolve from the same record context used by templates and computed fields; conditions re-evaluate when a referenced link changes mid-edit.
- Conditions receive raw parent/related values, so every operator behaves on these references exactly as it does on a local field — including each operator's existing behaviour when the referenced value is missing (no parent, no linked record, or empty).
- Designer: condition editors offer parent and linked record fields in the field picker, with validation consistent with the computed field editor.

PR #2279, opened 2026-08-28
feat: Reference parent and related record values in field and section conditions

Caedous:
## Ticket

#2278

## Description

Follow-up to #2235 (parent references) and #2256 (related references), extending the same `_PARENT.Field` and `Link-Field.Field` references to the conditional logic controlling field and section visibility. A child record can now show or hide fields and sections based on a value from its parent record or from a record linked through a single-link Related Records field.

Conditions already evaluate against a flat values map with per-operator missing-value semantics, so the condition compiler is untouched: a new `buildConditionValues` helper merges raw parent and related record values (from the same `RecordContext` used by templates and computed fields) into that map under their reference keys. Every operator then behaves on a reference exactly as on a local field, including when the value is unavailable — no parent, no linked record, or an empty field.

## Proposed Changes

- **data-model**: `buildConditionValues` merges parent/related values into condition evaluation; `encodeRelatedRef` added beside the existing reference helpers
- **forms**: merged values wired into `EditableFormManager` (initial visibility, refresh after parent/related resolution and on mid-edit link changes, validation recompilation); `StaticFormProgress` and `DataView` accept an optional record context
- **app**: the read-only record view resolves parent/related values and passes them to the data view and progress bar
- **web (designer)**: the condition editor's field picker offers grouped *Parent › Field* and *Link Field › Field* entries via an `extraEntries` extension to the shared field search; references resolve to their underlying field definitions so operator narrowing and option-aware value inputs (e.g. a parent Select offering its own options) work unchanged; condition summaries render reference labels
- **docs**: Referencing Parent and Linked Record Values section on the conditions authoring page, with pointers from the Templated String, Computed Number, and Computed Text pages

## How to Test

Automated: 14 new data-model tests (merge semantics plus end-to-end operator evaluation over merged values), 3 new forms validation tests, 2 new field-search tests; all existing suites green.

Manual, on a notebook with a parent form (with a Select field), a child form, and a single-link Related Records field:

1. In the designer, add a condition to a child-form field — the field picker lists *Parent › …* and *Link Field › …* entries; selecting a parent Select narrows operators to equal/not-equal and offers its options in the value dropdown:

condition editor

2. In the app, create a parent with the matching value and add a child record — the conditioned field is visible immediately at mount; under a non-matching parent it is hidden.
3. Link/relink the related record mid-edit — the link-conditioned field appears and disappears without saving.
4. A record with no parent or link behaves as an empty local field would; a hidden required field does not block saving; the read-only View tab and progress bar match the edit view.

Note: the pre-existing `PlanTemplateManager` "flag off" test fails whenever `VITE_ENABLE_PLANS_IN_DESIGNER` is unset (default true) — it fails identically on clean `main`; the test should stub the flag rather than read the ambient environment.

## Additional Information

Cross-form dependency tracking (delete/rename warnings when a referenced field changes on the parent or linked form) does not cover these references — the same pre-existing gap as for `_PARENT`/linked references in templates and computed fields from #2235/#2256. A consolidated follow-up issue may be worth filing.

## Checklist

### DIFF
--- a/app/src/gui/pages/viewRecord.tsx
+++ b/app/src/gui/pages/viewRecord.tsx
@@ -25,6 +25,10 @@ import {
   RecordID,
   RevisionHistoryEntry,
   formatTimestamp,
+  getRecordContextFromRecord,
+  RecordContext,
+  resolveParentValues,
+  resolveRelatedValues,
 } from '@faims3/data-model';
 import {
   DataView,
@@ -215,6 +219,7 @@ const InfoTabContent: React.FC<InfoTabContentProps> = ({
 interface ViewTabContentProps {
   /** The notebook tab this page sits under, which its record links stay on. */
   notebook: RecordRouteNotebook;
+  recordId: RecordID;
   formData: NonNullable<
     Awaited<ReturnType<DataEngine['form']['getExistingFormData']>>
   >;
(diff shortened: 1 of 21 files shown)


stevecassidy (approved):
Looks good, quite straightforward after the previous changes.

Issue #2266, opened 2026-08-21
Make the notebook tab part of the route


jayenashar:
The notebook tab is a `?tab=` search param mirrored into `tabIndex` state in `app/src/gui/components/notebook/index.tsx`, so coming back from a record loses it.

Make the notebook route a layout route with `<Outlet/>` and the tab a path segment, with the record routes nested under an optional `:tab?` so existing record links still resolve and land on the default tab. Back from a record is then just `..`.

FAIMS3 keeps no fixed list of tabs: `NotebookView` reads `:tab` and passes `tab` plus `actions.setTab` down through `NotebookViewComponentProps`, and each view maps its own slugs, falling back to its own default on an unknown one.

Also use it to return to the right list where we have my + other records.

jayenashar:
Superseded by the 2026-09-03 design call. The tab is out of the route and held in context as of #2281 (e0462e86c), so the notebook path is `/notebooks/:serverId/:projectId/:planId?` and gains nothing as a plan grows screens.

The thing this asked for, keeping the tab when you come back from a record, is met by the provider instead of the URL.

*Drafted with AI assistance, reviewed before posting.*

PR #2281, opened 2026-08-28
Allow more than one plan per notebook

### LINKED ISSUE #2280: Allow more than one plan per notebook
A notebook carries a single `plan`, so a project can only offer one data collection workflow. A template should be able to define several plan templates, each instantiated with its own config when a notebook is created from it, and the app should let the user choose which workflow to work in.

Each plan needs a label for that choice, and the order they are declared in is the order offered.

Nothing in production carries a plan yet, so no migration is needed — existing templates can be updated by hand.

jayenashar:
## Ticket

#2280

## Description

A notebook carries a single `plan`, so a project can only offer one data collection workflow.

video: https://github.com/user-attachments/assets/75d512f9-39e1-4c5e-bc68-d7626ce8a3de

## Proposed Changes

- A template carries `planTemplates` and a notebook carries `plans`, each with a `planId` and a `label`. Both are required and both are distinct within the set, since the chooser has only the label to tell two plans apart by. Declared order is the order offered.
- A plan may also carry a `description`, optional, shown under its label on the chooser and on the plan's own screen.
- Creating a notebook instantiates every plan template from its own config: `planConfigs`, keyed by plan id, replaces `planConfig`.
- A notebook with several plans opens on a button per plan. Picking one enters that plan for the visit, with its own tabs inside it.
- The notebook route gains a segment for the plan, `/:planId?`, so a link says which plan it means. The tab is not in the route: a view's tab lives in a context of its own, so a lone trailing segment is unambiguously a plan, and a plan gains screens without the route gaining segments.
- Two providers mount on the notebook route, above the record screens nested under it. One names the notebook once, with the plan it stands on, the route back out of it, and the one move within it; the record list, overview map, add buttons, record pages and delete read that rather than the route's own params, and the relative parent route a record used to exit by is gone. The other holds the tab, a context for the notebook's own view and a context for a plan's, each holding a tab per notebook and per plan within it, so opening a record and coming back returns to the tab that was left and a plan can model screens or regions rather than reusing the notebook's tabs. A view takes `tab: {current, select}`. Control points for the record form itself are #2290.
- `uiSpecification.settings` carries an optional `planChooserMarkdown` beside `showQrCodeButton`, authored in the designer and shown as rich text above the plan buttons. Absent or blank, the chooser heads itself.
- The designer authors, labels and reorders several plan templates.

## How to Test

Create a notebook from `api/notebooks/two-plans.json` and open it in the app: it offers a choice of the two plans, and picking one opens that plan.

For the template path, create a template from `api/notebooks/two-plan-templates.json`, then create a notebook from it passing that file's `planConfigs`. The result is the notebook in `two-plans.json`, which is what `api/test/notebookPlanTemplate.test.ts` asserts.

## Additional Information

No migration, and none of it fails loudly. A template still written to the singular `planTemplate`, or a notebook still on `plan`, loses it: `TemplateDefinitionSchema` and `NotebookDefinitionSchema` are plain `z.object`, so the stale key is stripped and the plan is silently gone. That is fine while nothing in production carries a plan (#2280), but it is worth watching against #2269: if those definition schemas ever go passthrough, a stale `planTemplate` stops being dropped and rides along in the stored document instead.

Repeated plan ids, repeated plan labels, and a config for a plan the template does not have, are errors rather than a quietly dropped plan.

A record's `planReference` is qualified with the plan id that claims it, so two plans reusing a reference key, or collecting the same form, keep their own records. Nothing in production carries a plan, so no record carries a reference and there was nothing to migrate.

A plan's screen is scoped to the plan: it offers an add button per form the plan view names, and its record list and overview map show only the records that claim it. A plan is not held to one form. A notebook's other forms are reached from the plan that collects them.

A view's tab is held above the record screens nested under the notebook route, so it survives opening a record and coming back, which is what #2266 put it in the route for. It does not survive a reload, nor leaving the notebook and coming back, since the provider unmounts with the route. Both were the URL's doing. No link can open a tab from outside a view either; the attachment dialog's "Go to settings" was the only user of that, and now goes to the notebook. Untested: what Capacitor does when the app is unloaded from memory rather than force-closed. If it restores the URL, that argues for a query param rather than a path segment.

One thing that predates this branch and still holds: the counted plan's target counts the record list, which the record table's search box filters, so typing there can drop the count below the target, hide the target-reached notice and offer the add button again. The count has always read that same query-filtered list.

## Checklist


library/data-model/src/uiSpecification/normalize.ts
```diff
@@ -142,19 +146,27 @@ export function normalizeNotebookTemplateUiSpecification(
     label: 'template uiSpecification',
   });
 
-  if (
-    definition.planTemplate &&
-    !safeValidatePlanTemplate(definition.planTemplate).success
-  ) {
-    throw new Error('Invalid plan template in template uiSpecification');
+  const duplicateIds = findDuplicatePlanIds(definition.planTemplates);
```
jayenashar: The PR body says a template still on the singular `planTemplate` "fails to load with a message asking for it to be moved into `planTemplates`", and the commit *fail loudly where a plan could be lost or misconfigured* says the same. Nothing checks for it: `TemplateDefinitionSchema` and `NotebookDefinitionSchema` are plain `z.object`, so the key is stripped and the plan is silently gone (`legacyNotebook.ts` already leans on that strip). The behaviour is fine given #2280 says nothing in production carries a plan, but the body should describe it.

Worth watching against #2269: if the definition schemas ever go passthrough too, a stale `planTemplate` stops being dropped and rides along in the stored document instead.

library/data-model/src/plans/countedPlan.ts
```diff
@@ -51,6 +51,7 @@ export const instantiateCountedPlan = ({
     throw new Error('Invalid counted plan template');
   }
   return {
+    planId: template.planId,
```
jayenashar: `planId` is written twice: here and in `listOfRecordsPlan.ts:67`, then overwritten in `templates.ts:673`, which documents that it goes last on purpose. Drop it from both instantiate functions and leave the api the single owner, so a third plan type has nothing to remember.

app/src/gui/components/notebook/types.ts
```diff
@@ -76,8 +77,15 @@ interface ComponentProps {
...
-  // The tab slug from the URL, which a view resolves with `useResolveTab`
+  // The tab slug from the URL, which a view resolves with `useResolveTab`.
+  // Scoped to this view's plan: with several plans the URL segment also names
+  // the plan, and that prefix is stripped before it reaches the view.
   tab?: string;
+  // The plan instance this view is rendering. A notebook may carry several, so
+  // a view must read this rather than reaching into the project for `plan`.
+  plan?: RegisteredPlan;
+  // Addresses `plan` within the notebook, for routes and plan references
+  planId?: string;
```
jayenashar: Unused, and duplicates `plan.planId`. Drop it and the assignment at `notebookView.tsx:298`.

app/src/gui/components/notebook/notebookView.tsx
```diff
@@ -249,30 +251,58 @@ function NotebookViewWithSpec({
...
+  );
 
   // Completion roll-up per plan-claiming record, for its cell's status; only a
   // registered plan view can display it, so gate the walks on one
   const planRecordStatusReports = usePlanRecordStatusReports({
     projectId: project.projectId,
     uiSpecification,
     records: records.allRecords,
-    enabled: PlanComponent !== undefined,
+    enabled: planViews.length > 0,
```
jayenashar: With the chooser on screen there is no view to display the roll-up, so this walks every record for nothing each time a multi-plan notebook opens. `enabled: activePlan !== undefined` matches the comment above.

jayenashar (commented):
Reading my own PR the way Peter and Steve would: duplication, scope, consistency with existing patterns, observed behaviour.

One thing to settle before this leaves draft: the PR body and one commit message both promise a loud failure on the old single-plan shape that the code does not implement (first comment).

web/src/designer/state/planTemplates-reducer.ts
```diff
@@ -0,0 +1,103 @@
...
+    },
+    /**
+     * Move a plan template one place up or down. The array order is the order
+     * the app's plan chooser offers the plans in, so this is how a template
+     * author sets that order. Ids are minted on add, so moving keeps each
+     * plan's own.
+     */
+    planTemplateMoved: (
+      state,
+      action: PayloadAction<{index: number; delta: number}>
```
jayenashar: `{index, delta}` is a new convention here: `fieldMoved`, `sectionMoved` and `viewSetMoved` all take `direction: 'up' | 'down'` (or `'left' | 'right'`). `arrayMove` from `@dnd-kit/sortable` is already in the tree if the body should be shorter too.

library/data-model/src/plans/types.ts
```diff
@@ -4,21 +4,74 @@ import z from 'zod';
...
+     * the plan is authored, so reordering the list cannot re-address it.
+     */
+    planId: PlanIdSchema,
+    /** Names the plan on the notebook's plan chooser. */
+    label: z.string().optional(),
   })
   .passthrough();
 export type PlanTemplate = z.infer<typeof PlanTemplateSchema>;
 
+const AuthoredPlanTemplateSchema = PlanTemplateSchema.omit({planId: true});
```
jayenashar: Export this. `CountedPlanDialog.tsx:77` and `ListOfRecordsPlanDialog.tsx:117` each repeat `.omit({planId: true})` on their own schema, and a third plan type repeats it again.

library/data-model/src/plans/types.ts
```diff
@@ -4,21 +4,74 @@ import z from 'zod';
...
+export const PlanIdSchema = z
+  .string()
+  .min(1)
+  .regex(ROUTE_SAFE, 'A plan id may not contain . / \\ ? # or %');
+
+/** A plan type keys the registry, and `derivePlanId` mints plan ids from it. */
+export const PlanTypeSchema = z
+  .string()
+  .min(1)
+  .regex(ROUTE_SAFE, 'A plan type may not contain . / \\ ? # or %');
```
jayenashar: Same regex as `PlanIdSchema` above, with a message differing by one word. One schema, or derive one from the other.

library/data-model/src/plans/types.ts
```diff
@@ -4,21 +4,74 @@ import z from 'zod';
...
-    planType: z.string(),
+    planType: PlanTypeSchema,
+    /**
+     * Identifies this plan among the notebook's plans, and addresses it in
+     * routes and record references. Carried over from the plan template.
+     */
+    planId: PlanIdSchema,
+    /**
+     * Names the plan on the notebook's plan chooser. Carried over from the plan
+     * template at instantiation; falls back to the plan type's own label.
```
jayenashar: Says the label falls back to the plan type's own label; `getPlanLabel` falls back to `planId`.

app/src/gui/components/notebook/plans/PlanChooser.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const PlanChooser = ({
+  plans,
+  onSelect,
+}: {
+  plans: {plan: RegisteredPlan}[];
+  onSelect: (planId: string) => void;
+}) => (
+  <Box sx={{p: 3, maxWidth: 480, mx: 'auto'}} data-testid="plan-chooser">
+    <Typography variant="h6" component="h2" sx={{mb: 2}}>
+      Choose a workflow
```
jayenashar: The designer says "Add Plan", "Remove plan?", "Counted plan"; this says workflow. "Choose a plan" keeps one word for the thing the author authored.

app/src/constants/routes.tsx
```diff
@@ -64,6 +66,33 @@ export const useResolveTab = <T extends string>(
...
 };
 
+/**
+ * Separates a plan's id from that plan's own tab slug in the route's single tab
+ * segment, e.g. `lab-samples.details`. Written whenever a plan view is on
+ * screen, a single-plan notebook included, so a link always says which plan it
+ * means. A bare slug is what the default record list writes, and is still read
+ * as a single-plan notebook's only plan, so older links resolve.
+ */
+export const PLAN_TAB_SEPARATOR = '.';
```
jayenashar: Why one segment carrying two things rather than `/:planId?/:tab?`? The separator has cost three follow-ups already (the trailing `<planId>.`, the `useResolveTab` rewrite guard, and the bare-segment match at `resolvePlanViews.ts:44-49`), and it is what forces the new charset restriction on plan ids. A route param puts the plan in the router; the bare-segment ambiguity needs the same resolution either way, so that is not what the separator buys.

app/src/gui/components/notebook/plans/CountedPlanView.tsx
```diff
@@ -22,15 +22,12 @@ export const CountedPlanView = (props: NotebookViewComponentProps) => {
...
+  // props rather than being read back off the project.
+  if (props.plan?.planType !== COUNTED_PLAN_TYPE) {
     return <div>CountedPlanView: Not a counted plan</div>;
   }
-  // this really is a counted plan
-  const plan: CountedPlan = project.uiDefinition.plan as CountedPlan;
-  if (!plan) {
-    return <div>No plan defined for this notebook</div>;
-  }
+  const plan: CountedPlan = props.plan as CountedPlan;
```
jayenashar: `RegisteredPlan` is a discriminated union on `planType`, so the guard above narrows on its own. `tsc --noEmit` passes in `app` with this cast and the one at `ListOfRecordsPlanView.tsx:89` removed.

app/src/gui/components/notebook/plans/resolvePlanViews.test.ts
```diff
@@ -0,0 +1,208 @@
...
+  ({
+    planId: LIST_OF_RECORDS_PLAN_TYPE,
+    planType: LIST_OF_RECORDS_PLAN_TYPE,
+    ...extra,
+  }) as RegisteredPlan;
+
+/** Stands in for the app registry; returns a marker per plan type. */
+const getView = (planType: string) =>
+  planType === COUNTED_PLAN_TYPE || planType === LIST_OF_RECORDS_PLAN_TYPE
+    ? (`view:${planType}` as unknown as string)
```
jayenashar: Already a string; the double cast can go.

api/notebooks/two-plans.json
```diff
@@ -0,0 +1,166 @@
+{
```
jayenashar, line 1: This is a notebook definition (`uiSpecification.plans`), so How to Test goes through create-from-scratch and never touches `planTemplates` or `planConfigs`, which is most of the `templates.ts` diff. A companion template fixture carrying `planTemplates` plus the example `planConfigs` would let a reviewer run that path by hand.

library/data-model/src/uiSpecification/normalize.ts
jayenashar: No code change: nothing in production carries a plan, so the silent strip is the behaviour I want. Body now says that, and carries the #2269 note.

library/data-model/src/plans/countedPlan.ts
jayenashar: ✅ 7d75eb0fc. Both instantiate functions dropped it, and `instantiatePlan` now returns `Omit<..., 'planId'>` so a third plan type cannot write one either.

app/src/gui/components/notebook/types.ts
jayenashar: ✅ f58d390cb, with the assignment site.

app/src/gui/components/notebook/notebookView.tsx
jayenashar: ✅ f58d390cb. Now `enabled: activePlan !== undefined`.

web/src/designer/state/planTemplates-reducer.ts
jayenashar: ✅ cabae8d41. `direction: 'up' | 'down'`, matching `fieldMoved` and `sectionMoved`. Left the splice body alone rather than pull in `arrayMove`, since the other reducers hand-roll theirs too.

library/data-model/src/plans/types.ts
jayenashar: ✅ 064e794ef. Exported, plus an `authoredSchema` helper the two dialogs now call on their own schemas, which is where the repetition actually was.

library/data-model/src/plans/types.ts
jayenashar: ✅ 7e30ad11d. One `routeSafeSchema` helper builds both. The route no longer splits on a dot, so the rule dropped `.` at the same time, and 1801ad9e3 adds back a refusal of `.` and `..`, which the router resolves away.

library/data-model/src/plans/types.ts
jayenashar: ✅ a28831c05.

app/src/gui/components/notebook/plans/PlanChooser.tsx
jayenashar: ✅ ef2f189b9. "Choose a plan", and the designer's label helper text with it.

app/src/constants/routes.tsx
jayenashar: ✅ f58d390cb. The route is `/:planId?/:tab?` now, and `PLAN_TAB_SEPARATOR`, `splitPlanTab`, `joinPlanTab` and the trailing-dot handling are gone. React Router puts a lone trailing segment in `planId`, so the bare-segment case resolves the same way as before: one that names no plan is read as the tab. Two follow-ups a pre-push review caught: 2c61f0a86, so a lone segment naming a plan this build cannot render asks instead of opening the only renderable one, and 2eb6123ba, so the QR scan redirect carries the plan.

app/src/gui/components/notebook/plans/CountedPlanView.tsx
jayenashar: ✅ 2ec0f65e1, both of them.

app/src/gui/components/notebook/plans/resolvePlanViews.test.ts
jayenashar: ✅ f58d390cb, in the rewrite of that file for the new route params.

api/notebooks/two-plans.json
jayenashar: ✅ 04ac58eeb. `api/notebooks/two-plan-templates.json` is the same notebook on the template side, with its `planConfigs` alongside, and a test asserts creating from it lands on the plans in `two-plans.json`.

app/src/gui/components/notebook/notebookView.tsx
```diff
@@ -328,17 +361,34 @@ function NotebookViewWithSpec({
       records,
       recordStatus.data,
       planRecordStatusReports,
+      planTab,
```
jayenashar: `tab` at line 352 is no longer read: the body takes `planTab` now. Drop it from the list.

app/src/constants/routes.tsx
```diff
@@ -64,44 +66,57 @@ export const useResolveTab = <T extends string>(
...
   tab?: string;
 }) {
   return (
     INDIVIDUAL_NOTEBOOK_ROUTE +
     serverId +
     '/' +
     projectId +
-    (tab ? '/' + tab : '')
+    [planId, tab]
+      .filter(segment => segment)
```
jayenashar: The first two segments are concatenated, the last two go through `filter`/`map`/`join`. One pass over all four reads the same and loses the `map`:

```ts
return (
  INDIVIDUAL_NOTEBOOK_ROUTE +
  [serverId, projectId, planId, tab].filter(Boolean).join('/')
);
```

`routes.test.tsx` passes on that, the tab-takes-the-plan-segment case included.

web/src/designer/integration/legacyNotebook.ts
```diff
@@ -118,26 +118,26 @@ export function tryNormalizeApiUiSpecification(
...
+  const planTemplates = Array.isArray(rawPlanTemplates)
+    ? (rawPlanTemplates as PlanTemplate[])
+    : undefined;
+
+  // Warn rather than fail on an invalid plan template so the template stays editable
+  const invalid = (planTemplates ?? [])
+    .map(planTemplate => safeValidatePlanTemplate(planTemplate))
+    .filter(result => !result.success);
+  const planWarning = invalid.length
+    ? `${invalid.length} of this template's plans failed validation and may need to be re-created.`
```
jayenashar: The warning lost what it was for. It used to end `: ${planResult.error.message}`; now the author gets a count and nothing to act on. `normalize.ts:158` names the offending plan by id on the same failure, and the id exists here too.

app/src/gui/components/notebook/plans/PlanChooser.tsx
```diff
@@ -0,0 +1,37 @@
+import {RegisteredPlan, getPlanLabel} from '@faims3/data-model';
+import {Box, Button, Stack, Typography} from '@mui/material';
+
+/**
+ * The first screen of a notebook carrying more than one plan: one button per
+ * plan, in the order the notebook declares them. Choosing one enters it for the
```
jayenashar: "Enters it for the rest of the visit" is the whole of it: `notebook.tsx:95` sets `backLink` to the notebook list unconditionally, and no plan view carries a switcher, so picking the wrong plan means leaving the notebook and coming back in. Browser and hardware Back do return here, since the chooser is pushed, but nothing on screen says so.

app/src/gui/components/notebook/plans/PlanChooser.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * it are the plan's own.
+ */
+export const PlanChooser = ({
+  plans,
+  onSelect,
+}: {
+  plans: {plan: RegisteredPlan}[];
+  onSelect: (planId: string) => void;
+}) => (
+  <Box sx={{p: 3, maxWidth: 480, mx: 'auto'}} data-testid="plan-chooser">
```
jayenashar: `mx: 'auto'` does not take: the chooser renders hard left in the app, not centred under the notebook title.

web/src/designer/components/plans/PlanTemplateManager.tsx
```diff
@@ -74,85 +84,145 @@ export const PlanTemplateManager = () => {
...
-            label={
-              formMissing
-                ? `${definition?.label ?? planTemplate.planType} plan: form missing`
-                : `${definition?.label ?? planTemplate.planType} plan: ${formLabel}`
-            }
-            color={formMissing ? 'warning' : 'default'}
-            onClick={() => setEditorPlanType(planTemplate.planType as string)}
-            onDelete={() => setRemoveConfirmOpen(true)}
-            deleteIcon={
+      <Stack spacing={1} sx={{width: '100%'}}>
```
jayenashar: `width: '100%'` wraps this onto its own line inside the Forms toolbar row (`design-panel.tsx:402`), so the plan rows render under the **Forms** heading with no heading of their own. As one chip it read as an adornment; as a row per plan plus Add Plan it reads as part of Forms, where Sections and Fields each get their own heading and tooltip.

Same rows: reordering is bare up/down icon buttons, where the forms and sections two rows below use the panel's `← → Reorder`.

web/src/designer/components/plans/PlanTemplateManager.test.tsx
```diff
@@ -74,27 +78,27 @@ describe('PlanTemplateManager feature flag', () => {
   });
 
   test('shows Add Plan in template mode when the flag is on and there is no plan', () => {
```
jayenashar, line 100: Still only the flag, with zero or one plan. What changed is untested: Add Plan stays available once a plan exists (it used to disappear), and rows move, label and remove by index. `planTemplates-reducer.test.ts` covers the reducer well; the component has nothing.

web/src/designer/plans.tsx
```diff
@@ -75,6 +82,16 @@ export const registerDesignerPlanType = (
       `Designer plan type ${definition.planType} is already registered`
     );
   }
+
+  // Plan ids are minted from the plan type, so it has to survive a route
+  // segment, the same rule the data-model registry holds.
+  const validType = PlanTypeSchema.safeParse(definition.planType);
+  if (!validType.success) {
+    throw new Error(
+      `Designer plan type ${definition.planType} cannot be registered: ${validType.error.issues[0].message}`
```
jayenashar: Same guard as `registry.ts:72-79`, same schema, message differing by one word. The comment above already calls it a copy.

jayenashar (commented):
Round two, same lens: duplication, scope, consistency with existing patterns, observed behaviour. Nothing below is a correctness bug.

Checked first: typecheck clean across app, web and api, all four suites green, and a localdev run of the seeded two-plan notebook. The template path instantiates both plans from their own configs in declared order, the chooser and every route case behave, and a designer save keeps the existing plan ids while minting a fresh one for a third plan of a type already present.

app/src/gui/components/notebook/notebookView.tsx
jayenashar: Dropped.

app/src/constants/routes.tsx
jayenashar: Done, one filter/join over all four segments.

web/src/designer/integration/legacyNotebook.ts
jayenashar: The warning now names each failing plan by id and carries its validation error, alongside the count.

app/src/gui/components/notebook/plans/PlanChooser.tsx
jayenashar: A plan view shown where there was a choice now carries a Change plan control above it, back to the chooser.

app/src/gui/components/notebook/plans/PlanChooser.tsx
jayenashar: The parent Stack resets its children's margins, so mx never had a chance. Centred by a full-width flex box instead.

web/src/designer/components/plans/PlanTemplateManager.tsx
jayenashar: Plans are their own section above Forms now, with a heading and tooltip of their own and Add Plan on the heading row. Reordering picks up the panel's icon buttons plus Reorder caption; the rows stack vertically so the arrows stay up/down.

web/src/designer/components/plans/PlanTemplateManager.test.tsx
jayenashar: Added: Add Plan staying offered once a plan exists, and moving, labelling and removing by position, plus the ends of the list having nowhere to move.

web/src/designer/plans.tsx
jayenashar: Both registries call assertRegistrablePlanType now, exported from data-model.

app/src/gui/components/notebook/notebookView.tsx
```diff
@@ -328,17 +362,51 @@ function NotebookViewWithSpec({
...
+  if (activePlan) {
+    const {Component} = activePlan;
+    // Only the browser's Back returns to the chooser, so where there was a
+    // choice to make, say so on screen too.
+    return planViews.length > 1 ? (
+      <Stack spacing={1}>
+        <Button
+          size="small"
+          startIcon={<ArrowBackIcon />}
+          onClick={() => showPlanTab({})}
```
jayenashar: This replaces, so Change plan overwrites the plan view's entry with the chooser while the pushed chooser is still behind it. History becomes chooser, chooser: the next Back does nothing visible and it takes two to leave.

The push at line 378 is what buys that, and it stopped earning its keep once this control existed. Drop it and every move within a notebook replaces, plan and tab alike: Back always leaves the notebook, Change plan is how you get back to the chooser, and line 151 loses its exception.

web/src/designer/components/plans/PlanTemplateManager.tsx
```diff
@@ -74,85 +89,187 @@ export const PlanTemplateManager = () => {
...
+                    </span>
+                  </Tooltip>
+                  <Typography variant="caption" sx={designerControlLabelSx}>
+                    Reorder
+                  </Typography>
+                </Stack>
+                <TextField
+                  size="small"
+                  label="Label"
+                  placeholder={planTemplate.planId}
```
jayenashar: Label is optional and `getPlanLabel` falls back to the plan id, which `derivePlanId` mints from the plan type. So a template with two counted plans and neither labelled ships a chooser reading `Counted` and `Counted-2`, on the one screen whose whole job is telling them apart.

Nothing in production carries a plan, so make `label` required on `PlanTemplateSchema` and `PlanSchema` rather than falling back: `getPlanLabel` collapses to `plan.label`, the conditional spread at `templates.ts:670` goes, and this row and the plan dialogs gate save on it.

web/src/designer/integration/legacyNotebook.ts
```diff
@@ -118,26 +118,30 @@ export function tryNormalizeApiUiSpecification(
...
-      planWarning = `This template's plan failed validation and may need to be re-created: ${planResult.error.message}`;
-    }
-  }
+  const planTemplates = Array.isArray(rawPlanTemplates)
+    ? (rawPlanTemplates as PlanTemplate[])
+    : undefined;
+
+  // Warn rather than fail on an invalid plan template so the template stays
+  // editable; name each failing plan, as normalizing a template does
+  const invalid = (planTemplates ?? []).flatMap(planTemplate => {
```
jayenashar, line 140: This validates each plan template and warns, but never checks for a repeated plan id. `normalizeNotebookTemplateUiSpecification` does, and the api runs it on template create and update, so an uploaded template with repeated ids opens in the designer and is refused on save. The rows key on `planTemplate.planId`, so they collide too.

`findDuplicatePlanIds` is exported and takes exactly this shape.

library/data-model/src/uiSpecification/normalize.ts
jayenashar: This block and lines 180-193 are the same fourteen lines twice, differing by the field name, the label and which validator runs. One helper over `(items, validate, label)` collapses both, and the next definition kind gets the rule for free.

app/src/gui/components/notebook/plans/PlanChooser.tsx
```diff
@@ -0,0 +1,44 @@
...
+ * The first screen of a notebook carrying more than one plan: one button per
+ * plan, in the order the notebook declares them. Choosing one enters it for the
+ * rest of the visit, so the plan view owns the whole screen and any tabs inside
+ * it are the plan's own.
+ */
+export const PlanChooser = ({
+  plans,
+  onSelect,
+}: {
+  plans: {plan: RegisteredPlan}[];
```
jayenashar: The chooser never reads `Component`, so this is the resolver's shape leaking into a presentational component. `RegisteredPlan[]` says what it renders, `plans={planViews.map(({plan}) => plan)}` at the call site, and the test stops wrapping each plan in an object to satisfy a field it does not use.

jayenashar (commented):
Round three, same lens: duplication, scope, consistency with existing patterns, observed behaviour.

Checked first: no `plan`/`planTemplate`/`planConfig` leftovers anywhere in src, fixtures or docs, and the two fixtures' uiSpecs are identical apart from the plans they carry. The route change is safe against the record child routes: `records/:recordId` outranks `:planId/:tab` on static-segment score, so record links resolve as before.

library/data-model/src/uiSpecification/normalize.ts
```diff
@@ -130,8 +134,36 @@ function normalizeUiSpecificationBundle<
...
+  label,
+}: {
+  plans: {planId: string}[] | undefined;
+  validate: (plan: unknown) => {success: boolean};
+  what: string;
+  label: string;
+}): void {
+  const duplicateIds = findDuplicatePlanIds(plans);
+  if (duplicateIds.length) {
+    throw new Error(`Repeated plan id ${duplicateIds.join(', ')} in ${label}`);
```
jayenashar, line 156: Plan ids are the only thing kept distinct here, and that is right: two plans are meant to share a form. MapGrid collects a cell's records in the field and lab check-in edits those same records later, so separating plans by form would forbid the pairing this ticket exists to enable. Separation has to come from the record's plan claim instead, and two places do not have one:

- `ListOfRecordsPlanView.tsx:57` keys `existingRecords` on a bare `planReference` across every record, and `:68` finds by the same key, so two plans reusing a reference key claim each other's rows.
- `CountedPlanView.tsx:34` counts `records.allRecords.filter(record => record.type === plan.formType)`, every record of the form whichever plan made it. With `allowExtraRecords: false` another plan's records take the target and the Add buttons go.

Both close by qualifying `planReference` with the plan id and reading it back the same way.

Which the body should not defer. "That needs a data migration and is left for separate work" contradicts the paragraph above it and #2280: nothing in production carries a plan, so no record carries a `planReference` and there is nothing to migrate. The qualification is free today and stops being free the moment a plan ships.

Same line, smaller: `label` is required but not distinct, so two plans may both be "Survey" on the one screen whose job is telling them apart.

web/src/designer/components/plans/CountedPlanDialog.tsx
```diff
@@ -101,39 +109,58 @@ export const CountedPlanDialog = ({
       <DialogContent sx={{...designerDialogContentSx, pt: 4}}>
         <Box sx={{maxWidth: 740, width: '100%', mx: 'auto'}}>
           <SimpleFieldWrapper
-            heading="Form"
-            helperText={
-              alertMessage ||
-              'Records of this form count towards the plan. The number required is set when a notebook is created from this template.'
-            }
+            heading="Label"
```
jayenashar: The Label block is the same in both dialogs (`ListOfRecordsPlanDialog.tsx:149`): heading, helper text, `data-testid`, `setLabel(initialTemplate?.label ?? '')`, the `trim()` on save and the `!label.trim()` disabled gate. A third plan type writes it a third time.

`label` sits on the base `PlanTemplateSchema` now, so it is not per-type. One shared field, or author it on the manager row, which already owns order and removal.

Also `authoredSchema(countedPlanTemplateSchema)` rebuilds the schema on every save and is the same value each time.

app/src/gui/components/notebook/notebookView.tsx
```diff
@@ -144,16 +146,17 @@ function NotebookViewWithSpec({
...
 
-  const {tab} = useParams<{tab?: string}>();
+  const {planId, tab} = useParams<{planId?: string; tab?: string}>();
 
-  // Replace rather than push, so leaving a notebook costs one Back press
-  const setTab = useCallback(
-    (nextTab: string) => {
+  // Every move within a notebook replaces, plan and tab alike, so leaving one
+  // costs a single Back press. Change plan is the way back to the chooser.
+  const showPlanTab = useCallback(
```
jayenashar: The contract 8eb590a66 fixed has no test. `resolvePlanViews` covers sixteen cases and `PlanChooser` three; the glue that carried the bug has none, and `notebookView` has no test file. What is worth pinning: choosing a plan, changing tab and Change plan all replace, so one Back leaves the notebook rather than two.

library/data-model/src/uiSpecification/types.ts
```diff
@@ -355,29 +355,31 @@ export type CompiledNotebookUiSpec = z.infer<
 
 /*
  * A template is a notebook definition that will be used to instantiate many notebooks.
- * It has the same uiSpec and metadata as a notebook but includes an optional plan template
- * that will be used to instantiate a plan when a notebook is created from the template.
+ * It has the same uiSpec and metadata as a notebook but includes optional plan templates,
+ * one per workflow, each instantiated into a plan when a notebook is created from the template.
```
jayenashar: ef2f189b9 settled on "plan" over "workflow"; 359, 364, 381 and 393 read "one per workflow" and "one workflow each".

jayenashar (commented):
Round four, same lens: duplication, scope, consistency with existing patterns, observed behaviour.

Checked first: every round-three point lands as its commit says, so those threads are closed. One open question below, on the form a plan collects.

app/src/gui/components/notebook/notebookView.tsx
jayenashar: ✅ 8eb590a66. Every move within a notebook replaces now, plan and tab alike; the push and the exception at line 151 are gone, so one Back leaves the notebook.

web/src/designer/components/plans/PlanTemplateManager.tsx
jayenashar: ✅ 1a5873db3. `label` is required on `PlanTemplateSchema` and `PlanSchema`, `getPlanLabel` is gone, `templates.ts` writes it unconditionally, and both dialogs gate Save on it.

web/src/designer/integration/legacyNotebook.ts
jayenashar: ✅ 395f07388. `findDuplicatePlanIds` refuses the design before it opens, so the designer and the api turn away the same template.

library/data-model/src/uiSpecification/normalize.ts
jayenashar: ✅ 091642cd2. One `assertPlansAddressable` over both.

app/src/gui/components/notebook/plans/PlanChooser.tsx
jayenashar: ✅ 907f28ce3. `RegisteredPlan[]`, mapped at the call site, and the test no longer wraps each plan.

library/data-model/src/uiSpecification/normalize.ts
jayenashar: ✅ 9fc36a460, f075e3bc1. A record's `planReference` is qualified with the plan id that claims it, and read back the same way: the list view keys its planned entries on the qualified reference, and the counted view counts and lists only what it claims. The add-record buttons carry the claim, and only for the plan's own form, so a record of another form does not claim a plan that does not collect it. The per-record status reports walk only the plan on screen's records rather than every claiming record.

Body fixed: the deferral is gone, since nothing carries a plan and there was nothing to migrate.

✅ 9c067bd8a, 7e791644e for the label. Required and distinct now, refused where a repeated id is, and the dialogs gate Save on it. The designer opens a stored template with a repeated label rather than refusing it, unlike a repeated id: the label is what the plan dialog edits, so refusing would wedge the author out of the one place that could fix it. It warns which, and the api still refuses to store it until one is renamed.

web/src/designer/components/plans/CountedPlanDialog.tsx
jayenashar: ✅ d029a0f24. One `usePlanLabel` hook and `PlanLabelField` component, which both dialogs use: the reset on open, the trim and the disabled gate live there once, and a third plan type gets them for free. Authoring it on the manager row would have left a new plan with no label until the row was filled in, which the required-label rule forbids.

The authored schema is built once at module scope now, not per save.

app/src/gui/components/notebook/notebookView.tsx
jayenashar: ✅ 05a8de8e1. `notebookView.test.tsx` pins it: choosing a plan, showing a tab and Change plan each replace, so one Back leaves the notebook. It fails if the flag comes off. Record create and view still push on purpose, so the spec says moves within the notebook rather than all navigation.

library/data-model/src/uiSpecification/types.ts
jayenashar: ✅ 56e343bed. All four say plan now.

app/src/gui/components/notebook/plans/CountedPlanView.tsx
```diff
@@ -109,6 +115,7 @@ export const CountedPlanView = (props: NotebookViewComponentProps) => {
                 project={project}
                 recordLabel={recordLabel}
                 refreshList={actions.refreshRecordList}
+                planClaim={{planReference, formType: plan.formType}}
```
jayenashar: A plan's screen offers Add buttons for the other plan's forms, and the record they create is claimed by neither plan.

`AddRecordButtons` renders one button per visible form (`add_record_by_type.tsx:165`), and `planClaim` claims only this plan's own (`:78`). `two-plans.json` has `visible_types: ["Site", "Feature"]`, so Survey shows an Add Feature button. Feature is the other plan's form, so the record is unclaimed: filtered out of Survey's table by `:39-42`, and out of the Features plan's cards, which key on the qualified reference. On main the table was `records.allRecords`, so it was at least visible.

Hand `plan.formType` to `AddRecordButtons` and offer only the form the plan collects. The other form is reachable from the plan that collects it, where its records are claimed and shown.

app/src/gui/components/notebook/plans/ListOfRecordsPlanView.tsx
```diff
@@ -50,8 +50,8 @@ export const ListOfRecordsPlanView = (props: NotebookViewComponentProps) => {
   const unclaimedMayBeStale =
     !status.canReadAllRecords || status.isDownloadingRecords;
 
-  // work out which records we've already created from the plan
-  // by checking the planReference field in the record metadata
+  // work out which records we've already created from the plan by checking the
+  // planReference field in the record metadata, which names the claiming plan
   const existingRecords: Record<string, boolean> = useMemo(() => {
     const existing: Record<string, boolean> = {};
     records.allRecords.forEach(record => {
```
jayenashar: The claims here are qualified by plan id, but the All Records table at `:191` is still `records.allRecords`, so it lists the other plan's records alongside this plan's. The counted view was narrowed to what it claims in f075e3bc1; two plan views now answer the same question two ways.

Narrow this one the same way, so both scope by plan id.

jayenashar (commented):
Round five, same lens, scoped to what carrying more than one plan changes: duplication, scope, consistency with existing patterns, observed behaviour.

Checked first: every round-four point lands as its commit says, so those threads are closed, and all four suites are green (data-model 613, app notebook and routes 43, web designer 78, api plan template 18).

web/src/designer/components/plans/PlanTemplateManager.tsx
```diff
@@ -213,31 +325,34 @@ export const PlanTemplateManager = () => {
...
           uiSpec={uiSpec}
           initialTemplate={
-            planTemplate?.planType === editorPlanType ? planTemplate : undefined
+            editing?.index === undefined
+              ? undefined
+              : planTemplates[editing.index]
           }
-          onClose={() => setEditorPlanType(null)}
+          // Every other plan's label, so the one being edited keeps its own
+          takenLabels={planTemplates
```
jayenashar, line 336: The designer's repeated-label check and the api's disagree on what counts as the same label. This passes the stored labels raw; `usePlanLabel` compares them against `value.trim()` (`PlanLabelField.tsx:56`). `findDuplicatePlanLabels` already defines the comparison, trimmed on both sides, and that is what the api refuses the save on.

Build `takenLabels` through that helper rather than a second rule.

Same thread: `legacyNotebook.ts:128` reads the plan templates off `candidate` and casts, where `parsed.data.planTemplates` is the copy `TemplateDefinitionSchema` already trimmed. Parsing per branch drops the cast, the `Array.isArray` guard and the untrimmed labels together. The comment above it, that the ternary-selected schema's inferred type drops them, is from when only one schema carried the key.

app/src/gui/components/notebook/notebookView.tsx
```diff
@@ -328,17 +358,51 @@ function NotebookViewWithSpec({
...
+
   // delegate to the plan view component
-  if (PlanComponent) {
-    return <PlanComponent {...props} />;
+  if (activePlan) {
+    const {Component} = activePlan;
+    // Back leaves the notebook, so where there was a choice of plan to make,
+    // this is the way back to it.
+    return planViews.length > 1 ? (
+      <Stack spacing={1}>
```
jayenashar: This relays out the plan view, and only where the notebook carries more than one plan.

`Stack` with `spacing` emits `& > :not(style):not(style) {margin: 0}` plus its own gap (`@mui/system/Stack/createStack.js:114`). `TabContext` renders no element of its own, so the plan view's `TabList` and every `TabPanel` become this Stack's children and lose their margins to it. The same plan lays out differently depending on how many plans sit beside it.

A plain `Box` around the button and the view keeps the spacing the view's own. Same trap as the chooser's `mx: 'auto'`.

app/src/gui/components/notebook/plans/CountedPlanView.tsx
jayenashar: ✅ ea6e8425d. `AddRecordButtons` takes the plan's form off `planClaim` and offers that alone, so every record it creates carries the claim and the conditional deciding whether to write one is gone. Three follow-ups a pre-push review caught: b77ff3e95, since a plan's form is chosen from every form rather than the visible ones, so filtering `visible_types` left a plan on a hidden form with no button at all; cb24aca1f, so the one button reads as the form it creates rather than "Add new Record"; and cfe6c41df, since the overview map was still plotting every plan's records beside a list that no longer did.

app/src/gui/components/notebook/plans/ListOfRecordsPlanView.tsx
jayenashar: ✅ 5f15f1f87. The table reads the plan's claims the same way the counted view does. 4e664fb44 then put the predicate in one place, `recordsClaimedBy`, keyed on plan id alone: the claim is minted by the plan that made the record, so the counted view's extra form check only ever disagreed, and a plan type carrying no form of its own scopes the same way.

Two consequences of the narrowing, also in 4e664fb44 and cfe6c41df: the second tab is "Collected" rather than "All", since it no longer lists all of the form's records, and the overview map beside it is scoped too, so tapping a pin can no longer open a record the list says is not there.

web/src/designer/components/plans/PlanTemplateManager.tsx
jayenashar: ✅ ccb6e242c, de535917a. `planLabelKey` holds the comparison once in data-model: `findDuplicatePlanLabels` uses it and so does the dialog, on both sides, so an uploaded template's untrimmed label now reads as taken in the designer instead of only being refused by the api. The manager still hands over the stored labels, since the rule they are compared by is no longer its to state.

`legacyNotebook.ts` parses in each branch, so the plan templates come off what the template schema parsed: the cast, the `Array.isArray` guard the schema already enforced, and the stale comment go together. 467811183 drops the spread of the result over itself that this left behind.

app/src/gui/components/notebook/notebookView.tsx
jayenashar: ✅ 013c5c6d5, 4ff7feb39. The `Stack` is gone, but a `Box` in its place broke the same parity the other way: the page renders the notebook inside `<Stack spacing={2}>`, so with one plan the plan view's own children are that Stack's children and with two the wrapper was. A fragment leaves them under the page's Stack either way, and the button takes that Stack's gap rather than carrying a margin of its own.

app/src/gui/components/notebook/plans/CountedPlanView.tsx
```diff
@@ -53,12 +55,7 @@ export const CountedPlanView = (props: NotebookViewComponentProps) => {
 
   const showAddRecordButtons = status.isAllowedToAddRecords && !targetReached;
 
-  // recordLabel based on viewsets
-  const recordLabel =
-    uiSpecification.visible_types?.length === 1
-      ? uiSpecification.viewsets[uiSpecification.visible_types[0]]?.label ||
-        uiSpecification.visible_types[0]
-      : 'Record';
+  const recordLabel = planRecordLabel({uiSpecification, plan});
```
jayenashar, line 53: `plan.label` renders in one place, `PlanChooser.tsx:38`. Once a plan is entered nothing names it: `:64` heads the screen with `Counted Plan: Collect 4 Site records`, the tabs read `Planned Sites`, and the control above says only `Change plan`. Two plans on one form, which is the pairing the ticket exists to enable, then differ on screen by that number alone.

Lead with the label: `Survey: collect 4 Site records`, and the same at `ListOfRecordsPlanView.tsx:101`. That is the word the author chose and the chooser offered. `Counted Plan` and `List of Records Plan` are the plan type's name, which a user is shown nowhere else.

app/src/gui/components/notebook/notebookView.tsx
```diff
@@ -307,7 +339,15 @@ function NotebookViewWithSpec({
         OverviewMap: () => (
           <OverviewMap
             serverId={project.serverId}
-            records={records}
+            // Scoped to the plan on screen, so the map plots what its lists hold
```
jayenashar: The `OverviewMap` handed to a plan view is narrowed to `recordsClaimedBy`, but `props.records` beside it is not, and `types.ts:22` still says "All records in the notebook". One props object gives a view a scoped map and an unscoped list, so the plan-scoping decision is made here and re-made in each view.

Scope `records` here when `activePlan` is set and feed the map from the same value, so a view is handed one answer. `myRecords` and `otherRecords` are read only by the default notebook view.

jayenashar (commented):
Round seven, same lens, scoped to what carrying more than one plan changes: duplication, scope, consistency with existing patterns, observed behaviour.

Checked first: every round-six point lands as its commit says, so those threads are closed. Neither of the two below is a correctness bug.

app/src/gui/components/notebook/plans/CountedPlanView.tsx
jayenashar: ✅ 0f1162914. Both views lead with `plan.label` and name the records by the form's label rather than its id, matching the tabs below. The tab list's `aria-label` carries the label too, since `Counted Plan tabs` had the same defect for a screen reader.

app/src/gui/components/notebook/notebookView.tsx
jayenashar: ✅ d6134266c. `planRecords` is scoped once and feeds both the props and the map, so both plan views drop their own filtering and the scoping tests move up to `notebookView`'s spec, where the decision now lives.

Three follow-ups: 104f1afd2 drops `myRecords` and `otherRecords`, which nothing reads through these props, since the default view builds its own lists from its own hook; 8e04c33bc renames the prop `planRecords`, so the name carries the scoping rather than a comment above it; d965af65e holds the map and the list to the same records in the spec, and covers a claim whose reference qualifies the plan id.

`usePlanRecordStatusReports` still takes the unscoped list and its own `planId`, which I left alone: it needs the whole notebook to version by, since a report spans a record's child subtree and a child need not be claimed, and computing nothing without a plan is what its `planId === undefined` branch buys.

app/src/constants/routes.tsx
```diff
@@ -64,44 +66,51 @@ export const useResolveTab = <T extends string>(
...
+ * still resolve. A lone trailing segment always lands in `planId`, so the
+ * notebook views read one that names no plan as a bare tab slug.
+ */
+export const NOTEBOOK_ROUTE_PATH = `${INDIVIDUAL_NOTEBOOK_ROUTE}:serverId/:projectId/:planId?/:tab?`;
 export const EDIT_RECORD_ROUTE_PATH = `${EDIT_RECORD_SEGMENT}/:recordId`;
 export const VIEW_RECORD_ROUTE_PATH = `${VIEW_RECORD_SEGMENT}/:recordId`;
 
 /**
- * @returns /<notebook-plural>/<server>/<project>[/<tab>]
+ * @returns /<notebook-plural>/<server>/<project>[/<plan>][/<tab>]
```
stevecassidy: Is there an issue here with the optionality of both plan and tab, if you just see /notebooks/server/project-x/foo you don't know whether foo is a plan or a tab. I can't quite see where you are disambiguating these in parsing a route - does planId take precidence since it's first in the pattern? Can we be assured that we won't get into an ambiguous situation?

app/src/gui/components/notebook/plans/PlanChooser.tsx
```diff
@@ -0,0 +1,44 @@
...
+          <Button
+            key={plan.planId}
+            data-testid="plan-chooser-option"
+            variant="outlined"
+            size="large"
+            fullWidth
+            onClick={() => onSelect(plan.planId)}
+            sx={{justifyContent: 'flex-start', py: 2, textTransform: 'none'}}
+          >
+            {plan.label}
```
stevecassidy: Maybe we need the option for more text here than just the label.  Can we add a 'description' field in the plan as well and display it here, label = heading, description = content

stevecassidy (commented):
I've not had time for a complete review but left some notes. Will look deeper later today.

web/src/designer/components/plans/PlanTemplateManager.tsx
```diff
@@ -74,85 +88,183 @@ export const PlanTemplateManager = () => {
...
+                  sx={{fontWeight: 700, mb: 0.5, lineHeight: 1.35}}
+                >
+                  Plans guide data collection in the {config.notebookNamePlural}
+                  created from this template.
+                </Typography>
+                <Typography
+                  variant="caption"
+                  sx={{display: 'block', lineHeight: 1.45}}
+                >
+                  Each plan names the form it collects and what is expected of
```
stevecassidy: 'A plan defines a pre-defined way to collect data and may present a customised user interface for the {config.notebookName}.  If there is more than one plan, the user will be offered a choice of which to follow in the order listed here.

app/src/gui/components/notebook/add_record_by_type.tsx
```diff
@@ -40,15 +46,17 @@ export default function AddRecordButtons({
...
+  const {planId, tab} = useParams<{planId?: string; tab?: string}>();
   const uiSpec = compiledSpecService.getSpec(uiSpecificationId);
 
   if (uiSpec === undefined) {
     return <CircularProgress thickness={2} size={12} />;
   }
   const showQRButton = uiSpec.settings.showQrCodeButton;
   const viewsets = uiSpec.viewsets;
-  const visibleTypes = uiSpec.visible_types;
+  // A plan collects one form, so its screen offers that form alone. Taken from
```
stevecassidy: We don't restrict plans to only collect one form - in fact I'd expect there will be plans that operate on a group of forms (calibration plan).

app/src/gui/components/notebook/add_record_by_type.tsx
```diff
@@ -24,12 +24,18 @@ type AddRecordButtonsProps = {
   project: Project;
   recordLabel: string;
   refreshList: () => void;
+  /**
+   * Restricts these buttons to the plan's own form and claims for the plan
```
stevecassidy: This isn't quite what this is doing - if planClaim is passed in it is because this was called from a plan view and we want to assert the planReference on the created record.  Since this component is creating buttons, we should really pass in the list of types it should make buttons for.  From the default view this will be visible_types but from a plan it could be some other list.  A plan might control a number of form types.

app/src/gui/components/notebook/plans/PlanChooser.tsx
```diff
@@ -0,0 +1,44 @@
...
+}) => (
+  // Centred by a full-width flex parent: a Stack resets its children's
+  // margins, so `mx: 'auto'` on the panel itself does not survive.
+  <Box
+    sx={{p: 3, display: 'flex', justifyContent: 'center'}}
+    data-testid="plan-chooser"
+  >
+    <Box sx={{width: '100%', maxWidth: 480}}>
+      <Typography variant="h6" component="h2" sx={{mb: 2}}>
+        Choose a plan
```
stevecassidy: Hmm, I think I'll want to customise this text. Not sure where we can put the configuration, maybe in the notebook settings? It should be markdown for flexibility.

stevecassidy (changes requested):
Happy with the implementation, just a few tweaks to the wording in a few places plus a note on visible_types.

app/src/constants/routes.tsx
jayenashar: ✅ ccfc11f83. `planId` takes a lone segment wherever the notebook declares a plan by that name, so the one ambiguous input, a plan id that also reads as a tab slug, opens the plan; the doc states it and two tests pin it.

web/src/designer/components/plans/PlanTemplateManager.tsx
jayenashar: ✅ f3c1f60bb, your wording, with the ordering note on its own line.

app/src/gui/components/notebook/add_record_by_type.tsx
jayenashar: ✅ 561c364eb. `AddRecordButtons` takes `formTypes` and `planReference`, and the `is_visible` check moved to the notebook call site with it (12380c1b9).

app/src/gui/components/notebook/add_record_by_type.tsx
jayenashar: ✅ 561c364eb. Nothing reads a plan's `formType` to pick the buttons now, so a plan over a group of forms offers one per form and claims every record they create.

app/src/gui/components/notebook/plans/PlanChooser.tsx
jayenashar: ✅ 7cd890960, optional on both plan schemas, label heading over description content, plain text since the option is a button. 12380c1b9 shows it on the plan's own screen too, since a one-plan notebook never reaches the chooser.

app/src/gui/components/notebook/plans/PlanChooser.tsx
jayenashar: Not settled here. It could live in `uiSpecification.settings` beside `showQrCodeButton`, or in the notebook metadata the Details tab already renders as markdown; @stevecassidy which do you want?

PeterBaker0 (changes requested):
Supportive of the feature but concerned about the amount of URL hacking.

Adding more optional overrides onto the route gets ambiguous, and this will keep growing as plans/tabs start to need more state. E.g. your client workflows might need internal workflow state, say, a `workflowStepId`. 

The constraint is that we open forms by routing to them, so context can only travel three ways: the URL, location state, or shared state like React context/redux/other.

I would propose we utilise a plain React context provider to pass state down. Where we mount the provider is an interesting question.

For example; make an extensible context with a baseline of `baseRoute`, `tab`, `plan`. Plan and tab views own that state (`setTab`, `setPlan`). 

`EditableFormManager` consumes it to build exit and navigation links instead of reconstructing them from URL fragments.

Critical state that needs to persist through reload or hard nav gets encoded in the URL as query strings. 

Plain React context is enough I think.

We will reach a limit with how smartly EditableFormManager can build the navigation event functions. We may need to provide overrides later. This context will be a mechanism to pass those in, with type safety and avoiding serialization challenges in the URL. E.g. functions/other behaviour. What if your workflow needs to tell the form to not worry about errors? Or have special behaviours? This context will provide a way to pass that in and we can incrementally expose control points for override from form manager.

app/src/gui/components/notebook/plans/PlanChooser.tsx
stevecassidy: Description isn't quite the right thing and we might want to use that for another purpose. Add a new property in settings alongside showQRCodeButton.

jayenashar:
✅ e7b0a83ba. a provider on the notebook route now carries the notebook, the way back out of it, and the one move within it; the record list, overview map, add buttons, record pages and delete read that instead of the route params, and the relative parent route records exited by is gone.

- mounted on the notebook route, so the record screens under it share one answer
- one absolute setter rather than setPlan/setTab: a lone trailing segment reads as either a plan or a tab, so only the notebook view, which knows the plans, can say which the pair is
- plan and tab stay in the path so a link still names its plan; state a link need not carry hangs off the provider instead of the route
- form manager control points (navigation overrides, per-workflow form behaviour) in #2290

app/src/gui/components/notebook/plans/PlanChooser.tsx
jayenashar: ✅ 3006dcd2f. `settings.planChooserMarkdown` beside `showQrCodeButton`, authored in the designer's info panel with the same markdown editor as the design purpose and rendered as rich text above the plan buttons; both two-plan fixtures carry one.

jayenashar:
@PeterBaker0 the tab is a path segment because of #2266: it was `?tab=` and coming back from a record lost it.  does `planId` stay in the path beside it, or move to a search param?

jayenashar:
> @PeterBaker0 the tab is a path segment because of #2266: it was `?tab=` and coming back from a record lost it. does `planId` stay in the path beside it, or move to a search param?

or a third option: make the second segment required, `/<server>/<project>/:notebookTabOrPlanId/:planTab?`, so the bare notebook path replaces to the default and there's one optional segment instead of two. it doesn't remove the plan-vs-tab tie-break, just names it.

app/src/constants/routes.tsx
jayenashar: ✅ e0462e86c. The ambiguity is gone rather than disambiguated: the tab is not in the route any more, so a lone trailing segment is always the plan. `resolvePlanViews` lost the bare-tab reading with it. A segment naming no plan the notebook declares is a spent link, and opens the notebook as though it named none, which is what an old tab-bearing URL now is.

jayenashar:
✅ e0462e86c, following a call with steve. The tabs are out of the URL and into context.

- The route is `/<notebooks>/:serverId/:projectId/:planId?` with the record screens under it. Nothing else goes in, so a plan gaining screens does not grow it.
- The provider mounts on the notebook route, above the record screens nested under it. That is what makes the tab survive opening a record and coming back: a provider inside the notebook view would not, since the view unmounts when the record child route renders, and returning to the tab is what #2266 added the segment for.
- A context each for the notebook's own view and for a plan's, so a plan can model screens or regions rather than reusing the notebook's tabs. Each holds a tab per notebook, and per plan within it, because the app bar links straight from one notebook to the next without unmounting anything.
- Views take `tab: {current, select}` in place of `tab` alongside `actions.setTab`.
- `notebookRoute.tsx` already carries the base route and the plan, and the record pages build their exit links from it. Control points for the form itself are #2290.

Two things it gives up. A reload forgets the tab, and so does leaving the notebook and coming back, since the provider unmounts with the route; both were the URL's doing before. And no link can open a tab from outside a view any more, which the attachment dialog's "Go to settings" was the only user of. Would you rather that stayed addressable as `?tab=`, which the context could seed from, or is a view's tab private to the view?

Untested: what Capacitor does when the app is unloaded from memory rather than force-closed. If it restores the URL, that argues for a query param rather than a segment.

stevecassidy (commented):
Thanks, I'm happy with this updated version now. I'll let @PeterBaker0 do a final review and approve.

app/src/gui/components/notebook/plans/planViewRecords.ts
```diff
@@ -0,0 +1,34 @@
...
+export const recordsClaimedBy = ({
+  records,
+  planId,
+}: {
+  records: MinimalRecordMetadata[];
+  planId: string;
+}): MinimalRecordMetadata[] =>
+  records.filter(record =>
+    claimsPlan({planReference: record.planReference, planId})
+  );
```
PeterBaker0, line 22: This feels too opinionated - and the surrounding machinery as well. The plans only ever receive records that they claim reference over. I think many plans would require context of records from other plans (e.g. your workflows stuff would probably cross-reference). So either

a) the plan view interface should pass through the full record list and the planRecord list separately
b) the plan view should only ever pass through the full record list, and plan view implementations which really want to concern themselves with claims can do that filtering themselves.

Probably a) is most simple to reuse your wiring. 

Also how does this interact with the overview map - I think this should be the custom plan view's responsibility to manage that - i.e. whether it shows all records, planRecords or some combination.

PeterBaker0 (changes requested):
Have looked through. I think the record context providers will pave the way to hopefully enabling suitable storage through the navigation events. Right now it is likely a bit heavy weight just to be doing some tab management, but do prefer it over the route URL hacking with the chained optional params. 

In the future I suspect plans will need to store custom state. They can do it in the component, in the URL or in context. Redux is not a nice target for extension, typically. 

Unless they can hyjack the mount tree somehow, they will only have their outlet to put providers in, so it would be hard for them to persist this state, that means they will need to add something into the parent(s) provider. So this API is probably not ready for that yet, presumably you'd want to add some open-ended storage to the mount stable provider e.g. NotebookViewTabProvider. Some open ended state e.g. customPlanState?: any; or similar. 

Not sure yet - this seems to be a bit of a tricky problem to keep this clean in the context of plugins. 

I have left a request changes for the way that records are filtered aggressively for plan views - this seems very opinionated and limiting for plan view providers - they may need to have visibility of all records, but only show relevant ones, or inform plan relevant records by context of others.

app/src/gui/components/notebook/notebookView.tsx
```diff
@@ -290,9 +308,7 @@ function NotebookViewWithSpec({
         isDownloadingRecords,
       },
       records: {
-        allRecords: records.allRecords,
-        myRecords: records.myRecords,
-        otherRecords: records.otherRecords,
+        planRecords,
```
stevecassidy, line 321: I think losing allRecords/myRecords/otherRecords here is significant - if we're going to pass in planRecords that seems ok but it should be alongside the others which our default view will need.

app/src/gui/components/notebook/plans/planViewRecords.ts
stevecassidy: I think a) would work best, it means the view component doesn't need to re-filter if it wants those records but leaves the other filtered lists as well.

stevecassidy (commented):
Added to comments on passing records

app/src/gui/components/notebook/plans/planViewRecords.ts
jayenashar: ✅ ffdf6234d, option (a). Props carry `notebookRecords`, the whole notebook, beside `planRecords`. A claim names the plan that *made* a record, so a plan that processes rather than collects is claimed nothing, and one that cross-references another's records reads them here. `planRecords` and the claim rule are unchanged, so a view wanting only its own still gets one answer without re-filtering.

On the map: `components.OverviewMap` now takes an optional `records`, defaulting to `planRecords`, so a view showing more than its own plan's records can hold the map to the same set.

Found this the hard way building on top of it: a plan that logs samples the field workflow collected was handed an empty list, and the view was dead.

app/src/gui/components/notebook/notebookView.tsx
jayenashar: `notebookRecords` is on the props now (ffdf6234d), so the whole notebook is there; `myRecords` and `otherRecords` are a `createdBy` filter over it.

❓Left those two off, because the default view does not read these props: `notebookView.tsx:394` renders `<NotebookComponent project={project} tab={tab} />` and builds its own lists from its own hook. When it is ported to this interface, per the TODO just above that line, happy to add what it actually needs then. Say the word if you would rather they went on now.

stevecassidy: Tom has just ported NotebookComponent to the new interface so it will use those props now (see #2289) waiting on this PR to merge that work.

app/src/gui/components/notebook/types.ts
```diff
@@ -38,6 +38,13 @@ interface RecordListProps {
   // The records the plan on screen claims, scoped before they reach a view, so
   // a view and the components it is handed answer alike.
   planRecords: MinimalRecordMetadata[];
+  // Every record in the notebook, whichever plan claims it. A claim names the
```
stevecassidy: I don't think this comment needs to talk about plans, it's just all the records.

app/src/gui/components/notebook/notebookView.tsx
jayenashar: ✅ 45af8c9ff. `myRecords` and `otherRecords` are back on the props beside `notebookRecords`, so #2289 has them when this merges. Thanks for the pointer, my reasoning was out of date the moment Tom's port landed.

app/src/gui/components/notebook/types.ts
jayenashar: ✅ 45af8c9ff. Cut to "Every record in the notebook."

app/src/gui/components/notebook/plans/planViewRecords.ts
jayenashar: Two more on this, both found building on top of the props:

`notebookRecords` is the notebook's records **matching the current search**, not all of them, so the comment on it said more than the prop delivers. Corrected in d9eb268eb.

That mattered, because the search survived a plan change. It filters the whole notebook and changing plan does not remount the view, so searching in one plan thinned what the next one was handed, down to a map grid drawing claimed cells as unclaimed. f935cac03 clears it while rendering the new plan, with a test that fails without the fix.

app/src/gui/components/notebook/notebookView.tsx
```diff
@@ -239,40 +237,70 @@ function NotebookViewWithSpec({
...
+  // than tabs leaves the tabs the default view carries alone.
+  const notebookTab = useNotebookTab();
+  const planTab = usePlanTab();
+  const tab = activePlan ? planTab : notebookTab;
 
-  // Completion roll-up per plan-claiming record, for its cell's status; only a
-  // registered plan view can display it, so gate the walks on one
+  // Completion roll-up per record the plan on screen claims, for its cell's
+  // status; only that plan's view can display it, so the walks stop at its own
   const planRecordStatusReports = usePlanRecordStatusReports({
```
stevecassidy, line 272: I can see the motivation for putting this here but I don't think we have a solid idea yet whether all plan types will want this information. If they don't care about it, this computes it anyway which could be costly.  I'd say we let the plan view do this if it is needed rather than having it out here.

stevecassidy (commented):
I just had one small comment, happy with the reverted view API now but suggest taking out the plan status reports.

app/src/gui/components/notebook/notebookView.tsx
jayenashar: `usePlanRecordStatusReports` is already gated on a view being registered (`enabled: PlanComponent !== undefined`), so a notebook without one does no walks. What it cannot see is whether the view that is registered wants the reports.

Moving the call into the view needs the hook reachable from outside `app/src/utils/`. It runs on the Status tab's query keys, so a report either surface computes serves the other, and a view doing its own walk gives that up.

Export it beside the view API for a view to call, or leave each view to its own walk?

stevecassidy: I'd say let the view take care of it if needed, if we find it generally useful we can promote it to here.

PeterBaker0 (approved):
Happy after changes addressing Steve's comment re: plan status walks. Lots of work in this and appreciate it. 

I hope that this is the right play - it's getting a bit frisky/hairy in terms of the patterns/use cases and less clear how this maps into the open source use cases. This increases code complexity, so hope its the right play longer term.

stevecassidy (approved):

PR #2282, opened 2026-08-31
fix: control-center-header-overflow

xinyuhoulucy:
# fix: control-center-header-overflow

## Description

Fix the web logo title overflow.

image

Should look like:

image


PeterBaker0 (approved):
Saw on your display how it looks for different sizes, works well. Thank you.

PR #2283, opened 2026-09-01
feat: filter by timestamps, list hydrated records, hydration optimisations

PeterBaker0:
- Add exclusive updatedAfter / updatedBefore filters on record list, hydrated list, legacy dump, and all export formats
- Builds a new index on records in the DataDB by last-updated so time-window queries don’t scan the whole notebook
- Add paginated GET …/records/hydrated with bulk hydration so clients can poll field values in one shot rather than having to do a metadata list then fan out a bunch of individual hydration requests
- Export UI: optional From/To range (inclusive local minutes → exclusive API bounds)
- Finish/nav now flushes pending edits and stamps updatedAt so the timestamp stays current
- Docs, API tests, and export e2e coverage for the new filters and hydrated polling pattern

The below is a prototype dashboard powered by this new wiring - it can hydrate 3000 spatial records very rapidly, with batches of 100 on the hydration endpoint. As you proceed through surveying, strictly only changed data is pulled, making it scalable for a 'real-time dashboard' use case with thousands of existing records. 

image


stevecassidy (approved):
Looks good, a nice refinement. Thanks.

PR #2286, opened 2026-09-02
Build/update android build

stevecassidy:
# Build/update android minimum API level

## Ticket

None

## Description

We can't upload new Android releases as the minimum required API level is now 36.

## Proposed Changes

- Updates the miminum target API to 36.
- Updates Gradle version to 8.14.5
- Update fastlane version in both android and ios
- don't run an automatic fastlane update in CI
- removes a few old files that are not needed

## How to Test

Tested with nightly android build and the resulting APK is showing as having the correct API version.  Also tested IOS build which was having the same Fastlane problem, now fixed.

## Checklist


ranisa-gupta16 (approved):
Looks good steve, thankyou for your efforts in fixing this build.. :-)

PR #2287, opened 2026-09-02
feat: any of or none of selectors for single select

PeterBaker0:
Adding is one of, is not one of operators for single select fields. This allows you to claim

"The user selected any of the following options" Or "The user did not select any of the following options"

### DIFF
--- a/library/data-model/src/uiSpecification/conditionals.ts
+++ b/library/data-model/src/uiSpecification/conditionals.ts
@@ -222,10 +222,8 @@ registerCompiler('is-one-of', (expression: ConditionalExpression) => {
     const present = values[expression.field];
     if (typeof present !== 'string') return false;
 
-    const sanitized = sanitizeComparisonInput(present);
-    return targets.some(
-      target => sanitizeComparisonInput(target) === sanitized
-    );
+    // Strict match, same as equal — no trimming.
+    return targets.some(target => target === present);
   };
 });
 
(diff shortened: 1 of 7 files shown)


library/data-model/src/uiSpecification/conditionals.ts
```diff
@@ -206,6 +206,53 @@ const sanitizeComparisonInput = (input: string): string => {
 export const isStringArray = (value: unknown): value is string[] =>
   Array.isArray(value) && value.every(item => typeof item === 'string');
 
+/**
+ * True when the scalar field value is any of the listed target strings.
+ * Missing field, non-string value, or an empty/invalid target list → false.
+ */
+registerCompiler('is-one-of', (expression: ConditionalExpression) => {
```
Caedous, line 213: `equal` uses strict `===` with no trimming, but `is-one-of` trims both sides. For example `'Igneous '` fails `equal: 'Igneous'` but passes `is-one-of: ['Igneous']`. The list operators also trim so just wanted to make sure that the difference from `equal` is on purpose.

Caedous (approved):
Looks good, one minor question but other than that good to go!

library/data-model/src/uiSpecification/conditionals.ts
PeterBaker0: I'll make it consistent since it is a controlled vocab on rhs.

PR #2288, opened 2026-09-02
feat: Reference notebook metadata values in templates, computed fields and conditions

### LINKED ISSUE #2284: Allow templated strings, computed fields and conditions to reference notebook metadata values
Follow-up to #2235 (parent record references), #2256 (related record references) and #2278 (both in conditions). Those changes let templated strings, computed fields and field/section conditions read values from outside the current record. This issue adds a third source: the notebook's own metadata, as set in the designer's Info panel.

Notebook metadata is a natural place for settings that apply across every record — a survey season, a site code, the set of procedures a project is using. Being able to reference those in a condition lets one notebook design adapt its forms to per-project settings without editing the forms themselves; the same references in templates and computed fields allow, for example, a season code in every record's identifier. (Related: FAIMS/GroundSight-FAIMS#118, private.)

Caedous:
## Ticket
Closes #2284.

## Description
Lets templated strings, computed fields (ComputedNumber/ComputedText) and field/section conditions reference values from the notebook's custom metadata (the custom key/value fields on the designer's Info panel), completing the reference family from #2235 (parent), #2256 (related) and #2278 (both in conditions). Templates use `{{_METADATA.key}}`, expressions `{_METADATA.key}`, and conditions take `_METADATA.key` as a field reference, with the keys offered in each editor's picker as "Notebook › key". Design in the issue; GroundSight motivation in FAIMS/GroundSight-FAIMS#118 (private).

Metadata turned out to be the simplest member of the family, and two design points are worth calling out. First, metadata keys are not part of the uiSpec, so the compile pass types any `_METADATA.` reference as string by prefix (every metadata value is a string) rather than validating keys — the designer checks key existence against its own metadata state and reports a targeted error, and at run time an undefined key reads as blank, matching the empty-local-field contract. Second, metadata is notebook-wide and static while a record is open, so there is no per-record resolution and no mid-edit re-resolution: the app passes the notebook's custom metadata into `RecordContext` at the two existing seams (`EditableFormManager` via a new prop, and the record view's context query) and the existing evaluation paths pick it up.

Removing a custom key that is referenced is blocked in the Info panel via the existing `findFieldDependencyReferences` scan against the encoded reference, mirroring how field deletion is blocked. A pre-existing selector-stability warning in the designer (inline `?? {}` for optional custom metadata) is fixed with a shared `selectCustomMetadata` selector.

## Commits
- feat(data-model): `_METADATA.<key>` references in templates, expressions and conditions — new `metadataReferences.ts` (prefix/encode/decode), `RefType` 'METADATA', reserved-prefix guard, `RecordContext.metadataValues`, merges in `conditionValues`/`templatedFields`/`computedFields`, and a test suite mirroring `parentReferences.test.ts`.
- feat(forms,app): supply notebook metadata to record context — `metadataValues` prop on `EditableFormManager`; both app pages pass `project.uiDefinition.metadata.custom`.
- feat(web): notebook metadata in the condition picker (grouped under "Notebook metadata"), the templated string Visual Builder, and the computed editor (insert picker + undefined-key error); Info panel removal blocking; `selectCustomMetadata`.
- docs: new "Referencing Notebook Metadata Values" sections on the conditions, templated string and both computed field pages.

## Testing
Automated: data-model 41 suites / 625 passing (13 new metadata tests covering encoding, templates, computed, conditions, and compile errors including the reserved-prefix and bare-prefix cases); forms 54; web green; api 32 files / 513 passing. The pre-existing `PlanTemplateManager` "flag off" failure reproduces identically on clean `main` (`VITE_ENABLE_PLANS_IN_DESIGNER` unset defaults true; noted on #2279 as well).

Manual pass on a Calibration/Test/Session notebook: pickers and Visual Builder list custom keys; the three designer error paths (undefined key, text in arithmetic, bare prefix) report correctly; removal of a referenced key is blocked naming the referring condition and expression, and succeeds once unreferenced; conditions and derived values behave correctly in the app including the staleness contract (stored templated value unchanged until next save; computed value and visibility update on open after a metadata change); read-only view and progress bar match; parent and related references from the earlier tickets unaffected.

removal blocking

condition editor showing metadata

## Additional Information
The fixed `information` metadata fields (name, version, description) are deliberately not referenceable — their internal names differ between the legacy and NotebookDefinition shapes, so exposing them would bake unstable names into notebooks; easy to add later if wanted. As with `_PARENT`/related references, cross-form dependency tracking aside, the designer's removal blocking covers metadata keys but there is no automated test for that path — the manual pass covers it; happy to add a `conditionReferences`-style test if preferred.

## Checklist


PeterBaker0 (changes requested):
Do now:
- Need sanitization or validation of metadata custom keys if they are used in references, as can break syntax/function e.g. 
* A key with a . works in conditions and expressions (_METADATA.foo.bar is one symbol) and fails in templates ({{_METADATA.foo.bar}} is Mustache nested lookup).
* Spaces or {} break {…} / {{…}} syntax.

- It is hard to tell when you remove a custom key from metadata that the error is referring to field IDs/values - no visual distinction between normal text and field references. Give more information e.g. form/section/field - also note there could be more than one.

Follow up (or fix now if you like)
 - follow up tickets to improve the custom metadata UX in designer (it's pretty plain/janky) and docs (e.g. docs/user/authoring/structure.md) - they are kind of out of date referring to this feature mostly being deprecated
- Follow up ticket to pass in the metadata into the designer preview context - right now doesn’t function in preview
- Docs are out of date in docs/user/field-types/shared-settings/conditions.md - mention not being able to refer to parents etc

Caedous:
All review items addressed on the branch, including both items originally slated as follow-ups. Six commits:

**Stale condition functions (the "caching" observation): fixed in a7719bcf.** Diagnosis first. The refresh path did its job: it fetched the new uiDefinition, dispatched `updateProjectDetails`, and recompiled into `compiledSpecService`. But components look the compiled spec up by an ID built from serverId + projectId, which never changes. Nothing they watch in Redux changed, so none of them re-rendered onto the new compilation, and the memos and query caches keyed on the stable ID kept serving work computed from the old spec object. Navigating out and back didn't help because React keeps the tree mounted. A browser reload remounts everything, which is why that was the one thing that picked it up.

Fix per Steve's suggestion: `buildCompiledSpecId` now includes an FNV-1a hash of a key-sorted serialisation of the uiSpec. A changed spec gets a new `uiSpecificationId`, and every consumer selecting that field re-renders and looks up the new compilation. An unchanged spec hashes identically, so a no-op refresh triggers nothing. `updateProjectDetails` removes the superseded cache entry, which also migrates legacy non-hashed IDs from persisted state on first refresh. Re-verified with the repro that failed before: condition edited from A to A OR B in the designer, plain Refresh in the app, and both new and existing records evaluate the new condition without a reload. The same mechanism covers the wider class raised in the thread: added and deleted fields, changed templates, and changed expressions now propagate on refresh too.

**Custom key validation: 1ba84b74.** Keys must match `[A-Za-z0-9_-]+` at creation, and the Info panel rejects anything else with a message explaining why: dots break template lookup (Mustache nesting), and spaces and braces break the `{…}` and `{{…}}` syntax itself. For legacy keys from older or hand-edited notebooks, the condition picker, Visual Builder and computed insert menu offer only reference-safe keys, and a computed expression referencing an existing but unsafe key gets a targeted "rename it to use only letters, numbers, hyphens and underscores" error instead of a confusing miss.

**Removal-blocked message: 41d7c693.** The alert now gives a count and locates every reference as Form › Section › Field with its kind:

> Cannot remove "tests": it is referenced by 2 places: Test › Section 1 › Wet Notes (field condition); Test › Section 1 › Test Code (computed expression).

**Preview metadata context: 07154dcb** (pulled in from the follow-up list). `PreviewFormManager` now takes the notebook's custom metadata and builds a shared preview context used by conditions, templated strings and computed fields. Previously the preview passed a fake context to templates only, computed fields got no context at all, and condition evaluation never saw context values, so parent and related references didn't function in preview either. Metadata references now behave live in the preview and update as the Info panel changes, without a reload.

**Info panel UX and authoring docs: 7aab5cc1** (the high-value subset of the UX follow-up). Each custom key now shows a "Referenced by …" caption listing its referring conditions, templates and expressions: the same information the removal block reports, surfaced before you try. The key name field validates as you type. `docs/user/authoring/structure.md` no longer describes custom fields as compatibility leftovers; it covers referencing and links to the conditions doc. The larger UX ideas (renaming a key with its references rewritten, richer editing) stay as a future ticket, since rename is a uiSpec-wide mutation that deserves its own review.

**Stale conditions doc: da85b5c5.** `docs/user/field-types/shared-settings/conditions.md` no longer claims conditions can't reference other records; it points at the parent, linked and metadata reference docs.

Full test ladder green after merging in latest main (data-model, forms, web, app typecheck, api, with the pre-existing `PlanTemplateManager` flag-off failure aside), and the app smoke re-verified end to end: refresh propagation, conditions, templates and computed fields all behaving on new and existing records.

PeterBaker0 (approved):
Looks good with those updates. 

Would you mind adding some unit tests to assert invariants for your hash function e.g. that it doesn't change based on field ordering, whitespace, etc.

Caedous:
Added in edadd6c1, covers key-order insensitivity at any depth, value and value-whitespace sensitivity (whitespace inside a value is data, so it must change the ID), array order preserved (field order is meaningful), null/empty/absent distinguished, and server/project scoping.

Issue #2275, opened 2026-08-25
Port NotebookComponent to the notebook view component interface


jayenashar:
`NotebookView` (`app/src/gui/components/notebook/notebookView.tsx`) resolves a view from the plan view registry and falls back to `NotebookComponent`, which still takes only `project` and builds its own record list, tabs and actions.

Port it to `NotebookViewComponentProps` so the fallback is just another registered view and there is one interface rather than two. Worth leaving until that interface has settled.

Raised by @stevecassidy in review of #2271.

🤖 Filed with [Claude Code](https://claude.com/claude-code)

Caedous:
Implementation PR: FAIMS/FAIMS3#2289 — NotebookComponent ported to `NotebookViewComponentProps`; the fallback now takes the same props as registered plan views, so there is one view interface rather than two.

PR #2272, opened 2026-08-22
feat: put the notebook tab in the route

jayenashar:
## Ticket

#2266

## Description

The notebook tab was a `?tab=` search param mirrored into `tabIndex` state, so coming back from a record lost it.

## Proposed Changes

- The notebook route is a layout route, `/<notebooks>/:serverId/:projectId/:tab?`, with the record routes nested under it. Leaving a record is `..`, which lands on the tab it was opened from, including the my-records and other-records lists.
- The tab segment is optional, so record links written without one still resolve and land on the view's default tab.
- FAIMS3 keeps no list of tabs. `NotebookView` reads `:tab` and passes `tab` plus `actions.setTab` through `NotebookViewComponentProps`; each view maps its own slugs and falls back to its own default for an absent or unknown one. The two plan views now use named slugs rather than tab indices.
- `getEditRecordRoute` and `getViewRecordRoute` build an absolute route from the notebook the record routes nest under, so a record link carries the tab of whichever route its component rendered under rather than a hardcoded origin.
- Deleting a record navigated twice, invisibly so while both navigations used the same absolute route. `RecordDelete` now does it once, which drops the `handleRefresh` and `serverId` props that only existed to build the route it left for.
- `TolerantPrivateRoute` moves to the layout route, so the record routes inherit it instead of repeating it.

## How to Test

Open a notebook, switch tabs, and watch the last path segment change. Open a record from My Records or Other Records and press Back: it returns to that list rather than the default tab. A tab URL such as `/<notebooks>/<server>/<project>/settings` opens that tab on its own, an older record link with no tab segment still resolves, and a slug no view carries, such as `/<notebooks>/<server>/<project>/nope`, gives way to the default tab in the URL.

`e2e/test/specs/app/notebook-tab-route.e2e.ts` covers all five, and adds the record page's Edit action and a delete, both from a tab that is not the default so a dropped tab reads as a different URL. Every step fails with the code it covers reverted.

## Additional Information

Base is #2271, which brings in `NotebookView` and the plan view registry. Retarget to `main` once that merges.

`?tab=` bookmarks no longer select a tab; they land on the view's default.

A tab slug the view does not carry replaces itself with the view's default, so the URL names the tab on screen and a record opened from there carries a slug that resolves.

`map` is labelled "Map" in the default view and "Overview Map" in the list-of-records plan view. Same slug, two labels, left as they are.

[...]

PR #2289, opened 2026-09-02
Port NotebookComponent to the notebook view component interface (#2275)

Caedous:
## Ticket

#2275

## Description

`NotebookComponent`, the fallback view for a notebook with no plan or no registered plan view, still took only `project` plus a `tab`/`setTab` side channel, and fetched its own record list, sync audit, permissions and compiled spec — all work `NotebookView` already does once to build `NotebookViewComponentProps` for the plan views. This ports the fallback onto that interface, so every notebook view consumes the same props and there is one interface rather than two.

## Proposed Changes

- `NotebookComponent` takes `NotebookViewComponentProps`. Its own `useRecordList`, `useRecordAudit`, `useIsAuthorisedTo`, spec lookup and query-client invalidations go; the record lists, sync status, permission flags and refresh action now come from `props`, resolved once in `NotebookView`.
- The fallback render in `NotebookView` becomes `<NotebookComponent {...props} />`, spreading the props it was already assembling, and the TODO from the #2271 review goes with it.
- Settings, Details and Map content render through `props.components`, matching the plan views, so the metadata display's `templateId` comes from `NotebookView`'s wiring rather than a local store selector.
- The record tables' `handleRefresh` is the invalidation-based `refreshRecordList` rather than the hook's `refetch` — equivalent effect, and `RecordsTable` does not call it.

## Testing

Manual pass against a notebook with no plan: the five tabs render with the tab slug following in the URL, a `/settings` deep link opens on that tab, an unknown slug falls back to `my-records`, the other-records tab appears only when other users' records exist, record creation lands on the edit page and the list and tab count refresh on return, and the Map, Details and Settings tabs show their content through the injected components. A Counted plan notebook renders its plan view unchanged. `#2272`'s `notebook-tab-route.e2e.ts` exercises the default view's tabs, deep links, create and delete, which now all run through the ported paths.

## Additional Information

Tab slugs, labels, visible-type count filtering, the closed-notebook banner and the push-only sync banner are all unchanged, so the fallback behaves exactly as before — the point being that the view component interface is feature-complete against the prior view, as noted in review of #2271. No interface changes were needed; `NotebookComponent` stays a direct import in `notebookView.tsx` rather than a registry entry, since the registry is keyed by plan type and the fallback exists precisely for notebooks without one.

## Checklist


app/src/gui/components/notebook/DefaultNotebookView.tsx
```diff
@@ -1,30 +1,15 @@
...
-} from '../../../utils/customHooks';
-import CircularLoading from '../ui/circular_loading';
 import {DE_ACTIVATE_VERB} from '../workspace/notebooks';
 import AddRecordButtons from './add_record_by_type';
-import {MetadataDisplayComponent} from './MetadataDisplay';
-import {OverviewMap} from './OverviewMap';
 import PushOnlySyncBanner from './PushOnlySyncBanner';
 import {RecordsTable} from './record_table';
-import NotebookSettings from './settings';
+import {NotebookViewComponentProps} from './types';
```
stevecassidy, line 12: I feel like this shouldn't be index.tsx any more and the component should be something like DefaultNotebookView.   Since we have another big change in this part of the code pending let's wait until that is merged, pull those changes here and then do the rename.

stevecassidy (commented):
All good, only suggestion is about renaming a file, let's wait for that before merging.

app/src/gui/components/notebook/DefaultNotebookView.tsx
Caedous: Sounds like a plan

Caedous: #2281 merged in and conflicts resolved: the default view now uses `tab.current`/`tab.select`, takes `formTypes` on the add buttons, and passes `notebookRecords` to the injected map so a plan-less notebook still plots everything. Rename done as its own commit: `index.tsx` → `DefaultNotebookView.tsx`, component `DefaultNotebookView`

stevecassidy (approved):
Looks good now, thanks.

PR #2294, opened 2026-09-08
fix: export download link URL permission leakage

PeterBaker0:
Summarised version of the pen-test finding was basically "the link that's generated as part of the export workflow is too a) visible b) reusable c) leakable". The URL stood on it's own as a complete secret, and could leak in logs, browser history etc. It was also reusable, so could encourage 'spamming' or DDOSing. 

This PR moves to a cookie exchange flow where the API sets a cookie on the first call, then the browser sends that back. This maps based on the opaque ID from the first URL redirect to an ephemeral and one-use artifact in the auth DB. This is wired into TTL cleanup so we don't accrue them. This also tightens up the rate limiting and other controls, specifically for export.

Needed some thought around CORS, as the CORS on API means the cookie exchange was not possible due to browser security constraints. Some settings around this; 'lax' setting for eTLD + 1 (e.g. api.* to web.*). Some work done to ensure the right CORS origins are setup for the API.


api/src/corsAllowlist.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Origins Conductor may CORS-allow: Conductor, Control Centre, Designer,
+ * the field app, and http(s) entries from {@link config.redirectWhitelist}.
+ */
+export const buildCorsAllowlist = (): string[] => {
+  const candidates = [
+    config.conductorPublicUrl,
+    config.webAppPublicUrl,
+    config.designerUrl,
```
stevecassidy: designer doesn't have a URL...

api/src/downloadCookie.ts
```diff
@@ -0,0 +1,155 @@
+/**
+ * Browser-only download-grant cookie. The secret never appears in the URL.
+ * Headless clients ignore this cookie and use Authorization: Bearer instead.
+ */
+
+import {Response} from 'express';
+import {config} from './buildconfig';
+import {DOWNLOAD_GRANT_EXPIRY_MS} from './couchdb/downloadGrants';
+
+export const DOWNLOAD_COOKIE_NAME = 'faims_download';
```
stevecassidy, line 10: We should derive this from the product name (Fieldmark/DSS/Groundsight) so that if someone finds it they don't think it's from a third party source.  However, we don't have any config values for that in the API just now (I think) - should pass APP_NAME in - add it separately for now and we'll bring it together in the config refactor.

api/src/api/notebooks.ts
```diff
@@ -199,48 +195,95 @@ const UpdatedTimeQuerySchema = z
...
 // Formats requiring a view ID
-const REQUIRES_VIEW_ID: DownloadFormat[] = ['csv'];
-
-// Download tokens last this long
-const DOWNLOAD_TOKEN_EXPIRY_MINUTES = 5;
+const REQUIRES_VIEW_ID: ExportFormat[] = ['csv'];
 
-const generateDownloadToken = async ({
-  user,
+/**
```
stevecassidy: maybe put the export stuff in a separate module - this one is very big now and it would be good to have this fn nearer the route that it serves.

stevecassidy (approved):
Looks good, one request for a bit of code re-org but otherwise good. Thanks.

api/src/downloadCookie.ts
PeterBaker0: https://github.com/FAIMS/FAIMS3/issues/2297

api/src/api/notebooks.ts
PeterBaker0: Done

Issue #2296, opened 2026-09-08
Add support for plan instantiation when creating a notebook from a template in web/


Caedous:
The API already handles plan instantiation: `CreateNotebookFromTemplateSchema` accepts an optional `planConfig`, and `createNotebookFromTemplate` (`api/src/couchdb/templates.ts`) validates it against the plan type's `configSchema`, calls `instantiatePlan`, and rejects the request when the template has a `planTemplate` but no `planConfig` is supplied.

The web manager's create-from-template flow does not send `planConfig`, so a template that carries a plan cannot currently be turned into a notebook from `web/`.

Add plan instantiation support to the notebook creation flow in `web/`: when the selected template has a `planTemplate`, render a configuration step for the plan type and pass the resulting `planConfig` with the create request. Cover at least the two built-in plan types:

- **Counted** - `numberRequired` and `allowExtraRecords`
- **List of Records** - the list of records pre-filling the template's `recordFields`

Templates without a `planTemplate` are unaffected. Unregistered plan types should block creation with a clear message rather than submit an invalid request.

Originates from FAIMS/GroundSight-FAIMS#140 (private), part of the data collection plans work moved over in #2271.

PR #2298, opened 2026-09-08
Feature/2296 plan instantiation web

Caedous:
## Ticket
#2296
## Description
The API already instantiates plans: `createNotebookFromTemplate` validates a `planConfig` against the plan type's `configSchema`, calls `instantiatePlan`, and rejects creation when a template has a `planTemplate` but no config was supplied. The web manager could not send one, so any template with a plan was uncreatable there and the create form carried a placeholder disabling submission. This replaces the placeholder with a configuration step for the two built-in plan types and sends the result as `planConfig`.
## Proposed Changes
- New `web/src/components/plans/`: a config registry mirroring the Designer's plan registry (`registry.ts`, lazy built-ins, injectable for tests). A `PlanConfigType` is either field-based, supplying `fields(context)` and `toConfig(values, prefix)`, or component-based, supplying a `ConfigForm`. Shared types and label helpers sit in `types.ts` so nothing imports in a cycle.
- Counted is field-based (`countedPlanFields.ts`): `numberRequired` as a number field with the data-model schema and `allowExtraRecords` as a checkbox, appended to the create form under a `PlanHeading` carrying the plan's label and description, so validation, layout and the disabled-submit state come from the `Form` module. Field names are prefixed by position since a `planId` may contain characters react-hook-form reads as a path.
- List of Records is component-based (`ListOfRecordsPlanConfigForm.tsx`): an editable table with a column per `recordFields` entry, choosing the input by the field's `type-returned` (Integer/Number as numeric, Bool as checkbox, else text) and giving rows sequential reference ids never reused after removal, since the id becomes the record's `planReference`. Only fields in `LIST_PLAN_SUPPORTED_FIELD_TYPES` (data-model) get a column, and the Designer's picker offers only those; an existing template's other fields are named in a note as entered on each record in the app. Rendered in the form footer through `PlanConfigSection`, which also explains an unregistered plan type.
- `ListOfRecordsPlanDialog` excludes unsupported field types from its picker and marks an already-chosen one with a warning chip.
- `usePlanConfigs` drives both create forms: one section per plan template, `appendTo` for the field-based ones, `footer` for the component-based ones, `toPlanConfigs(values)` assembling `planConfigs` keyed by `planId` on submit, and `planSubmissionGate` holding submission while the template loads or fails, a component-based plan lacks a config, or a plan type is unregistered.
- `create-project-from-template.tsx` (template page) and `create-project-form.tsx` (notebooks-page picker) both use it; the picker fetches the full template once one is selected via a `Field.onChange` watcher added to `Form`. Failed creation surfaces the API's message via `errorMessageFromNotebookJsonBody` in both.
- `createProjectFromTemplate` in `project-hooks.ts` accepts optional `planConfigs`; both create dialogs get `max-h-[90vh] overflow-y-auto` so a table cannot push the submit button off-screen.
- New `Data-Collection-Plans.md` developer doc covering the data model, instantiation, Designer authoring, the two config-type kinds and app views, registered in the docs index.
## Testing
`planConfig.test.tsx` (26 tests) covers the registry (built-ins and their kinds, duplicate registration), `countedPlanFields` (prefixed names, form label, positive-whole-number schema, config assembly), the List form (typed values keyed by reference, references not reused after removal, incomplete when emptied, unsupported-type flag), `planSubmissionGate` (no plans, loading, error, field-based plans left to the form, component-based blocked until configured, unregistered types) and `PlanConfigSection` for registered and unregistered types. Manual pass on three templates: Field-Sampler with a Counted plan, Field-Sampler with a List of Records plan over Date and Time plus a Number field, and `api/notebooks/two-plan-templates.json` carrying one of each. Template page: Counted renders as form fields under the plan's label and description, blocks an empty or zero count with the field's own message, creates with `5`; List renders the bordered table with the Date and Time column omitted and named in the note below the table and the Number column numeric; the two-plan template shows both sections and blocks until the list has a row.

counted-plan-setup

list-plan-setup

both-plan-setup

Notebooks-page picker: selecting the two-plan template fetches it and renders the same two sections beneath the picker; switching templates drops or resets them.

both-plan-setup-from-notebook

The created notebooks' `plans` arrays carry each plan's `planId`, `label` and the entered values, with numbers stored as numbers; the app offers the plan chooser for the two-plan notebook and opens each plan's view with the values pre-filled.

From the notebooks page, picking a plan template reports the API's "plan config must be provided" message.
## Additional Information
The web unit suite is not in CI (`build-lint.yml` filters to api, data-model, app, forms and infra). Running it locally, `PlanTemplateManager.test.tsx`'s "flag off" case fails on `main` as well: its `vi.mock` of `buildconfig` does not intercept in Vitest browser mode, so the component reads the real `VITE_ENABLE_PLANS_IN_DESIGNER`.
## Checklist


web/src/components/plans/CountedPlanConfigForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+        />
+        <Label
+          htmlFor="plan-allow-extra"
+          className="font-normal cursor-pointer"
+        >
+          Allow extra records beyond the number required
+        </Label>
+      </div>
+      <p className="text-sm text-muted-foreground">
+        The {config.notebookName} will track progress towards this count.
```
PeterBaker0: Can you use the built in Form module that /web already heavily uses - feel free to extend it as needed. It manages validation, consistent styling/display etc.

web/src/components/plans/ListOfRecordsPlanConfigForm.tsx
PeterBaker0, line 1: If this one is too custom to use a normal form - makes sense to do something special.

PeterBaker0 (dismissed):
1) I think this doesn't stand on its own / isn't really merge-ready until we have the Project -> Create Project -> Select template workflow. This is the more common approach in the UI we've seen users rely on. I've pushed a draft implementation up to **feat/project-plan-config-instantiation** - please feel free to merge as is, use for inspiration, or ignore. 
2) Some concerns around field selection in the list plan selector - we can choose any field, and then put any text data in - for now we either need to a) support the types completely (i.e. have custom data entries for each data type) b) only display options which we can support with simple string input c) show some aggressive warnings and ticket up for later?
3) Please use the <Form> component in Web wherever possible, extend as needed, rather than re-writing a custom form in the plan config. Probably makes sense for list of records plan as it's such a custom interface.

Caedous:
> 1. I think this doesn't stand on its own / isn't really merge-ready until we have the Project -> Create Project -> Select template workflow. This is the more common approach in the UI we've seen users rely on. I've pushed a draft implementation up to **feat/project-plan-config-instantiation** - please feel free to merge as is, use for inspiration, or ignore.
> 2. Some concerns around field selection in the list plan selector - we can choose any field, and then put any text data in - for now we either need to a) support the types completely (i.e. have custom data entries for each data type) b) only display options which we can support with simple string input c) show some aggressive warnings and ticket up for later?
> 3. Please use the  component in Web wherever possible, extend as needed, rather than re-writing a custom form in the plan config. Probably makes sense for list of records plan as it's such a custom interface.

Round two pushed, on top of a merge of `main` so it picks up #2281.

- Counted is now field-based: the registry entry supplies `Field`s appended to the create form under the plan's label and description, validated and styled by `Form`; the config is assembled from form values on submit. `CountedPlanConfigForm` is gone.
- List of Records stays a component in the footer, per your note.
- One config per plan template, keyed by `planId`, sent as `planConfigs`; both entry points render a section per plan and the gate waits for all of them. Tested on `two-plan-templates.json`.
- Field types, option (b): `LIST_PLAN_SUPPORTED_FIELD_TYPES` in data-model (String, Integer, Number, Bool). The Designer's picker only offers those, and the config table only gives columns to those, with a note naming any other field an existing template carries. Full per-type entry can be a later ticket.

Screenshots updated in the description.

strict types for prefill

stevecassidy (approved):
Looks good now.  I think we'll want CSV upload of values for list of records plan but that can come later.

PR #2303, opened 2026-09-09
feat: create a linked record as one engine operation

jayenashar:
## Ticket

None.

## Description

Three places each built "create a record and link it to its parent" by hand: the notebook view's new action, the related-record field's create button, and the navigation "create another" path. Each hardcoded a parent/child edge and its vocab pair, and only one honoured the field's `multiple`.

## Proposed Changes

- `createRelatedRecord` on the data engine: reads the field, derives the related form, the relation and its vocab pair, refuses a link the field cannot take, creates the record with the right edge, and returns the value the parent's field must then hold.
- All three callers use it. Writing the parent's side stays with each, because a form writes it through the open form and then commits, while a view with no open form writes a revision.
- `relationTypeToPair` moves to the engine and the field's copy re-exports it.
- `canEditRecord` on the notebook view's actions calls `canEditProjectRecord` rather than rebuilding it.

## How to Test

`pnpm --filter @faims3/data-model test` and `pnpm --filter @faims3/app test`. Through the UI, creating from a related-record field and from a view should both leave the parent listing the new record.

## Additional Information

Both refusals happen before the record is written, so a link the field cannot take no longer strands a child its parent never lists. The child is still written before the parent's link, so a failure between the two is reported as created-but-not-linked.

A `multiple: false` field could previously collect a second record through the field's create button, since that path always appended a list. It now refuses like the others.

## Checklist

### DIFF
--- a/app/src/gui/components/notebook/notebookView.tsx
+++ b/app/src/gui/components/notebook/notebookView.tsx
@@ -12,6 +12,7 @@ import {
   FormUpdateData,
   MinimalRecordMetadata,
   ProjectStatus,
+  relatedRecordAvpEntries,
 } from '@faims3/data-model';
 import NotebookComponent from '.';
 import {addAlert} from '../../../context/slices/alertSlice';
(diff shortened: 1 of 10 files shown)


app/src/gui/components/notebook/notebookView.tsx
```diff
@@ -232,6 +257,141 @@ function NotebookViewWithSpec({
...
+          addAlert({
+            message: 'You do not have permission to add to that record',
+            severity: 'error',
+          })
+        );
+        return;
+      }
+
+      // The pair a Child related-record field stores, the parent's view first.
+      const relationTypeVocabPair: [string, string] = [
```
stevecassidy: If we're doing this it should work for any related record, not just parent/child, these should come from the related record field being modified.

app/src/gui/components/notebook/notebookView.tsx
```diff
@@ -232,6 +257,141 @@ function NotebookViewWithSpec({
...
 
+  /**
+   * Create a child record and navigate to its edit page, writing both halves of
+   * the link so the parent form reads as it would after an in-form create.
+   */
+  const createChildRecord = useCallback(
+    async ({
+      formType,
+      parentRecordId,
+      parentFieldId,
```
stevecassidy: we should have an optional planReference since the plan may want to mark the new record

app/src/gui/components/notebook/notebookView.tsx
```diff
@@ -232,6 +257,141 @@ function NotebookViewWithSpec({
     ]
   );
 
+  /**
+   * Create a child record and navigate to its edit page, writing both halves of
+   * the link so the parent form reads as it would after an in-form create.
+   */
+  const createChildRecord = useCallback(
```
stevecassidy, line 263: It feels like this should really be a combination of createRecord and navigateToRecord where the createRecord call is able to handle the relationship setup.  It also feels like this should be in the data engine rather than here in the UI code - this is a core operation and the fact that we have to wrangle internal details here means we got the abstraction wrong.  createRecord claims to be able to create relationships (see the tests in engine.form.tests.ts) so if it isn't doing it right we should fix that rather than patching here.

stevecassidy: Ok, I see that createRecord's ability to make relationships is that it can set the value in the parent, the logic to update the parent field with a new created relation is in the RelatedRecordField implementation - which handles the general relation case, not just parent/child.   Maybe we can extract that code into the engine and then call if from here and the RelatedRecordField?  The mutation function in the field embodies the two operations of creating the new record and updating the parent field which is what you need here.

stevecassidy (commented):
Thanks for highlighting this gap, I've left a couple of notes - I think we need to do this more centrally.

app/src/gui/components/notebook/notebookView.tsx
jayenashar: Done in b9e7adb31. `FormOperations.createRelatedRecord` now reads the field, derives the related form, the relation and its vocab pair, refuses a link the field cannot take, and only then creates the record with the right edge. It returns the value the parent's field must hold.

Writing that value stays with each caller, because the two write paths are genuinely different: the field writes it through the open form and then commits, so a revision written underneath would be lost to that commit; a view with no open form has to write a revision. Everything that was actually duplicated (the derivation, the cardinality rule, the append) is in the engine.

Three call sites use it now, not two: the notebook view's action, this field's create mutation, and `useNavigationDataPrep`'s "create another" path, which was building the same edge from `head.relationType` and ignoring `multiple`.

Two things worth flagging:
- Refusing before the write matters. An earlier pass validated after `createRecord` and could leave a child whose parent never listed it.
- A `multiple: false` field could previously collect a second record through this field's create button, since it always appended an array. It now stores what the engine returns, so that path refuses like the others.

app/src/gui/components/notebook/notebookView.tsx
jayenashar: Agreed, and fixed. Nothing is hardcoded to parent/child now: `createRelatedRecord` reads `relation_type` off the field and hangs the new row off `parent` or `linked` accordingly, with the pair from `relationTypeToPair`.

That helper already existed in `fieldRegistry/fields/RelatedRecord/utils.ts`, returning exactly the pair I had written out by hand. It moved into the engine and that file now re-exports it, so there is one definition. It also kept its `RelatedType` parameter rather than widening to `string`, or a typo would silently select the Linked pair.

The action lost its `formType` argument entirely, since `related_type` comes from the field.

app/src/gui/components/notebook/notebookView.tsx
stevecassidy: This looks good but the name isn't right now - createRelatedRecord?

stevecassidy (approved):
Happy with this change now. Suggest the action passed in is called createRelatedRecord since it is now more general.   A good clean up in the end and fixing inconsistent behaviour.

app/src/gui/components/notebook/notebookView.tsx
jayenashar: renamed in 3a6737fae. the view action is `createRelatedRecord` now, matching the `engine.form.createRelatedRecord` it delegates to, and the two doc comments and the `console.error` that still said "child record" went with it. no behaviour change, and the notebookView tests still pass.

the parameters stayed `parentRecordId` and `parentFieldId`, since that is what the engine op takes for a Linked relation too.

library/forms/lib/fieldRegistry/fields/RelatedRecord/utils.ts
```diff
@@ -2,12 +2,6 @@
...
-export function relationTypeToPair(type: RelatedType): [string, string] {
-  if (type === 'faims-core::Child') {
-    return ['has child', 'is child of'];
-  } else {
-    return ['is linked to', 'is linked from'];
-  }
-}
+// One definition, in the engine, so this field and a view creating a related
+// record from outside a form agree on what a relation stores.
+export {relationTypeToPair} from '@faims3/data-model';
```
PeterBaker0: The field can just import this directly from the data model.

PeterBaker0 (approved):
Looks good thanks - just the one tidy up from the classic AI 're-export' for no reason

Issue #2312, opened 2026-09-13
Injected notebook view components remount on every record poll


jayenashar:
`app/src/gui/components/notebook/notebookView.tsx` defines `NotebookSettings`, `MetadataDisplayComponent` and `OverviewMap` inline inside the `props` useMemo, so every recompute is a new element type and React remounts the subtree. `records` is a memo dep and the record list polls every 10s, so an open settings dialog closes itself.

`OverviewMap` closes over `planRecords`, so hoisting it needs its records passed in rather than captured.

*Drafted with AI assistance, reviewed before posting.*

PR #2313, opened 2026-09-13
fix(app): stop the injected notebook view components remounting on every record poll

jayenashar:
## Ticket

#2312

## Description

`notebookView` built the three components it injects into a notebook view inside the `props` useMemo. A component rebuilt on each recompute is a new element type, so React unmounted and remounted the subtree under it. An open settings dialog closed itself while the user was reading it.

## Proposed Changes

- The three injected components are memoised one by one, each keyed on what it actually reads, rather than together inside the props memo.
- `NotebookSettings` and `MetadataDisplayComponent` read neither the records nor anything else that polls, so they now keep their identity.
- `OverviewMap` does read the records, so it is still rebuilt when they change. `useRecordList` shares its array structurally, so a poll returning the same records rebuilds nothing.
- A test asserting the settings component survives a record change with its state intact.

## How to Test

Open a notebook, go to Settings, open the deactivate dialog and leave it open. Before this change it closed on its own within about ten seconds. `npx vitest --run src/gui/components/notebook` from `app/`.

## Additional Information

`OverviewMap` keeps a `selectedFeature` that it clears only when the user closes the popover, so a pin for a record that has since gone can outlive it. That predates this change and is untouched here.

An earlier commit on this branch gave `OverviewMap` a stable identity by holding the records in a ref. Review pointed out that reads a value React has not committed, so a concurrent render could plot records from an abandoned pass. The second commit replaces it.

## Checklist

### DIFF
--- a/app/src/gui/components/notebook/notebookView.tsx
+++ b/app/src/gui/components/notebook/notebookView.tsx
@@ -31,7 +31,7 @@ import CircularLoading from '../ui/circular_loading';
 import {getNotebookView, PlanChooser, resolvePlanViews} from './plans';
 import {recordsClaimedBy} from './plans/planViewRecords';
 import {useRecordAudit} from '../../../utils/apiHooks/notebooks';
-import {useCallback, useMemo, useState} from 'react';
+import {useCallback, useMemo, useRef, useState} from 'react';
 import {config} from '../../../buildconfig';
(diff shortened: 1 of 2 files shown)


PeterBaker0 (approved):
Looks good

PR #2315, opened 2026-09-13
feat: include a record's child records in its History tab

### LINKED ISSUE #2314: A record's history cannot show work done on its children
`getHistoryData` (`library/data-model/src/databaseEngine/engine.ts`) returns one record's revisions, so the History tab on the record page shows nothing about the records hanging off it. Reading a parent's whole story means opening each child by hand.

Reaching a child's own History tab is not always possible either. `RelatedRecord` renders a link card to the child's read-only page only on mobile, past `RENDER_NEST_LIMIT`, or on a cycle; within that limit on desktop the child renders inline as an accordion, which has no tabs.

The recursive status report already walks the Child-type links and handles deleted, unreadable and cross-project children plus cycles, so the same traversal can serve both.

jayenashar:
## Ticket

Closes #2314

## Description

A record's History tab showed only that record's own revisions, so work done on its children was invisible from the parent. Reaching a child's own History tab is not always possible either: within `RENDER_NEST_LIMIT` on desktop the child renders inline as an accordion, which has no tabs.

## Proposed Changes

- `computeRecursiveRecordHistory` returns a record's revisions together with its children's, walking the Child-type links recursively.
- The whole child traversal moves into `databaseEngine/childRecordTree.ts` as `walkChildRecordTree`, and both this walk and the recursive status report now call it, so the two cannot disagree about what counts as a child. Deleted, unreadable and cross-project children drop out, cycles are cut, and Linked relations are not followed. A node's own work runs inside the same `Promise.all` as its child branches, so fetching a trail does not wait for the subtree below it.
- The History tab renders that tree: each child field lists the records under it with their own trails, and each child links to its own page.
- The tab's query moves onto the hydration key prefix as `buildRecordHistoryKey`, beside `buildStatusReportKey`. It sat outside that prefix before, so it survived notebook removal and project-wide resets and could retry into a torn-down database handle.

## Behaviour change

A record whose form is no longer in the ui-spec now raises an error on the History tab rather than rendering an unlabelled node with no children. That is what the Status tab already did, and sharing one traversal is what brings the two into line.

## How to Test

`pnpm exec turbo test typecheck lint build --filter=@faims3/data-model --filter=@faims3/app`, then open a record with child records and look at its History tab. The typecheck matters here: `tests/recordHistory.test.ts` is in the data-model tsconfig now, so a wrong engine call in that suite fails the build instead of passing silently.

## Additional Information

[Demo](https://drive.google.com/file/d/1HsTpw-uQzo5-AHCBj36u6A8Ul1oGKsyr/view) (2:26, captioned): a record is created with no children, its History tab shows only its own revision, a child is added, and the child's trail then appears under the parent.

A record linked from two fields of the same parent is walked once rather than once per field, and revision anchors are scoped to their record so the same record under two fields cannot produce duplicate DOM ids.

No unit test was added for the tab itself: the equivalent recursive view, `app/src/gui/components/record/status.tsx`, has none either, and this layer is covered by e2e.

## Checklist


library/data-model/src/databaseEngine/recordHistory.ts
```diff
@@ -0,0 +1,141 @@
...
+const isChildHistory = (
+  outcome: RecursiveRecordHistory | null | undefined
+): outcome is RecursiveRecordHistory => !!outcome;
+
+/**
+ * One node of the walk: the record's own revisions plus its children's. Null
+ * when the record is a deleted child, or would close a cycle, matching the
+ * status report's treatment of the same data.
+ */
+async function walk(
```
jayenashar, line 39: same shape as `statusReport.ts`'s walk: cycle cut, `getExistingFormData`, deleted check, `Promise.all` over distinct children with `absorbSkippableChildError`, `flatMap` to `childFields`. it has already drifted, since `statusReport.ts:307` throws `UnknownFormTypeError` for a `formId` missing from `viewsets` and this doesn't, so the same record errors on the status tab and renders here as an unlabelled node with no children. should the traversal be shared too, not just the link collection?

jayenashar (commented):
self-review before this leaves draft.

library/data-model/src/databaseEngine/recordHistory.ts
```diff
@@ -0,0 +1,141 @@
...
+        },
+      ];
+    }
+  );
+
+  return {
+    recordId,
+    hrid: node.context.hrid,
+    formId: node.formId,
+    entries: await engine.form.getHistoryData({recordId}),
```
jayenashar, line 102: evaluated after the `Promise.all`, so every node waits for its whole subtree before fetching its own trail.

library/data-model/src/databaseEngine/types.ts
```diff
@@ -776,6 +776,31 @@ export const revisionHistoryEntry = z.object({
...
+  entries: RevisionHistoryEntry[];
+  /** One entry per Child-type field that links to at least one live record. */
+  childFields: RecursiveRecordHistoryChildField[];
+}
+
+/** One Child-type field, and the histories of the records it links to. */
+export interface RecursiveRecordHistoryChildField {
+  fieldId: string;
+  /** Form the children are created from (`related_type`). */
+  relatedFormId: string;
```
jayenashar, line 800: never read: `HistoryNode` labels each child from its own `formId`. `status.tsx:109` uses the equivalent field for its "no X records yet" copy, which this tree has no equivalent of.

library/data-model/tests/recordHistory.test.ts
```diff
@@ -0,0 +1,270 @@
...
+  const databaseName = 'test-record-history-db';
+
+  // The same notebook the status report is tested against, so both walks are
+  // exercised over one shape of child links.
+  const uiSpecPath = path.join(__dirname, 'statusReportUiSpec.json');
+  const {uiSpec: rawUiSpec} = JSON.parse(
+    fs.readFileSync(uiSpecPath, 'utf-8')
+  ) as NotebookDefinition;
+  compileUiSpecConditionals(rawUiSpec);
+  const uiSpec = rawUiSpec as unknown as CompiledNotebookUiSpec;
```
jayenashar, line 43: this load, plus `CHILD_PAIR`, `link`, `beforeEach`/`afterEach`, `create` and `childField`, is verbatim from `statusReport.test.ts`. `tests/mocks.ts` is the precedent for sharing them.

library/data-model/tests/recordHistory.test.ts
```diff
@@ -0,0 +1,270 @@
...
+    updatedBy = USER
+  ) =>
+    engine.form.updateRevision({
+      revisionId: created.revisionId,
+      recordId: created.recordId,
+      update,
+      mode: 'merge',
+      updatedBy,
+    });
+
```
jayenashar, line 82: nothing calls this, and no test gives a record a second revision.

```suggestion
```

app/src/gui/pages/viewRecord.tsx
```diff
@@ -391,60 +400,18 @@ const ViewTabContent: React.FC<ViewTabContentProps> = ({
...
-
-  if (!historyData) {
-    return (
-      <Box sx={{p: 2}}>
-        <Typography color="error">Record data not found.</Typography>
-      </Box>
-    );
-  }
+}> = ({entries, recordId, uiSpec}) => {
+  const historyData = entries;
```
jayenashar, line 413: leftover from the extraction, since `entries` is the prop. the map callback at `:431` shadows it too.

app/src/gui/pages/viewRecord.tsx
```diff
@@ -509,6 +475,127 @@ const HistoryTabContent: React.FC<{
...
+ */
+const HistoryTabContent: React.FC<{
+  recordId: RecordID;
+  projectId: ProjectID;
+  dataEngine: DataEngine;
+  uiSpec: NonNullable<ReturnType<typeof compiledSpecService.getSpec>>;
+  notebook: RecordRouteNotebook;
+}> = ({recordId, projectId, dataEngine, uiSpec, notebook}) => {
+  const {data, isError, isPending, error} = useQuery({
+    queryKey: ['historyData', projectId, recordId],
```
jayenashar, line 553: record-scoped queries here sit under `HYDRATION_KEY_PREFIX` through `buildStatusReportKey` (`customHooks.tsx:166`), which is how `cancelProjectQueries` and `invalidateProjectHydration` reach them. the key is changing anyway, so should it move onto that prefix?

library/data-model/src/databaseEngine/recordHistory.ts
jayenashar: yes, shared in 7ff4d37. the walk now lives in `library/data-model/src/databaseEngine/childRecordTree.ts` as `walkChildRecordTree`, and both `computeRecordStatusReport` and `computeRecursiveRecordHistory` call it, so the cycle cut, the deleted check, the skip rules and the one-fetch-per-record come from one place.

worth naming as a behaviour change: a record whose form is missing from `viewsets` now throws `UnknownFormTypeError` on the history tab too, so it errors instead of rendering an unlabelled node, matching the status tab. new tests in `tests/recordHistory.test.ts` cover the root and the child case.

library/data-model/src/databaseEngine/recordHistory.ts
jayenashar: fixed in 7ff4d37. the walker takes a `startOwnWork` callback and awaits it in the same `Promise.all` as the child branches, so a node's own `getHistoryData` overlaps its subtree instead of queueing behind it. the status report's scoring goes through the same hook.

library/data-model/src/databaseEngine/types.ts
jayenashar: dropped in 7ff4d37. `status.tsx:109` reads the status report's child-field type, which keeps it; the history tree has no use for it.

library/data-model/tests/recordHistory.test.ts
jayenashar: done in 7ff4d37. `tests/childTreeTestSupport.ts` now holds the ui-spec load, `CHILD_PAIR`, `link`, `createTestEngine`, `createRecord` and `childField`, and both suites import it.

library/data-model/tests/recordHistory.test.ts
jayenashar: kept and now used, in 7ff4d37, rather than deleted: multi-revision children are the point of the feature, so the gap was the missing test, not the helper. `a later edit to a child shows in its trail` edits a child as another user and asserts two entries in the child's trail.

the helper was also broken. it passed `mode: 'merge'`, which is not in `AvpUpdateMode`, and updated the creation revision in place, so the record never got a second revision. it now does `createRevision` and then `updateRevision` with `mode: 'parent'`. `tests/recordHistory.test.ts` is in the data-model tsconfig include now, so that class of error fails the build.

app/src/gui/pages/viewRecord.tsx
jayenashar: removed in fa87457, and the map callback's array parameter renamed `newestFirst` so it no longer shadows.

app/src/gui/pages/viewRecord.tsx
jayenashar: moved in fa87457. `buildRecordHistoryKey` sits beside `buildStatusReportKey` in `customHooks.tsx` and returns `[HYDRATION_KEY_PREFIX, projectId, recordId, 'recordHistory']`, so `cancelProjectQueries` and `invalidateProjectHydration` reach it.

PR #2316, opened 2026-09-14
Feat/2311 list of forms plan type

### LINKED ISSUE #2311: Add a ListOfForms plan type and drive the default notebook view from a set of forms
`DefaultNotebookView` (`app/src/gui/components/notebook/DefaultNotebookView.tsx`) already presents a set of forms for creation and browsing, but the set is fixed to `visible_types`, and `RecordsTable` reads `visible_types` again on its own for row filtering and column shape. Neither can be pointed at a different set of forms; `CountedPlanView` and `ListOfRecordsPlanView` pass a plan's records to `RecordsTable`, which drops them whenever the plan's form is not in `visible_types`.

We need a workflow for creating calibration records for equipment: present the known calibration forms and allow records of those forms to be created and browsed. That is the default view driven by a configured list of forms rather than by `visible_types`.

Add a `ListOfForms` plan type configured with a list of form names to present, with a view in the app, and consider making it the default view when no plan is specified, configured from `visible_types`. `RecordsTable` will need to accept a set of forms rather than reading `visible_types` directly.

Notebooks without a plan are unaffected.

Originates from the GroundSight calibration workflow (FAIMS/GroundSight-FAIMS, private), part of the data collection plans work moved over in #2271.

Caedous:
# Feat: Add a ListOfForms plan type and drive the default notebook view from a set of forms
## Ticket
Closes #2311.
## Description
Adds a `ListOfForms` plan type whose template names the forms to present, with a view in the app that presents those forms for creating and browsing records. `DefaultNotebookView` already did this for the notebook's visible forms, with the set fixed to `visible_types` and `RecordsTable` reading `visible_types` again on its own, so neither could be pointed at a different set. The default view now renders the same component configured from `visible_types`, and a `ListOfForms` plan renders it configured from the plan. Notebooks without a plan behave as before.
## Proposed Changes
In data-model, `listOfFormsPlan.ts` defines the type: template and plan carry `formTypes: string[]`, at least one and none repeated, and the config schema is empty because the template fixes the forms. `builtins.ts` and `PlanTypeMap` register it.

In the app, the body of `DefaultNotebookView` moves to `ListOfFormsView`, which takes the forms to present and an optional `planReference`. `DefaultNotebookView` renders it with `getVisibleTypes`. `ListOfFormsPlanView` renders it with `plan.formTypes`, lists every notebook record of those forms rather than only the records the plan claims, and claims the records it creates for the plan; `plans/index.tsx` registers it. `RecordsTable` gains a `formTypes` prop for row filtering and column shape in place of reading `visible_types` itself. The Counted and List of Records views pass their own form, which stops those tables dropping the plan's records when its form is not in `visible_types`.

In web, `ListOfFormsPlanDialog` authors the plan in the designer with a label, a description and a checkbox per form, saved in the order the template declares them, and `plans.tsx` registers it. The web manager's config registry gains a field-based entry with no fields, so a template carrying this plan is not blocked at notebook creation. Two things surfaced during the manual test: `Form` only rendered a divider placed before a field, so a plan contributing no fields lost its heading; it now renders dividers placed past the last field, and `PlanHeading` takes a note for that case. The designer's plan row summarised a plan by a single `formType`; it now lists all the plan's forms.

`Data-Collection-Plans.md` covers the new type and the `formTypes` prop on `RecordsTable`.
## How to Test
1. `pnpm --filter @faims3/data-model build`, then `npx jest` in `library/data-model`: 45 suites pass, including `plans.listOfFormsPlan.test.ts`
2. `pnpm --filter @faims3/app test`: 38 files pass, including `ListOfFormsPlanView.test.tsx`
3. `pnpm --filter web exec vitest run src/components/plans src/designer/components/plans/ListOfFormsPlanDialog.test.tsx`: `planConfig.test.tsx` (18) and `ListOfFormsPlanDialog.test.tsx` (5) pass
4. On a local stack with `VITE_ENABLE_PLANS_IN_DESIGNER=true`, open a template with two forms in the designer, add a List of Forms plan and tick both forms. Reopen the plan: both stay ticked, and the plan row reads `List of Forms plan: <form>, <form>`
5. Create a notebook from the template in the web manager: the create form shows the plan's heading with "This plan needs no configuration." and submits
6. Activate the notebook in the app. It opens into the plan view with an alert naming the plan and its forms, one add button per form, and My Records / Map / Details / Settings tabs. Add a record from each button: both list under My Records with a Type column
7. Open a notebook with no plan: add buttons, tab counts, tables and map are unchanged
## Additional Information
`PlanTemplateManager.test.tsx` fails two flag-dependent tests locally whichever way `VITE_ENABLE_PLANS_IN_DESIGNER` is set: its mock of `buildconfig` does not take effect in browser mode, so the component reads the real flag. Noted on #2278 and #2296; not changed here.

**Authoring the plan in the designer:**

lof plan

**Creating a notebook from the template:**

createproject list of forms

## Checklist


stevecassidy (approved):
This turned out to be a very neat refactoring of the default view, very nice! Thanks.

PR #2317, opened 2026-09-14
feat: notebook sem versioning

PeterBaker0:
A significant upgrade to notebook versioning; implements a semantic versioning protocol instead of plain whole-number versioning. Each semantic level (major, minor, patch) introduces semantics which are consistently applied in the app as to how to handle these differences.

- The major version (1.0.0, 2.0.0 etc) indicates a complete breaking structural change, and the app will not attempt to render or handle that notebook.
- Minor changes (1.1.0, 1.2.0 etc) are handled by the app by showing a warning to the user that the notebook is in a version the app can't understand, but it proceeds to attempt to handle it in the safest way it can.
- Patch changes (1.0.1, 1.0.2 etc) are completely handled as per normal by the app


docs/developer/docs/source/markdown/MetadataMigrationGuide.md
```diff
@@ -56,6 +56,12 @@ Attempt to coordinate step 1/2 below.
 
 The app behaviour is likely to be unstable or completely broken when the app is on v1.5.2 or prior while the backend is ahead, and vice versa.
 
+**Semver epoch (later release).** A subsequent release replaces the linear `uiSpec.schemaVersion` sequence (`1.0` … `7.0`) with strict semantic versions starting at **`1.0.0`**, and adds a tiered compatibility check in the app (see [Notebook migrations](./NotebookMigrations.md)). That release:
+
+- Rewrites every pre-semver design (`legacy → 1.0.0`) on API startup (`MIGRATE_NOTEBOOKS_ON_STARTUP`) and on read paths; there is no Couch DB version bump for it.
```
stevecassidy, line 61: It might be time to get rid of this - I don't think we would have this disabled now.

stevecassidy (commented):
I like the approach, seems sound.   I don't think we should allow skipping of migrations on startup now though.

docs/developer/docs/source/markdown/MetadataMigrationGuide.md
PeterBaker0: actioning

PeterBaker0 (commented):
Self-reviewed and happy now once I fix dodgey fallback to api/.env!

stevecassidy (approved):
Happy with this, thanks.

PR #2319, opened 2026-09-16
fix: fastlane version bug with openSSL

PeterBaker0:
Issue with openSSL and fastlane version compat. 

Runs a mac silicon fastlane update in the bundles.


ranisa-gupta16 (approved):
thankyou..

PR #2321, opened 2026-09-16
docs: add CITATION.cff

saross:
## Ticket

None. Raised from work on the Fieldmark website's `/cite` page, which
currently has to tell researchers there is no software citation yet.

## Description

Adds a `CITATION.cff` at the repository root. GitHub reads this file and
renders a "Cite this repository" button with APA and BibTeX forms, and
reference managers and citation tooling read it directly.

Research software is funded and staffed on the strength of the record it
can show. At the moment a project that uses this platform has nothing to
cite but the companion paper, and nothing at all that points at the
software as software.

Documentation only. No code changes.

## Proposed Changes

- Add `CITATION.cff` (CFF 1.2.0) naming the software, with the JCAA
  companion paper under `preferred-citation`, so citing tools offer both.

## How to Test

- `python3 -c "import yaml; yaml.safe_load(open('CITATION.cff'))"` parses.
- Optionally `cffconvert --validate` if you have it installed.
- Once merged, the "Cite this repository" button appears in the sidebar of
  the repository home page.

## Additional Information

Two things I would like a reviewer's judgement on, and one unrelated
finding.

**The author list is a starting point, not a settled answer.** It names
the project team as listed on the companion paper. The people who have
written most of this codebase are not on it. Please correct it: it is a
one-line-per-person change, and it should be decided by the people it
names rather than by me.

**Title.** Given as "Fieldmark (built on FAIMS)" so that both names are
findable, since work published before 2024 generally names the software
FAIMS or FAIMS Mobile.

**The repository has no licence as far as GitHub is concerned.** Not
introduced by this PR, but surfaced by it, because `CITATION.cff` has a
`license` field. There is no `LICENSE` file at the repository root, and
the root `package.json` says `"license": "ISC"`. Every actual LICENSE file
in the tree is Apache 2.0 (`app/`, `library/data-model/`, `docs/user/`),
and the source headers say Apache 2.0 with Macquarie University copyright.
I have put `Apache-2.0` in the citation file on that basis. If that is
right, a root `LICENSE` file and a corrected `package.json` would make the
licence visible on the repository page, where people look for it.

Verification for the metadata in the file: all six ORCIDs checked against
the ORCID public API on 2026-09-17; the paper's volume, issue, and page
range (5(1), 271-285) checked against CrossRef and the journal's own
metadata; version and release date taken from tag `v1.6.3`, dated
2026-09-03.

## Checklist

- [x] I have confirmed all commits have been signed.
[...]

PR #2322, opened 2026-09-17
docs: add a root LICENSE and correct the manifest

### LINKED ISSUE #1650: [BUG] No global license for the repository
Some components here in this repository have license files, which seem to denote the license of their directories' contents. There is no global license file however, which means that the license of the overall repository / project is unclear. 

**Expected behavior**
There should be a global license file.

saross:
## Ticket

Closes #1650.

Also surfaced independently whilst writing https://fieldmark.au/open-source,
and previously flagged in the Additional Information of #2321.

## Description

**GitHub reports no licence for this repository.** There is no `LICENSE`
file at the root, so the repository page shows no licence badge and
`GET /repos/FAIMS/FAIMS3` returns `"license": null`.

This is #1650, open since 30 September 2025 and raised by someone outside
the project: *"Some components here in this repository have license files,
which seem to denote the license of their directories' contents. There is
no global license file however, which means that the license of the
overall repository / project is unclear."* Someone assessing the project
could not tell what licence it carries, and said so.

The project is Apache 2.0 and says so in three places already:
`app/LICENSE`, `library/data-model/LICENSE`, `docs/user/LICENSE`, and in
the source headers, which carry Apache 2.0 with Macquarie University
copyright. What is missing is the declaration in the one place tooling
and readers look.

This matters beyond tidiness. The Fieldmark open-source page makes
openness the argument and links here, so a security reviewer or a
procurement assessor following that link lands on a repository with no
visible licence.

Documentation and metadata only. No code changes.

## Proposed Changes

- Add `LICENSE` at the repository root, copied verbatim from
  `docs/user/LICENSE`.
- Correct the root `package.json` from `"license": "ISC"` to
  `"license": "Apache-2.0"`.

On the choice of source: `docs/user/LICENSE` carries the complete Apache
2.0 text including the appendix, whilst `app/LICENSE` and
`library/data-model/LICENSE` are the same text truncated before it.
Copying the complete one means all four files now agree, and GitHub's
licence detection wants the unmodified text.

On `package.json`: `"ISC"` looks like a leftover scaffold default. It is
what npm metadata would carry for the project, so it is worth correcting
alongside the file. Happy to drop that hunk and leave it to the team if
you would rather this PR touched only the LICENSE file.

## How to Test

- After merge, the repository page shows an Apache-2.0 badge, and
  `gh api repos/FAIMS/FAIMS3 --jq .license.spdx_id` returns `Apache-2.0`.
- `diff LICENSE docs/user/LICENSE` is empty.

## Additional Information

Two things deliberately not changed, each worth its own decision:

- The appendix still carries the unmodified
  `Copyright [yyyy] [name of copyright owner]` template. That is the
  standard boilerplate for applying the licence rather than part of the
  grant, and leaving it unmodified is normal, but you may prefer it
  filled in.
- There is no `NOTICE` file. Apache 2.0 does not require one, and adding
  it is a question about attribution across Macquarie University, CSIRO,
  and Electronic Field Notebooks rather than a mechanical change.

Confirmed by Shawn Ross on 2026-09-17 that the project is Apache 2.0.

## Checklist


PeterBaker0 (approved):
Approved. Thanks for fixing. 

I also think the old approach of embedding headers with copyright statements into every file is a bit redundant. I don't often see that in OSS. Usually the head license trickles down, and if pertinent, we could have some copyright statement. 

It's an interesting question though, as does Macquarie University hold the copyright to _all_ user contributed material. For example, CSIRO has contributed significant IP to this project and therefore hold joint copyright over it. We haven't explicitly waived that right, and therefore retain copyright. 

My preference would be that we systematically wipe the license/copyright headers from all files. We then, if necessary, consider the statement of copyright, and consider the inclusion of significant joint contributors to that list.

PR #2324, opened 2026-09-17
v4 is still pointing to old packages, point it to right packages to i…

ranisa-gupta16:
## Description

GitHub's CI runner auto-updated the Android SDK cmdline-tools (first to 16.0, now 20.0). These newer versions removed the old tools package that setup-android tries to install by default. When sdkmanager can't find tools, the whole step fails with exit code 1.

What we changed:

Upgraded android-actions/setup-android from v3 to v4 (as Steve suggested — v4 handles license acceptance automatically)
Specified explicit packages (platform-tools, platforms;android-36, build-tools;36.0.0) so the action doesn't try to install the removed tools package

Why specifying packages is needed even on v4: The v4 action still defaults to installing tools, which no longer exists in the Android SDK repository. By listing only the packages we actually need, we skip the missing one entirely.

## Checklist


PeterBaker0 (approved):
Worth testing - thanks

PR #2329, opened 2026-09-19
chore: share one runtime stage across the dev api, app and web images

### LINKED ISSUE #2331: The dev Dockerfile builds three near-identical runtime stages
`Dockerfile` defines separate `api`, `app` and `web` stages from `node:24-slim`, each repeating the same apt, corepack and `COPY --from=builder` steps. They differ only in `EXPOSE` and `CMD`, and `docker-compose.yml` overrides both with its own `ports:` and `command:` on all three services, so neither line in the image is ever used.

Nothing notices when one breaks: `CMD [pnpm", "run", "web-dev"]` at line 87 is missing its opening quote and has never failed, because compose supplies the command instead.

`.dockerignore` excludes `dist`, `build`, `coverage` and `.turbo` at the repository root only, so each package's own copy still enters the build context: `api/build` 3.7M, `app/dist` 1.3M, `web/dist` 8.0M, `library/data-model/build` 6.9M, `library/forms/dist` 4.4M.

Found while tracking down local disk use from the dev images.

jayenashar:
# chore: share one runtime stage across the dev api, app and web images

## Ticket

Closes #2331

## Description

The dev image built three near-identical copies of the monorepo. Each service stage ran its own `COPY --from=builder /usr/src .`, and because the api stage installed gdal first it sat on a different parent chain, so its copy could not share layers with app and web.

The in-image `turbo build` was also producing output nothing could read. Compose bind-mounts `library/data-model` and `library/forms` over this image, so those build outputs were masked at runtime, and the dev servers compile their own sources at start anyway.

## Proposed Changes

- All three service stages derive from one `runtime` stage that copies the monorepo once. `gdal-bin` and `curl` move into it.
- Remove `RUN pnpm turbo build` from the dev image.
- Exclude nested `node_modules` from the build context. Only the root-level directory was ignored, so each package's `node_modules` symlink farm was copied over the image's own install; those links resolve only while the host install matches the lockfile. `turbo build` used to fail loudly on a stale host install, so with it gone this exclusion is what keeps the image's own dependency graph intact.
- Fix `CMD [pnpm", "run", "web-dev"]` in the web stage, which was missing its opening quote and silently fell back to shell form.

Measured on one machine, building api+app+web from an identical clean slate:

| | first build | another worktree, warm cache |
|---|---|---|
| before | 10.14 GB | 6.39 GB |
| after | 6.08 GB | 2.32 GB |

## How to Test

```
./localdev.sh --all --build
```

api on 8080, app on 3000, web on 3001 and couchdb on 5984 should all answer, and `docker compose ps` should show api healthy. `docker compose exec api ogr2ogr --version` confirms GeoPackage export still has its binary.

To see the build cost, run `docker builder prune -af` first, then `docker system df` before and after the build.

## Additional Information

Trade-offs worth knowing:

- First container start goes from roughly 15s to 110s, since vite and tsc now compile at startup instead of using prebuilt output. Later starts are unaffected.
- The images are no longer runnable standalone via `docker run` without the compose bind mounts. They are referenced only by `docker-compose.yml`; production uses `Dockerfile.build`, which is untouched.
- app and web now carry `gdal-bin`, which only api uses. BuildKit stores that layer once, so it costs nothing locally.

Verified by bringing the stack up: api, app, web and couchdb all returned 200, the api healthcheck reported healthy, and `ogr2ogr` resolved inside the api container.

## Checklist


.dockerignore
```diff
@@ -10,4 +10,5 @@ node_modules
 dist
 build
 coverage
-.turbo
\ No newline at end of file
+.turbo
+**/node_modules
```
jayenashar: `dist`, `build`, `coverage` and `.turbo` above have the same root-only problem this line fixes: each package's own copy still lands in the context (in my checkout api/build 3.7M, app/dist 1.3M, web/dist 8.0M, library/data-model/build 6.9M, library/forms/dist 4.4M). with turbo build gone nothing in the image regenerates them, so api/build becomes whatever the host last compiled, and the two library copies are dead weight under the compose bind mounts. `**/dist`, `**/build`, `**/.turbo` alongside this one.

Dockerfile
```diff
@@ -26,19 +26,19 @@ RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
   pnpm install --frozen-lockfile
 
 
-# Build stage
-FROM base AS builder
+# Source stage. Nothing is built here: compose bind-mounts the workspace libraries
+# over this image, so anything built here is hidden anyway (localdev.sh builds them).
+FROM base AS source
```
jayenashar: `COPY turbo.json ./` at line 22 now feeds a build that no longer happens, and it sits ahead of `pnpm install`, so editing turbo.json still invalidates the install layer. `COPY . .` here brings it in anyway.

Dockerfile
```diff
@@ -50,38 +50,20 @@ RUN corepack enable
...
-ENV PATH="$PNPM_HOME:$PATH"
-RUN corepack enable
-
-
-WORKDIR /usr/src
-COPY --from=builder /usr/src .
+FROM runtime AS web
 EXPOSE 3001
-CMD [pnpm", "run", "web-dev"]
+CMD ["pnpm", "run", "web-dev"]
```
jayenashar: compose sets `command:` on all three services (docker-compose.yml:26, :48, :70), so none of these CMDs ever run, which is why the missing quote went unnoticed. `ports:` supersedes EXPOSE the same way. with both inert the three stages are identical; any reason not to drop them and point all three compose targets at `runtime`?

PeterBaker0 (changes requested):
So the Dockerfile no longer builds the packages?

As per the removal of this line

```text
[-] RUN pnpm turbo build --filter=@faims3/api --filter=@faims3/app --filter=@faims3/web
```

I don't like this compromise - is there a way to do some build optimisation without this trade off?

This will also make cold start way worse for auto scaling scenarios right? 115 seconds is a very long cold start - and it's always going to be 'first start'... What does a non first-start mean in Docker land? One where you started with a volume mount and it compiled files into the persisted mount? This will never happen in a cloud context.

jayenashar:
> So the Dockerfile no longer builds the packages?
> 
> As per the removal of this line
> 
> ```
> [-] RUN pnpm turbo build --filter=@faims3/api --filter=@faims3/app --filter=@faims3/web
> ```
> 
> I don't like this compromise - is there a way to do some build optimisation without this trade off?
> 
> This will also make cold start way worse for auto scaling scenarios right? 115 seconds is a very long cold start - and it's always going to be 'first start'... What does a non first-start mean in Docker land? One where you started with a volume mount and it compiled files into the persisted mount? This will never happen in a cloud context.

sorry i thought there was a different Dockerfile for prod.  let me get back to you.

jayenashar:
put the build back, and renamed the file to `Dockerfile.dev`.

the only thing that reads any of that build output is the api container: `watch-api` is nodemon running `tsc --incremental && node .`, and compose mounts api/src, views, keys, public, test and .env but not api/build, so the build directory and its tsbuildinfo sit in the container's own layer. data-model and forms are masked by their bind mounts and localdev.sh builds them on the host before `compose up`, and app and web are vite dev servers over src, so nothing ever read their output.

cost of keeping it, building `--target runtime` both ways here: 1.55 GB to 1.69 GB, and the build step takes 169s. most of the 138 MB lands under node_modules rather than in the output.

the rename turned up one thing worth knowing: web had no `dockerfile:` line and was taking the default name, and the third `dockerfile: Dockerfile` in compose is couchdb's own file under api/couchdb. both named explicitly now.

PR #2332, opened 2026-09-21
fix: mobile app API CORS hotfix

PeterBaker0:
Adds proper origins for android/iOS apps - breaking apps right now.


ranisa-gupta16 (approved):
thanks for the prompt fix..

PR #2333, opened 2026-09-21
Fix/record search fails with related record values

stevecassidy:
# Fix/record search fails with related record values

## Ticket

None

## Description

Searching records via useRecord crashes silently when there are related record values in some records.   To reproduce, add a record with a related record, then use the search bar in the notebook - no results are shown.  

This is due to a bug/quirk in pouchdb - the format of regex is not well documented, we supply a JS RegExp object for case insensitive searches and this causes a crash when a document property has a value that is an array of objects.    We give a string pattern for case-sensitive search and this doesn't cause the crash.   The issue is with supplying a RegExp object, but this is the only way to send a case-insensitive regexp to the pouchdb engine. 

## Proposed Changes

Adds a test of searchRecordsByRegex which fails with the existing code.   Replaces the use of regex in the find query with an in-memory match after retrieving the data property from all avps. 

This pulls some work from pouchdb into the main codebase which could have a performance hit.   

## How to Test

Create notebook records with related records fields.   Search for a value - you should now see some results in the search.

## Checklist


stevecassidy:
Pouchdb issue filed: https://github.com/apache/pouchdb/issues/9278

PeterBaker0 (approved):
As mentioned - potential performance issues. Searching client side over the synced records has the potential to be quite performant though - we can likely build a client side index if the need arises. It'd be nice to build a fuzzy search. But given the previous approach just doesn't work - worthwhile for now.

